PR #6, opened 2018-10-05
add logo

MaxUlysse:


README.md
```diff
@@ -1,3 +1,4 @@
+# ![mag](https://raw.githubusercontent.com/nf-core/mag/master/docs/images/mag_logo.png.png)
```
ewels: Do you need the full URL here? Would `docs/images/mag_logo.png.png` not be better?

I guess you're going to say something like this is needed for dockerhub etc. In which case, ok 😉

ewels: Double `.png` here too

MaxUlysse: That was exactly that, dockerhub etc.

PR #12, opened 2019-01-21
Psuedo-PR for review

HadrienG:
😨

PR #15, opened 2019-03-06
WIP: Add Illumina / Nanopore Hybrid Assembly

d4straub:
This is still work in progress and needs polishing, also documentation isn't updated yet, however core functions work.

### Additions:
- Input via manifest file to allow Illumina & Nanopore files for multiple samples
- Illumina reads: PhiX removal
- Nanopore reads quality plots: https://github.com/wdecoster/NanoPlot
- Nanopore reads quality filtering: https://github.com/rrwick/Filtlong
- Nanopore reads removing adapters: https://github.com/rrwick/Porechop
- Nanopore reads: Lambda removal: https://github.com/wdecoster/nanolyse
- Hybrid assembly: https://github.com/ablab/spades
- ~~Assembly graph visualization: https://github.com/rrwick/Bandage~~
- Bin quality check: https://gitlab.com/ezlab/busco & http://quast.sourceforge.net/metaquast

### Problems:
Most of the above mentioned additions require python 3, therefore I upgraded from python 2 to 3, this means:
- replace CheckM with BUSCO for quantitative measures for the assessment of genome assembly
- loose CheckM merge genome bins
- loose RefineM

Solution might be to either come up with alternative programs or add a second environment with python 2.7 and CheckM/RefineM as done in https://github.com/nf-core/rnafusion/blob/dev/download-singularity-img.nf

### Ideas for further improvement:
- Assembly: allow assembly of multiple samples instead/ in addition to treating them individually
- Binning: improve binning by using depth information from several samples when available with `jgi_summarize_bam_contig_depths --outputDepth depth.txt *.bam` as decribed in https://bitbucket.org/berkeleylab/metabat

### DIFF
--- a/bin/scrape_software_versions.py
+++ b/bin/scrape_software_versions.py
@@ -11,8 +11,15 @@ regexes = {
     'fastp': ['v_fastp.txt', r"fastp (\S+)"],
     'megahit': ['v_megahit.txt', r"MEGAHIT v(\S+)"],
     'metabat': ['v_metabat.txt', r"version (\S+)"],
-    'checkm': ['v_checkm.txt', r"CheckM v(\S+)"],
-    'refinem': ['v_refinem.txt', r"RefineM v(\S+)"]
+    #'checkm': ['v_checkm.txt', r"CheckM v(\S+)"],
+    #'refinem': ['v_refinem.txt', r"RefineM v(\S+)"],
+    'NanoPlot': ['v_nanoplot.txt', r"NanoPlot (\S+)"],
+    'Filtlong': ['v_filtlong.txt', r"Filtlong v(\S+)"],
+    'porechop': ['v_porechop.txt', r"(\S+)"],
+    'NanoLyse': ['v_nanolyse.txt', r"NanoLyse (\S+)"],
+    'SPAdes': ['v_spades.txt', r"SPAdes v(\S+)"],
+    'BUSCO': ['v_busco.txt', r"BUSCO (\S+)"],
+    'Bandage': ['v_bandage.txt', r"Version: (\S+)"]
 }
 results = OrderedDict()
 results['nf-core/mag'] = '<span style="color:#999999;\">N/A</span>'
@@ -22,8 +29,15 @@ results['fastqc'] = '<span style="color:#999999;\">N/A</span>'
 results['fastp'] = '<span style="color:#999999;\">N/A</span>'
 results['megahit'] = '<span style="color:#999999;\">N/A</span>'
 results['metabat'] = '<span style="color:#999999;\">N/A</span>'
-results['checkm'] = '<span style="color:#999999;\">N/A</span>'
-results['refinem'] = '<span style="color:#999999;\">N/A</span>'
+#results['checkm'] = '<span style="color:#999999;\">N/A</span>'
+#results['refinem'] = '<span style="color:#999999;\">N/A</span>'
+results['NanoPlot'] = '<span style="color:#999999;\">N/A</span>'
+results['Filtlong'] = '<span style="color:#999999;\">N/A</span>'
+results['porechop'] = '<span style="color:#999999;\">N/A</span>'
+results['NanoLyse'] = '<span style="color:#999999;\">N/A</span>'
+results['SPAdes'] = '<span style="color:#999999;\">N/A</span>'
+results['BUSCO'] = '<span style="color:#999999;\">N/A</span>'
+results['Bandage'] = '<span style="color:#999999;\">N/A</span>'
 
 # Search each file using its regex
 for k, v in regexes.items():
--- a/bin/scrape_software_versions.py
+++ b/bin/scrape_software_versions.py
@@ -11,8 +11,6 @@ regexes = {
     'fastp': ['v_fastp.txt', r"fastp (\S+)"],
     'megahit': ['v_megahit.txt', r"MEGAHIT v(\S+)"],
     'metabat': ['v_metabat.txt', r"version (\S+)"],
-    #'checkm': ['v_checkm.txt', r"CheckM v(\S+)"],
-    #'refinem': ['v_refinem.txt', r"RefineM v(\S+)"],
     'NanoPlot': ['v_nanoplot.txt', r"NanoPlot (\S+)"],
     'Filtlong': ['v_filtlong.txt', r"Filtlong v(\S+)"],
     'porechop': ['v_porechop.txt', r"(\S+)"],
@@ -29,8 +27,6 @@ results['fastqc'] = '<span style="color:#999999;\">N/A</span>'
 results['fastp'] = '<span style="color:#999999;\">N/A</span>'
 results['megahit'] = '<span style="color:#999999;\">N/A</span>'
 results['metabat'] = '<span style="color:#999999;\">N/A</span>'
-#results['checkm'] = '<span style="color:#999999;\">N/A</span>'
-#results['refinem'] = '<span style="color:#999999;\">N/A</span>'
 results['NanoPlot'] = '<span style="color:#999999;\">N/A</span>'
 results['Filtlong'] = '<span style="color:#999999;\">N/A</span>'
 results['porechop'] = '<span style="color:#999999;\">N/A</span>'
--- a/bin/scrape_software_versions.py
+++ b/bin/scrape_software_versions.py
@@ -17,7 +17,6 @@ regexes = {
     'NanoLyse': ['v_nanolyse.txt', r"NanoLyse (\S+)"],
     'SPAdes': ['v_spades.txt', r"SPAdes v(\S+)"],
     'BUSCO': ['v_busco.txt', r"BUSCO (\S+)"],
-    'Bandage': ['v_bandage.txt', r"Version: (\S+)"]
 }
 results = OrderedDict()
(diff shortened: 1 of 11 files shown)


bin/scrape_software_versions.py
```diff
@@ -11,8 +11,15 @@
     'fastp': ['v_fastp.txt', r"fastp (\S+)"],
     'megahit': ['v_megahit.txt', r"MEGAHIT v(\S+)"],
     'metabat': ['v_metabat.txt', r"version (\S+)"],
-    'checkm': ['v_checkm.txt', r"CheckM v(\S+)"],
-    'refinem': ['v_refinem.txt', r"RefineM v(\S+)"]
+    #'checkm': ['v_checkm.txt', r"CheckM v(\S+)"],
```
MaxUlysse: Why keeping it here as a comment when you removed it from the `environment.yml`?

d4straub: I believe I removed now all checkm/refinem code

d4straub:
The workflow works fine with test samples and `-profile conda`. 
Focusing now on the container, there appear some library problems. When these container issues are solved I am going to test on "real" data and also add a test data set/profile.

ewels:
> Focusing now on the container, there appear some library problems.

If you mean that the CI tests are failing because of missing tools, this could be because the docker container will only update _after_ this pull-request is merged.

This pipeline has not yet been released, so we're less strict with pull-requests than we'd usually be (this starts after the first [mega-PR](https://github.com/nf-core/mag/pull/12) has been reviewed for the first release). So if @HadrienG is happy with this PR and you think it's ready to be merged, I say go for it. The next time you make a PR with any change, the tests should pass because the docker container will have updated.

d4straub:
> If you mean that the CI tests are failing because of missing tools, this could be because the docker container will only update _after_ this pull-request is merged.

Unfortunately I don't mean the CI tests but the containers I build locally based on the `Dockerfile` and `environment.yml`

HadrienG:
Hej folks!

Sorry for the long silence. I'll get on this this week.
Thanks @d4straub for what looks like a lot of work already!

d4straub:
Hi, I am right now testing the pipeline for (1) a non-complex Illumina-only and (2) a very complex Illumina-Nanopore hybrid assembly. While the first one is going well, it looks like the hybrid assembly process will need a couple of hours/days more. I'll let you know how it goes.

main.nf
```diff
@@ -218,24 +301,129 @@ process get_software_versions {
...
     """
 }
 
 
+/*
+ * Trim adapter sequences on long read nanopore files
+ */
+process porechop { 
+    tag "$id"
+
```
HadrienG: porechop should be optional in my opinion, as the most recent ONT basecaller (guppy) now also removes adapter sequences.

main.nf
```diff
@@ -218,24 +301,129 @@ process get_software_versions {
...
+    """
+    porechop -i ${lr} -t "${task.cpus}" -o ${id}_porechop.fastq
+    """
+}
+
+/*
+ * Remove reads mapping to the lambda genome.
+ * TODO: add lambda phage to igenomes.config?
+ */
+if (!params.keep_lambda) {
```
HadrienG: I think you can use a `when` clause instead of the `if` here: [https://www.nextflow.io/docs/latest/process.html#when](https://www.nextflow.io/docs/latest/process.html#when)

main.nf
```diff
@@ -284,10 +474,66 @@ process fastp {
     }
 }
 
+/*
+ * Remove PhiX contamination from Illumina reads
+ * TODO: PhiX into/from iGenomes.conf?
+ */
+if(!params.keep_phix) {
```
HadrienG: Couldn't we use the same process for both long and short reads? minimap2 (and nanolyse) should work fine enough with short reads, and will be much faster.

HadrienG:
One thing we need to discuss is that I'm not convinced with should get rid of `checkm` and `refinem`.

Are they a nightmare to integrate to the pipeline? Absolutely, but that they are written in python2 and that they have horribly inconsistent and clunky command-line arguments is not good enough argument to remove them in the pipeline without replacing them with something.

Busco is an ok replacement for the QC step of `checkm`, but we still miss the merging and refining steps. In my experience `refinem` often improve the bins a tiny bit. Have you had a different experience?

main.nf
d4straub: As far as I understand the when clause would also not forward any results, therefore the pipeline would just discard that info. This is fine for final results such as BUSCO, but not with intermediate steps like this one. If this step is skipped the files still have to be forwarded into the following channels, as see in 
https://github.com/d4straub/mag/blob/5b925b16f728ea65cb6fbbbc534b623072abd5a8/main.nf#L365-L368

main.nf
d4straub: I'll do that

main.nf
d4straub: minimap2 might be an option for short reads, I chose nanolyse for long reads since it is a specialized tool for this task. I am not insisting on bowtie2.

d4straub:
I have no experience with `refinem`, I think it would be nice to have and try out.

My reasons why this isn't in here and I wont be able to do that soon is that (1) I have no experience with it (positive or otherwise), therefore no strong urge to use it, (2) have to prioritize my time and this is low priority, (3) nf-core suggests having one container but it would require a second container, if I am not mistaken.

My next steps would be to fix small problems that you pointed out or that may occur with my current assemblies and add a test data set.

HadrienG:
I range to your opinion, let's simply have BUSCO for now.

I have an (unmerged) illumina test dataset [here](https://github.com/HadrienG/test-datasets/tree/mag), do you want me to simulate long reads and add them to it?

main.nf
HadrienG: this is correct, you'd still need a if clause to set up the Channels correctly

conf/binac_smp.config
```diff
@@ -0,0 +1,23 @@
+//Profile config names for nf-core/configs
```
HadrienG: I think cluster configs should be in [https://github.com/nf-core/configs](https://github.com/nf-core/configs)

d4straub: Absolutely!

However, to assemble a complex metagenome SPAdes needed ~350 GB memory (and MEGAHIT <40 GB), which isnt provided by the cluster queue in the current nf-core/configs/binac.config. Therefore I had to do an additional config that specifies for that cluster a queue for high memory jobs and integrated it (temporary) here. Ultimately I would like to use the multiple queue logic discussed here: https://github.com/nf-core/configs/issues/39
I honestly dont know how we could right now make the workflow work for the cluster I work with, without that specific config. And I certainly dont want to have that in nf-core/configs since it is only necessary for this pipeline and might be not needed in future at all.

d4straub:
> I range to your opinion, let's simply have BUSCO for now.

Alright, we can later attempt to re-implement `refinem`

> I have an (unmerged) illumina test dataset [here](https://github.com/HadrienG/test-datasets/tree/mag), do you want me to simulate long reads and add them to it?

That would be great!

> When you're done we can merge in `nfcore/dev`, then I'll have a few issues with my own code to resolve.

Great, I'll let you know, hopefully very soon.

d4straub:
I am planning one more round of adjustments:
- make porechop optional
- adjust base.config once more so that with 3 retries all memory/time requirements are met

d4straub:
Alright, so far its working.

I think we could merge this branch into dev and continue to improve it from there.

E.g. removing conf/binac_smp.config, use miniasm2 instead of bowtie2, use PhiX and lambda phage reference genome from iGenomes, ...

HadrienG:
Here they are: [https://github.com/HadrienG/test-datasets/tree/mag/test_data](https://github.com/HadrienG/test-datasets/tree/mag/test_data).

Do you wanna fix the test before merging?

d4straub:
Hey, the following error occured:

> ERROR ~ Argument of `file` function cannot be null

--> The manifest file is supposed to be tab separated. Also, I think absolute file paths are needed.

I downloaded your files, edited the manifest and tested again.
Unfortunately the long read file was not accepted. The error message I received was:

> ValueError: Records in Fastq files should start with '@' character

--> A 'nanopore' fastq file is expected and you provided a fasta. 

Could you edit the manifest file and upload a fastq file that mimics nanopore?

d4straub:
I actually think we should merge it soon that we can both actually test these kind of things. Only after the merge the container will be build and running the pipeline will work for you as well.

HadrienG:
Unfortunately [NanoSim](https://github.com/bcgsc/NanoSim) does not seem to output fastq. I'l try and fake some quality scores.

Concerning the manifest, it seems impossible to me to write absolute paths, since we're testing on different machines.

d4straub:
I mean file paths pointing to the github files so that they will be downloaded.
Also, the fastq files from nanopore are somehow expected to contain '@' instead of '>' for each sequence header.

HadrienG:
done and done: <https://github.com/HadrienG/test-datasets/tree/mag>

d4straub:
Could you please use the attached file as your manifest.txt: 
[manifest.txt](https://github.com/nf-core/mag/files/3089198/manifest.txt)
Difference is (1) it is tab separated and (2) it uses the correct paths

d4straub:
Great, running with `-profile test_hybrid` or `-profile test` works now.

HadrienG:
Let's merge this then!

PR #22, opened 2019-08-06
Pseudo-PR for review

HadrienG:


main.nf
```diff
@@ -3,76 +3,118 @@
...
-    =========================================
     Usage:
 
     The typical command for running the pipeline is as follows:
 
     nextflow run nf-core/mag --reads '*_R{1,2}.fastq.gz' -profile docker
+    nextflow run nf-core/mag --manifest 'manifest.tsv' -profile docker
 
     Mandatory arguments:
       --reads                       Path to input data (must be surrounded with quotes)
```
alneberg: Maybe wanna clarify if this is for short reads, long reads or both?

docs/usage.md
```diff
@@ -27,38 +70,46 @@ results         # Finished results (configurable, see below)
...
+  - A generic configuration profile to be used with [Docker](http://docker.com/)
+  - Pulls software from dockerhub: [`nfcore/mag`](http://hub.docker.com/r/nfcore/mag/)
+- `singularity`
+  - A generic configuration profile to be used with [Singularity](http://singularity.lbl.gov/)
+  - Pulls software from DockerHub: [`nfcore/mag`](http://hub.docker.com/r/nfcore/mag/)
+- `test`
+  - A profile with a complete configuration for automated testing
+  - Includes links to test data so needs no other parameters
 
 ### `--reads`
```
alneberg: Maybe wanna clarify if this is for short reads, long reads or both?

main.nf
alneberg: It seems to be using `--readPaths` instead?

main.nf
```diff
@@ -167,46 +486,736 @@ process fastqc {
...
+        """
+        bowtie2-build --threads "${task.cpus}" "${assembly}" ref
+        bowtie2 -p "${task.cpus}" -x ref -1 "${reads[0]}" -2 "${reads[1]}" | \
+            samtools view -@ "${task.cpus}" -bS | \
+            samtools sort -@ "${task.cpus}" -o "${name}.bam"
+        samtools index "${name}.bam"
+        jgi_summarize_bam_contig_depths --outputDepth depth.txt "${name}.bam"
+        metabat2 -t "${task.cpus}" -i "${assembly}" -a depth.txt -o "MetaBAT2/${name}" -m ${min_size}
+
+        #if bin foolder is empty
```
alneberg: Typo

main.nf
```diff
@@ -167,46 +486,736 @@ process fastqc {
...
+    def name = "${assembler}-${sample}"
+    if ( !params.singleEnd ) {
+        """
+        bowtie2-build --threads "${task.cpus}" "${assembly}" ref
+        bowtie2 -p "${task.cpus}" -x ref -1 "${reads[0]}" -2 "${reads[1]}" | \
+            samtools view -@ "${task.cpus}" -bS | \
+            samtools sort -@ "${task.cpus}" -o "${name}.bam"
+        samtools index "${name}.bam"
+        jgi_summarize_bam_contig_depths --outputDepth depth.txt "${name}.bam"
+        metabat2 -t "${task.cpus}" -i "${assembly}" -a depth.txt -o "MetaBAT2/${name}" -m ${min_size}
```
alneberg: I think the binning performance would improve a lot if one could use multiple sample bam files as input to metabat.  Is it too much for the first release? Also, moving the mapping out as an independent step would improve stability and cause less time consuming reruns.

main.nf
MaxUlysse: I think we got that covered with #25

main.nf
HadrienG: Done!!!

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,47 @@
+# nf-core/mag: Contributing Guidelines
+
+Hi there! Many thanks for taking an interest in improving nf-core/mag.
+
+We try to manage the required tasks for nf-core/mag using GitHub issues, you probably came to this page when creating one. Please use the pre-filled template to save time.
+
+However, don't be put off by this template - other more general issues and suggestions are welcome! Contributions to the code are even more welcome ;)
+
+> If you need help using or modifying nf-core/mag then the best place to ask is on the pipeline channel on [Slack](https://nf-core-invite.herokuapp.com/).
```
MaxUlysse:
```suggestion
> If you need help using or modifying nf-core/mag then the best place to ask is on the pipeline channel on [Slack](https://nf-co.re/join/slack).
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,47 @@
...
+If any failures or warnings are encountered, please follow the listed URL for more documentation.
+
+### Pipeline Tests
+Each nf-core pipeline should be set up with a minimal set of test-data.
+Travis CI then runs the pipeline on this data to ensure that it exists successfully.
+If there are any failures then the automated tests fail.
+These tests are run both with the latest available version of Nextflow and also the minimum required version that is stated in the pipeline code.
+
+## Getting help
+For further information/help, please consult the [nf-core/mag documentation](https://github.com/nf-core/mag#documentation) and don't hesitate to get in touch on the pipeline channel on [Slack](https://nf-core-invite.herokuapp.com/).
```
MaxUlysse:
```suggestion
For further information/help, please consult the [nf-core/mag documentation](https://github.com/nf-core/mag#documentation) and don't hesitate to get in touch on the pipeline channel on [Slack](https://nfcore.slack.com/channels/mag).
```

docs/output.md
```diff
@@ -1,35 +1,165 @@
...
-  * FastQC report, containing quality metrics for your untrimmed raw fastq files
-* `zips/sample_fastqc.zip`
-  * zip file containing the FastQC report, tab-delimited data file and plot images
+- `report.txt`: Tab-delimited result file. See the [centrifuge manual](https://ccb.jhu.edu/software/centrifuge/manual.shtml#centrifuge-classification-output) for information about the fields
+- `kreport.txt`: Classification in the Kraken report format. See the [kraken manual](http://ccb.jhu.edu/software/kraken/MANUAL.html#sample-reports) for more details
+- `taxonomy.krona.html`: an interactive pie chart produced by [KronaTools](https://github.com/marbl/Krona/wiki)
+
+### CAT
+
+<!-- TODO: CAT output -->
```
MaxUlysse: Need docs

nextflow.config
```diff
@@ -3,73 +3,95 @@
...
+
+// Load nf-core custom profiles from different Institutions
+try {
+  includeConfig "${params.custom_config_base}/nfcore_custom.config"
+} catch (Exception e) {
+  System.err.println("WARNING: Could not load nf-core/config profiles: ${params.custom_config_base}/nfcore_custom.config")
+}
+
+profiles {
+  awsbatch { includeConfig 'conf/awsbatch.config' }
```
MaxUlysse:
```suggestion
```

bin/summary_busco.py
```diff
@@ -0,0 +1,40 @@
...
+
+#USAGE: ./summary.busco.py *.txt
+
+import re
+from sys import argv
+
+#"# Summarized benchmarking in BUSCO notation for file MEGAHIT-testset1.contigs.fa"
+#"	C:0.0%[S:0.0%,D:0.0%],F:0.0%,M:100.0%,n:148"
+
+regexes = {
```
ewels: Is this meant to be here? Looks like it's from the software versions script, and is immediately overwritten..

conf/base.config
```diff
@@ -11,29 +11,95 @@
...
+  //SPAdes returns error(1) if it runs out of memory (and for other reasons as well...)!
+  //exponential increase of memory and time with attempts
+  withName: spades {
+    cpus = { check_max (10 * task.attempt, 'cpus' ) }
+    memory = { check_max (64.GB * (2**(task.attempt-1)), 'memory' ) }
+    time = { check_max (24.h * (2**(task.attempt-1)), 'time' ) }
+    errorStrategy = { task.exitStatus in [143,137,1] ? 'retry' : 'ignore' }
+  }
+  withName: spadeshybrid {
+    cpus = { check_max (10 * task.attempt, 'cpus' ) }
```
ewels: You could probably make this config quite a bit shorter by using labels instead of referring to each process by name. But this should be fine as it is for now 👍 (next hackathon..??)

conf/binac_smp.config
```diff
@@ -0,0 +1,23 @@
+//Profile config names for nf-core/configs
+params {
+  config_profile_description = 'BINAC cluster profile for jobs requiring >128 GB memory provided by nf-core/configs.'
```
ewels: This should be removed as it is now available in nf-core/configs

conf/test_hybrid.config
```diff
@@ -0,0 +1,19 @@
+/*
+ * -------------------------------------------------
+ *  Nextflow config file for running tests
+ * -------------------------------------------------
+ * Defines bundled input files and everything required
+ * to run a fast and simple test. Use as follows:
+ *   nextflow run nf-core/methylseq -profile test
```
ewels:
```suggestion
 *   nextflow run nf-core/methylseq -profile test_hybrid
```

main.nf
```diff
@@ -81,83 +123,360 @@ if( !(workflow.runName ==~ /[a-z]+_[a-z]+/) ){
...
-    }
-} catch (all) {
-    log.error "====================================================\n" +
-              "  Nextflow version $params.nf_required_version required! You are running v$workflow.nextflow.version.\n" +
-              "  Pipeline execution will continue, but things may break.\n" +
-              "  Please run `nextflow self-update` to update Nextflow.\n" +
-              "============================================================"
+summary['Run Name']         = custom_runName ?: workflow.runName
+summary['Reads']                = params.reads
+summary['Fasta Ref']            = params.fasta
```
ewels: I don't think this is defined or used anywhere?

main.nf
```diff
@@ -81,83 +123,360 @@ if( !(workflow.runName ==~ /[a-z]+_[a-z]+/) ){
...
 
+/*
+ * short read preprocessing options
+ */
+params.adapter_forward = "AGATCGGAAGAGCACACGTCTGAACTCCAGTCA"
+params.adapter_reverse = "AGATCGGAAGAGCGTCGTGTAGGGAAAGAGTGT"
+params.mean_quality = 15
+params.trimming_quality = 15
+params.keep_phix = false
+// params.phix_reference = "ftp://ftp.ncbi.nlm.nih.gov/genomes/genbank/viral/Enterobacteria_phage_phiX174_sensu_lato/all_assembly_versions/GCA_002596845.1_ASM259684v1/GCA_002596845.1_ASM259684v1_genomic.fna.gz"
```
ewels: What's going on with this commented line?

main.nf
```diff
@@ -81,83 +123,360 @@ if( !(workflow.runName ==~ /[a-z]+_[a-z]+/) ){
...
+params.phix_reference = "$baseDir/assets/data/GCA_002596845.1_ASM259684v1_genomic.fna.gz"
+
+/*
+ * binning options
+ */
+params.skip_binning = false
```
ewels: All of these defaults should go in `nextflow.config` instead of `main.nf`

The only exceptions are where we are setting `params.something` with more complex logic. For example, the iGenomes references.

main.nf
```diff
@@ -81,83 +123,360 @@ if( !(workflow.runName ==~ /[a-z]+_[a-z]+/) ){
...
-Channel
-    .fromFilePairs( params.reads, size: params.singleEnd ? 1 : 2 )
-    .ifEmpty { exit 1, "Cannot find any reads matching: ${params.reads}\nNB: Path needs to be enclosed in quotes!\nNB: Path requires at least one * wildcard!\nIf this is single-end data, please specify --singleEnd on the command line." }
-    .into { read_files_fastqc }
+if(!params.skip_busco){
+    Channel
+        .fromPath( "${params.busco_reference}", checkIfExists: true )
+        .set { file_busco_db }
+} else {
+    file_busco_db = Channel.from()
```
ewels:
```suggestion
    file_busco_db = Channel.from(false)
```
I forget why the `false` is needed, but I remember hitting problems without it before I think..

main.nf
```diff
@@ -81,83 +123,360 @@ if( !(workflow.runName ==~ /[a-z]+_[a-z]+/) ){
...
+}
+
+if(params.centrifuge_db){
+    Channel
+        .fromPath( "${params.centrifuge_db}", checkIfExists: true )
+        .set { file_centrifuge_db }
+} else {
+    file_centrifuge_db = Channel.from()
```
ewels: As above

main.nf
```diff
@@ -81,83 +123,360 @@ if( !(workflow.runName ==~ /[a-z]+_[a-z]+/) ){
...
+if(params.manifest){
+    manifestFile = file(params.manifest)
+    // extracts read files from TSV and distribute into channels
+    Channel
+        .from(manifestFile)
+        .ifEmpty {exit 1, log.info "Cannot find path file ${tsvFile}"}
+        .splitCsv(sep:'\t')
+        .map { row ->
+            def id = row[0]
+            def lr = returnFile( row[1] )
```
ewels: Instead of this, I think you can just use `checkifExists`
```suggestion
            def lr = file( row[1], checkIfExists: true )
```

Then you can remove your custom function above 👍

main.nf
```diff
@@ -81,83 +123,360 @@ if( !(workflow.runName ==~ /[a-z]+_[a-z]+/) ){
...
+summary['Config Profile']       = workflow.profile
+if(params.config_profile_description) summary['Config Description'] = params.config_profile_description
+if(params.config_profile_contact)     summary['Config Contact']     = params.config_profile_contact
+if(params.config_profile_url)         summary['Config URL']         = params.config_profile_url
+if(params.email) {
+  summary['E-mail Address']     = params.email
+  summary['MultiQC maxsize']    = params.maxMultiqcEmailFileSize
 }
+log.info summary.collect { k,v -> "${k.padRight(18)}: $v" }.join("\n")
+log.info "\033[2m----------------------------------------------------\033[0m"
```
ewels:
```suggestion
log.info "-\033[2m----------------------------------------------------\033[0m-"
```
This fixes the asci colouring..

main.nf
```diff
@@ -81,83 +123,360 @@ if( !(workflow.runName ==~ /[a-z]+_[a-z]+/) ){
...
+    tag "$id"
+    publishDir "${params.outdir}/QC_longreads/NanoPlot_${id}", mode: 'copy'
+    
+    input:
+    set id, file(lr), type from files_nanoplot_raw.mix(files_nanoplot_filtered)
+
+    output:
+    file '*.png'
+    file '*.html'
+    file '*.txt'
```
ewels:
```nextflow
file '*.{png,html,txt}'
```

main.nf
```diff
@@ -167,46 +486,759 @@ process fastqc {
 
     script:
     """
-    fastqc -q $reads
+    fastqc -t "${task.cpus}" -q $reads
```
ewels: Huh, I wonder if we should be doing this for all pipelines... 🤔

ewels: Could you please add a PR to the main template for this?

main.nf
```diff
@@ -167,46 +486,759 @@ process fastqc {
...
+        """
+    }
+    else {
+        """
+        fastp -w "${task.cpus}" -q "${qual}" --cut_by_quality5 \
+            --cut_by_quality3 --cut_mean_quality "${trim_qual}"\
+            --adapter_sequence="${adapter}" --adapter_sequence_r2="${adapter_reverse}" \
+            -i ${reads} -o "${name}_trimmed.fastq.gz"
+        """
+    }
```
ewels: As this is mostly the same, it might be nicer to have a little groovy code above this which sets the additional command line optoins if needed. eg:

```nextflow
script:
def pe_input = params.singleEnd ? '' :  "-I \"${reads[1]}\""
def pe_output1 = params.singleEnd ? "-o \"${name}_trimmed.fastq.gz\"" :  "-o \"${name}_trimmed_R1.fastq.gz\""
def pe_output2 = params.singleEnd ? '' :  "-O \"${name}_trimmed_R2.fastq.gz\""
"""
fastp -w ... $pe_input $pe_output1 $pe_output2
"""
```

..etc

ewels: This means that there is no duplication of the script block, so less vulnerable to copy pasta errors in the future

main.nf
```diff
@@ -167,46 +486,759 @@ process fastqc {
...
+    file("report.txt")
+    file("kreport.txt")
+
+    script:
+    if ( !params.singleEnd ) {
+    """
+    centrifuge -x "${db_name}" \
+        -p "${task.cpus}" \
+        -1 "${reads[0]}" \
+        -2 "${reads[1]}" \
```
ewels: As above, the differences between these two seem to be fairly minor.

main.nf
```diff
@@ -167,46 +486,759 @@ process fastqc {
...
+    script:
+    if ( !params.singleEnd ) {
+    """
+    kraken2 \
+        --report-zero-counts \
+        --threads "${task.cpus}" \
+        --db database \
+        --fastq-input \
+        --report kraken2_report.txt \
+        --paired "${reads[0]}" "${reads[1]}" \
```
ewels: As above again

main.nf
```diff
@@ -167,46 +486,759 @@ process fastqc {
...
+        --threads "${task.cpus}" \
+        --memory "$maxmem" \
+        --pe1-1 ${sr[0]} \
+        --pe1-2 ${sr[1]} \
+        --nanopore ${lr} \
+        -o spades
+    cp spades/assembly_graph_with_scaffolds.gfa ${id}_graph.gfa
+    cp spades/scaffolds.fasta ${id}_scaffolds.fasta
+    cp spades/contigs.fasta ${id}_contigs.fasta
+    cp spades/spades.log ${id}_log.txt
```
ewels: Why do you need to copy these? Can you not just specify their original paths in the output?

main.nf
HadrienG: it's already in the template 😁

main.nf
HadrienG: There are no options in SPAdes to change the prefix of the output files. If we have several samples there will be conflicting names. or you mean I should rename on the fly in the publishDir code? (using a saveAs closure to add a prefix to all the files felt complicated)

HadrienG: Renaming in the `script` chunk felt easier because then we can catch the four files we want with `pattern: "${id}*"` in the `publishDir`, not sure how to combine that with renaming with `saveAs`

main.nf
ewels: Hah, great!

main.nf
ewels: I’ll try to write a suggestion (or maybe someone else can). `saveAs` for publishDir and can use a var alongside the files in the output channels to organise into folders in downstream processes if needed.

main.nf
HadrienG: done!

main.nf
HadrienG: I could `mv` instead of `copy` though, as seen in the assembly step of the `bacass` pipeline

ewels: Are you sure that the `publishDir` thing is working? To me, it looks like it is setting the filename for anything that doesn't contain `".fastq.gz"` to `null`, which means that it won't be saved.. My expectation is that _only_ the `fastq.gz` files are being saved to the results directory here.

main.nf
```diff
@@ -167,46 +486,708 @@ process fastqc {
...
+ */
+
+ files_lr_filtered
+    .combine(trimmed_sr_spadeshybrid, by: 0)
+    .set { files_pre_spadeshybrid }
+
+process spadeshybrid {
+    tag "$id"
+    publishDir "${params.outdir}/", mode: 'copy', pattern: "${id}*",
+        saveAs: {filename -> filename.indexOf(".fastq.gz") == -1 ? "Assembly/SPAdesHybrid/$filename" : null}
```
ewels: I think that this should give the expected result - saving only the files you want to the results directory and giving them nice filenames. Without the need for a `cp` or `mv`.

```suggestion
        saveAs: {filename ->
            if (filename.indexOf(".fastq.gz") > 0) "$filename"
            else if (filename.indexOf("assembly_graph_with_scaffolds.gfa") > 0) "${id}_graph.gfa"
            else if (filename.indexOf("scaffolds.fasta") > 0) "${id}_scaffolds.fasta"
            else if (filename.indexOf("contigs.fasta") > 0) "${id}_contigs.fasta"
            else if (filename.indexOf("spades.log") > 0) "${id}_log.txt"
            else null
        }
```

main.nf
```diff
@@ -167,46 +486,708 @@ process fastqc {
...
+ * metaSpades hybrid Assembly
+ */
+
+ files_lr_filtered
+    .combine(trimmed_sr_spadeshybrid, by: 0)
+    .set { files_pre_spadeshybrid }
+
+process spadeshybrid {
+    tag "$id"
+    publishDir "${params.outdir}/", mode: 'copy', pattern: "${id}*",
```
ewels:
```suggestion
    publishDir "${params.outdir}/", mode: 'copy',
```
`pattern` would need to be dropped in combination with the suggestion below.

ewels:
```suggestion
    publishDir "${params.outdir}/Assembly/SPAdesHybrid/", mode: 'copy',
```
In fact, I think you probably want the `/Assembly/SPAdesHybrid/` subfolder? Better to specify this here instead of in the `saveAs` bit.

ewels: You could also specify `$id` in the main `publishDir` bit here and keep all the filenames unchanged if you want to keep things simple. Then you'd just have a different subfolder for each sample.

main.nf
```diff
@@ -167,46 +486,708 @@ process fastqc {
...
+        saveAs: {filename -> filename.indexOf(".fastq.gz") == -1 ? "Assembly/SPAdesHybrid/$filename" : null}
+
+    input:
+    set id, file(lr), file(sr) from files_pre_spadeshybrid  
+
+    output:
+    set id, val("SPAdesHybrid"), file("${id}_graph.gfa") into assembly_graph_spadeshybrid
+    set val("SPAdesHybrid"), val("$id"), file("${id}_scaffolds.fasta") into assembly_spadeshybrid_to_quast
```
ewels: Drop the `${id}_` if you're not doing the `cp` / `mv` of course.

Also it's slightly confusing that in the three channels you're doing:

* id, tool, file
* tool, id, file
* tool, id, file

I usually try to keep the order of arguments consistent, for my own sanity 😆 

Note that we should now be using `path` instead of `file` I think..

main.nf
ewels: Added example code snippets above. Hopefully that all makes sense, let me know if it doesn't!

Alternatively, feel free to ignore me and tell me to make my own PR instead of hassling you..

main.nf
HadrienG: I had gotten your solution to work but then I have file name conflicts in the `busco_plot` process (cause I collect the assemblies there).

I'll go with `move` for now. I'm not saying make your own PR yet, but I'll delay solving this for some ulterior minor release 😄

d4straub: Hm, well I run into that problem before, maybe should have said this. I renamed this output files because of name conflicts downstream. Of course it would be possible to rename the input in the busco process but than the in the report the sample name would be messed up...

ewels: Can you just rename the busco inputs using wildcards? See https://www.nextflow.io/docs/latest/process.html#multiple-input-files

ewels: Oh sorry, you said @d4straub that you need the sample name for the MultiQC report.. Fair enough 🤷‍♂ 😄

Issue #27, opened 2019-12-18
MetaBAT2 binning discards unbinned contigs


d4straub:
I recognized that the current MetaBAT2 binning discards contigs that do not bin with any other contigs. So when looking at the pipeline output any single-contigs are discarded (e.g. for BUSCO or CAT taxa assignments).
Sounds not too shocking, however I have now a dataset with several really large contigs, e.g. a 4.3Mb sequence that does not appear in the output (MultiQC, BUSCO, CAT). This isn't the way the pipeline should handle the data I think. Unbinned contigs need to be reported more obvious. E.g. with metabat's parameter ` --unbinned`.

d4straub:
I actually solved this.

PR #29, opened 2019-12-19
Keep large unbinned contigs for downstream steps

d4straub:
This solves #27


d4straub:
Checks seem to fail because of internet connectivity? Weird.

d4straub:
Ok, BUSCO obviously is now available in v4beta (we are using v3) and the location of the database changed. Hope I fixed this now. Default database is just 8Mb, maybe we could either upgrade to v4 or store the v3 default database in nf-core/test-datasets that we do not loose it again...

HadrienG:
Do you mind if I review/merge this after 1.0.0? I'd like the first release to be done before Christmas 😄 

Concerning Busco I changed the url in dev, I can make a PR to update to v4 (after 1.0.0 ! 😁 )

d4straub:
Fine for me, but I think that's a major flaw. The solution proposed here works fine with test data, but it could be that a pooled bin that is produced right now and forwarded to downstream processes is too big with real data and needs a change.

If you manage to get your release out before I optimize this step here (holiday right now, not working on it atm), go ahead ;)

HadrienG:
A pooled bin being too big seems like a separate issue? Since you are not forwarding the unbinned contigs to downstream processes?

(Happy holidays!)

d4straub:
Still linting errors but coming closer ...

d4straub:
No errors any more, review please.

bin/split_fasta.py
```diff
@@ -0,0 +1,54 @@
...
+input_file = argv[1]
+length_threshold = int(argv[2])
+max_sequences = int(argv[3])
+min_length_to_retain_contig = int(argv[4])
+
+# Base name for file output
+out_base = (os.path.splitext(input_file)[0])
+
+# Read file
+fasta_sequences = SeqIO.parse(open(input_file),'fasta')
```
HadrienG: files should be opened with the `with` context manager. that way the file handles will close automatically at the end of the scope.

HadrienG: i.e.
```python
with open(input_file) as f:
    fasta_sequences = SeqIO.parse(f, 'fasta')
    # rest of the code that needs the file to be open.
```

main.nf
```diff
@@ -906,12 +913,12 @@ process metabat {
     def name = "${assembler}-${sample}"
     """
     jgi_summarize_bam_contig_depths --outputDepth depth.txt ${bam}
-    metabat2 -t "${task.cpus}" -i "${assembly}" -a depth.txt -o "MetaBAT2/${name}" -m ${min_size}
+    metabat2 -t "${task.cpus}" -i "${assembly}" -a depth.txt -o "MetaBAT2/${name}" -m ${min_size} --seed 1 --unbinned
```
HadrienG: shouldn't there rather been no seed by default but an option to fix the seed?

d4straub: My aim was to make sure that every time the workflow runs the same results are created. Therefore I fixed the seed here. But later I realized that at least SPAdes is not producing the identical results, obviously there is also some randomness in there and I haven't seen an option to fix the seed there. So I don't feel that it necessary at all any more.

d4straub: Changed in https://github.com/nf-core/mag/pull/29/commits/17065a9c9ffd51de75c67c00a71f3e276f8050b8

bin/split_fasta.py
d4straub: Right! Changed in https://github.com/nf-core/mag/pull/29/commits/8effb0b97c6d894e046f28b6c243df751e39a261

PR #31, opened 2019-12-21
fix broken reference links in readme

### LINKED ISSUE #30: Links in readme are broken
All the external links in README.md are broken; e.g. https://github.com/OpenGene/fastp links to the non-existent page https://github.com/nf-core/mag/blob/master/fastp.

heuermh:
Fixes #30


HadrienG:
Hi,

Good catch! I've modified the PR for keeping the links inline and rebased to dev, but otherwise it looks good!

PR #49, opened 2020-05-31
Host read removal with Bowtie 2

skrakau:
Here is a suggestion for host read removal using bowtie2. For the host reference sequence either iGenomes or a user specified Fasta reference file can be used.

I also added a MultiQC section for this to display which fraction of reads maps against the host reference and is thus filtered out using a custom content file as suggested by @ewels and @drpatelh (https://github.com/ewels/MultiQC/issues/1199), instead of using the standard Bowtie 2 MultiQC format (with its multiple different mapping categories).  Moreover, I separated the MultiQC FastQC for before and after preprocessing.

If using this bowtie2 based strategy would be fine for you, I can add a test.

## PR checklist
 - [ ] `README.md` is updated


d4straub:
Bowtie2 host removal is fine.
A test and documentation would be great.

skrakau:
Ok thanks, will do!

main.nf
```diff
@@ -459,6 +501,91 @@ process fastp {
...
+                -x ${index[0].getSimpleName()} \
+                -1 "${reads[0]}" -2 "${reads[1]}" \
+                $sensitivity \
+                --un-conc-gz ${name}_host_unmapped_%.fastq.gz \
+                --al-conc-gz ${name}_host_mapped_%.fastq.gz \
+                1> /dev/null \
+                2> ${name}.bowtie2.log
+
+        zcat ${reads[0]} | echo "Read pairs before removal: \$((`wc -l`/4))" >>${name}_remove_host.log
+        zcat ${name}_host_unmapped_1.fastq.gz | echo "Read pairs after removal: \$((`wc -l`/4))" >>${name}_remove_host.log
```
ewels, line 560: Is this just for debugging? I can't see where `${name}_remove_host.log` is used elsewhere.

Might be a bit of an expensive operation if so, `zcat` on a big FastQ file can take quite a while. And I guess we get this info from your bowtie logs MultiQC module anyway?

skrakau: True, I originally copied it from the `remove_phix` process, but it can be removed here.

bin/multiqc_to_custom_tsv.py
```diff
@@ -0,0 +1,114 @@
...
+        yamlFile = os.path.join(MultiQCDataDir,yamlFile)
+        if os.path.exists(yamlFile):
+            MetricsDict = yaml_fields_to_dict(YAMLFile=yamlFile,AppendDict=MetricsDict,FieldMappingList=mappingList)
+            FieldList += [x[0] for x in mappingList]
+        else:
+            print('WARNING: File does not exist: {}'.format(yamlFile))
+
+    if MetricsDict != {}:
+        make_dir(os.path.dirname(OutFile))
+        fout = open(OutFile,'w')
```
HadrienG, line 76: you should use `with` for all file handles, even in "w" mode

main.nf
```diff
@@ -459,6 +501,91 @@ process fastp {
...
+    output:
+    set val(name), file("${name}_host_unmapped*.fastq.gz") into ch_trimmed_reads_host_removed
+    file("${name}.bowtie2.log") into ch_host_removed_log
+    file("${name}_host_mapped*.read_ids.txt") into ch_mapped_read_ids
+
+    when: params.host_fasta || params.host_genome
+
+    script:
+    def sensitivity = params.host_removal_verysensitive ? "--very-sensitive" : "--sensitive"
+    if ( !params.single_end ) {
```
HadrienG, line 548: for avoiding to duplicate most of the process code (sin gle end and paired end), you can define an input parameter, like at https://github.com/nf-core/mag/blob/master/main.nf#L672

bin/multiqc_to_custom_tsv.py
skrakau: Thanks, will change that

main.nf
skrakau: ok, but in this case there are multiple parts in the process code affected: 
1.  `-1 "${reads[0]}" -2 "${reads[1]}"`
2. `--un-conc-gz ${name}_host_unmapped_%.fastq.gz \`
    `--al-conc-gz ${name}_host_mapped_%.fastq.gz \`
3. `zcat ${name}_host_mapped_1.fastq.gz | awk '{if(NR%4==1) print substr(\$0, 2)}' > ${name}_host_mapped_1.read_ids.txt`
   `zcat ${name}_host_mapped_2.fastq.gz | awk '{if(NR%4==1) print substr(\$0, 2)}' > ${name}_host_mapped_2.read_ids.txt`

Maybe in this case it doesn't necessarily get cleaner if solved like this?

skrakau:
OK, I now added a test using the `--host_fasta` option and some brief documentation.

Currently this host removal only works for short reads. If `--host_fasta` or  `--host_genome` is specified in combination with `--manifest`, an error with a corresponding message is returned. Is that OK?

skrakau:
In the MAG `test.config` file `temp_dir` is set to a folder within `outdir`.  So far I didn't do that for `test_host_rm.config`.  Is this necessary for this pipeline, and if yes why?

skrakau:
Hi @ewels or @apeltzer, could one of you tell me by any chance what the purpose of the line `temp_dir = "./tests/tmp_dir"` in the `test.config` files is (it's also used in the ampliseq pipline, but didn't find any documentation)? Still wondering if it is ok to just omit it in the new `conf/test_host_rm.config` file for this PR.

ewels:
Link to line in question:

https://github.com/nf-core/mag/blob/4c2f61cbbb2be56d9a176947a758cdb4d7941910/conf/test.config#L18

Nothing to do with me I'm afraid. [Looks like](https://github.com/nf-core/mag/commit/e4641a3685ea8ebc5ca0332b65d7bcb11ed667b3#diff-bf64226c704a1bd946ee7c05397d3a68) it was added by @HadrienG 

Phil

ewels:
(but I agree, I can't see anything obvious that it is doing, and I suspect that it can be removed from both config files)

skrakau:
OK, thanks @ewels !

d4straub:
> Currently this host removal only works for short reads. If `--host_fasta` or `--host_genome` is specified in combination with `--manifest`, an error with a corresponding message is returned. Is that OK?

I think that's fine. Using proper settings for ONT qc (e.g. `--longreads_length_weight 1`) will remove those that have no matches with Illumina reads, so ONT reads can be filtered that way as well. And SPAdes anyway is using ONT reads only for resolving uncertainties in the assembly graph including gap closure and therefore ONT host reads are relatively unlikely to mess up the assembly, I think.

skrakau:
Ok, will change it then so that the already filtered short reads will be used to filter the long reads.

skrakau:
Hi @d4straub, 

thanks for your input. The following points were added/changed now:

- filtlong uses now the already filtered short reads as a reference (to indirectly filter out long host reads), for this I changed the channel `files_all_raw` to `files_long_raw`and joined this with the filtered short reads 
- accordingly I also changed the order of the processes in the code just for the sake of clarity: first short reads preprocessing, then long reads preprocessing
- if host removal is turned on in combination with long reads, warnings are given out to inform the user that long reads are only filtered indirectly, and informing about the effect of `--longreads_length_weight` (if > 1)
- added some information about this to the docs

I tested this locally for a very basic example, and with `--longreads_length_weight 1` long reads mapping to host sequences were discarded as expected, reads mapping to the 'minigut' sequences not. 

- added missing parameters to the summary

Best,
Sabrina

conf/test_host_rm.config
```diff
@@ -0,0 +1,27 @@
...
+  max_memory = 6.GB
+  max_time = 48.h
+  host_fasta = "https://github.com/nf-core/test-datasets/raw/mag/host_reference/genome.hg38.chr21_10000bp_region.fa"
+  readPaths = [
+    ['test_minigut_hg38host', ['https://github.com/nf-core/test-datasets/raw/mag/test_data/test_minigut_hg38host_R1.fastq.gz', 'https://github.com/nf-core/test-datasets/raw/mag/test_data/test_minigut_hg38host_R2.fastq.gz']],
+    ['test_minigut_sample2_hg38host', ['https://github.com/nf-core/test-datasets/raw/mag/test_data/test_minigut_sample2_hg38host_R1.fastq.gz', 'https://github.com/nf-core/test-datasets/raw/mag/test_data/test_minigut_sample2_hg38host_R2.fastq.gz']]
+  ]
+  centrifuge_db = "https://github.com/nf-core/test-datasets/raw/mag/test_data/minigut_cf.tar.gz"
+  kraken2_db = "https://github.com/nf-core/test-datasets/raw/mag/test_data/minigut_kraken.tgz"
+  skip_krona = true
```
d4straub, line 24: Any particular reason for skipping this in this test?

d4straub:
If I am not mistaken there are 3 tests now:
- Illumina data
- Illumina data + host removal
- Illumina + ONT data
wouldn't it be good to test also 
- Illumina + ONT data + host removal
My reasoning is that there are now some channels that are not tested otherwise. And further changes breaking processes that use these channels might go undetected when the test is missing. What do you think?

conf/test_host_rm.config
skrakau: No, just copied it from `test.config`. If it should not be skipped I guess it would make more sense to turn it on in the main `test.config`, or?

d4straub: Than probably I had a particular reason not to do this but do not remember!
It might be not bad to produce a krona plot in one of the tests, but because I might be the source of the problem here, this shouldn't be your problem.

skrakau:
Ok, I added a corresponding test, but it is more for channel testing, as I didn't add host reads to the long read dataset.

skrakau:
I made the saving of the host read ids optional and added some sorting, since the order of the Bowtie 2 results is not reproducible.

PR #59, opened 2020-06-30
Adjusted handling of MetaBAT2 output files

skrakau:
I introduced a bug by updating Metabat2 to v2.15 in PR #56: this version of MetaBAT2 outputs additional files "\*.lowDepth.fa" and "\*.tooShort.fa", which might be empty and cause problems when passed forward to the process `quast_bins`.

To fix this, and since those files should anyway not be processed further, I moved all the unbinned FASTA files that are smaller than `--min_length_unbinned_contigs` (i.e. \*.tooShort.fa, \*.lowDepth.fa, \*.unbinned.pooled.fa, \*.unbinned.remaining.fa) to a folder `discarded`. However, those files will still be published. Additionally I renamed the file "\*.unbinned.fa.remaining" to "\*.unbinned.remaining.fa" for the sake of consistency.

@d4straub do you think that's OK in this way and the naming of the files ending up in the results folder should stay like this? If yes, some information about this could be added to the `output.md`

## PR checklist
 - [ ] `README.md` is updated

### DIFF
--- a/bin/split_fasta.py
+++ b/bin/split_fasta.py
@@ -51,5 +51,5 @@ for index, row in df.iterrows():
 
 print("write "+out_base+".pooled.fa")
 SeqIO.write(pooled, out_base+".pooled.fa", "fasta")
-print("write "+out_base+".fa.remaining")
-SeqIO.write(remaining, out_base+".fa.remaining", "fasta")
+print("write "+out_base+".remaining.fa")
+SeqIO.write(remaining, out_base+".remaining.fa", "fasta")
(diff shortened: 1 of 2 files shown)


d4straub:
Yes, I think this looks good. I am not sure if anybody needs so many files in the "discarded" folder. The alternative would be to merge these. However, it also doesn't hurt to have an explanation why the contigs do not end up in the final MAGs, i.e. the docs need a definition what contigs end up in what file.

skrakau:
unused FASTA files are now also in the final results located within a `discarded` subfolder.

Still needs to be documented.

PR #62, opened 2020-07-03
Get rid of duplicated outputs to avoid 'Failed to publish' warnings

skrakau:
When running the pipeline there were quite a number of "Failed to publish file" warnings. These originate from output files which are specified multiple time, e.g. because they are directed to multiple output channels or due to the additional use of wildcards for output folders. 

Fixed this so that each file is not specified more than once as output, while using multiple output channels (`...into (channel1, channel2)`). For one channel for which `mode flatten` was used before in the process output, with the goal to emit each bin individually, I changed this by using the `flatMap` operator afterwards. In this way its output can be given out together with the other channels (and the option `mode` is anyway deprecated).

## PR checklist
 - [ ] `README.md` is updated

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -865,8 +865,7 @@ process megahit {
     set val(name), file(reads) from trimmed_reads_megahit
 
     output:
-    set val("MEGAHIT"), val("$name"), file("MEGAHIT/${name}.contigs.fa") into assembly_megahit_to_quast
-    set val("MEGAHIT"), val("$name"), file("MEGAHIT/${name}.contigs.fa") into assembly_megahit_to_metabat
+    set val("MEGAHIT"), val("$name"), file("MEGAHIT/${name}.contigs.fa") into (assembly_megahit_to_quast, assembly_megahit_to_metabat)
     file("MEGAHIT/*.log")
 
     when:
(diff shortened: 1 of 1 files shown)


main.nf
```diff
@@ -1081,6 +1075,10 @@ process busco_download_db {
 }
 
 metabat_bins
+    .flatMap { it -> def list = []
+                     for (c in it[2]){ list << [ it[0], it[1], c] }
+                     list
+    }
```
d4straub, line 1081: This seems so much less readable then "mode flatten", are you sure this is the right way?

skrakau: :D no, unfortunately I am not, but I did not find any other way to achieve this.

So we have in channel `metabat_bins`:
[assembler, sample1, [bin1, bin2, bin3]]
[assembler, sample2, [bin7, bin8]]
...

and need:
[assembler, sample1, bin1]
[assembler, sample1, bin2]
[assembler, sample1, bin3]
[assembler, sample2, bin7]
[assembler, sample2, bin8]
...

@ggabernet do you know by any chance a way how this could be achieved differently?

d4straub: I actually also had a look and no better idea. But this replacement for `mode flatten` seems not like an improvement to me. But of course, if this is the only way to replace it, so be it.

skrakau:
OK, thanks for reviewing! I completely see your point about the readability, just thought getting rid of all the warnings in general would be important, for users but also for further development.

main.nf
ggabernet: Just checked out the `transpose` operator, which should be doing exactly this: https://www.nextflow.io/docs/latest/operator.html#transpose

d4straub: This seems definitely nicer. Good catch!

PR #65, opened 2020-07-10
Added parameters to ensure reproducible results from tools

skrakau:
So far it was not possible to generate reproducible results with the tools MEGAHIT, SPAdes and MetaBAT2 within this pipeline. For MetaBAT2 this could be simply changed by using its seed parameter. I set it to 1 by default and added this as a pipeline parameter `--metabat_rng_seed`.

For MEGAHIT and SPAdes there is unfortunately no seed option. MEGAHIT can only ensure reproducible results when run single-threaded (https://github.com/voutcn/megahit/issues/170). SPAdes on the other hand was designed to generate reproducible results for a given number of threads (https://github.com/ablab/spades/issues/111). 

For this reason I added parameters to fix the number of cpus for those processes, for MEGAHIT to 1 (`--megahit_fix_cpu_1`) and for SPAdes to some fixed number (`--spades_fix_cpus`, `--spadeshybrid_fix_cpus`). For the sake of runtime efficiency, by default those parameters are not set in order to allow MEGAHIT running multi-threaded.

When these parameters are used, it is also ensured that the number of cpus will not be increased for retries. This is important for reproducing previous results but also to generate reproducible results, because otherwise a different number of cpus could be used for different samples. 

I wrote new functions (defined in `nextflow.config`) to set the cpus accordingly for those processes, which are called in the `base.config` file. In theory, those settings can be overwritten by the user with an additional custom config file (`-c`), changing the cpus of the corresponding processes. (I did not find any way how to check for this)  To make sure that only with, for example,  `--spades_fix_cpus` specified cpus are used, this is checked in the actual process script. If the `task.cpus` does not fit to the parameters, an error is returned.

It is also additionally checked at the beginning of the pipeline if the number of cpus is available, i.e. if `--spades_fix_cpus`, `--spadeshybrid_fix_cpus` <= `max_cpus` (instead of allowing that the number of cpus is reduced to `max_cpus`in `check_max()`).

The parameters are added to the summary.

Any feedback welcome :)

## PR checklist
 - [x] Documentation in `docs` is updated
 - [ ] `README.md` is updated

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -86,6 +86,13 @@ First, go to the [nf-core/mag releases page](https://github.com/nf-core/mag/rele
 
 This version number will be logged in reports when you run the pipeline, so that you'll know what you used when you look back in the future.
 
+Additionally, to ensure reproducible results from the individual tools there are a few aspects to be aware of.
+SPAdes is designed to be deterministic for a given number of threads. Thus to reproduce the results, just make sure to use the same number of threads when re-running the pipeline.
+To ensure reproducible results from MEGAHIT, run MEGAHIT single threaded by specifying `cpus = 1` for this process.
+
+MetaBAT2 is run by default with a fixed seed within this pipeline.
+
+
 ## Main arguments
 
 ### `-profile`
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -86,12 +86,10 @@ First, go to the [nf-core/mag releases page](https://github.com/nf-core/mag/rele
 
 This version number will be logged in reports when you run the pipeline, so that you'll know what you used when you look back in the future.
 
-Additionally, to ensure reproducible results from the individual tools there are a few aspects to be aware of.
-SPAdes is designed to be deterministic for a given number of threads. Thus to reproduce the results, just make sure to use the same number of threads when re-running the pipeline.
-To ensure reproducible results from MEGAHIT, run MEGAHIT single threaded by specifying `cpus = 1` for this process.
+Additionally, to enable also reproducible results from the individual assembly tools this pipeline provides extra parameters. SPAdes is designed to be deterministic for a given number of threads. To generate reproducible results set the number of cpus with `--spades_fix_cpus` or `--spadeshybrid_fix_cpus`. This will overwrite the number of cpus specified in the `base.config` file and additionally ensure that it is not increased in case of retries for individual samples. MEGAHIT only generates reproducible results when run single-threaded.
+You can fix this by using the prameter `--megahit_fix_cpu_1`. In both cases, do not specify the number of cpus for these processes in additional custom config files, this would result in an error.
 
-MetaBAT2 is run by default with a fixed seed within this pipeline.
-
+MetaBAT2 is run by default with a fixed seed within this pipeline, thus producing reproducible results.
 
 ## Main arguments
 
@@ -242,6 +240,20 @@ Database for taxonomic binning with kraken2 (default: none). E.g. "<ftp://ftp.cc
 Database for taxonomic classification of metagenome assembled genomes (default: none). E.g. "<tbb.bio.uu.nl/bastiaan/CAT*prepare/CAT_prepare_20190108.tar.gz>"
 The zipped file needs to contain a folder named "\_taxonomy*" and "_CAT_database_" that hold the respective files.
 
+## Assembly options
+
+### `--megahit_fix_cpu_1`
+
+Fix number of CPUs for MEGAHIT to 1. Not increased with retries (default: false). See also [Reproducibility](#reproducibility).
+
+### `--spades_fix_cpus`
+
+Fixed number of CPUs used by SPAdes. Not increased with retries (default: none).
+
+### `--spadeshybrid_fix_cpus`
+
+Fixed number of CPUs used by SPAdes hybrid. Not increased with retries (default: none).
+
 ## Binning options
 
 ### `--min_contig_size`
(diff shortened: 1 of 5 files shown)


docs/usage.md
```diff
@@ -235,6 +241,20 @@ Database for taxonomic binning with kraken2 (default: none). E.g. "<ftp://ftp.cc
...
 
+## Assembly options
+
+### `--megahit_fix_cpu_1`
+
+Fix number of CPUs for MEGAHIT to 1. Not increased with retries (default: false). See also [Reproducibility](#reproducibility).
+
+### `--spades_fix_cpus`
+
+Fixed number of CPUs used by SPAdes. Not increased with retries (default: none).
```
d4straub, line 252: default: false ?

main.nf
```diff
@@ -873,9 +904,12 @@ process megahit {
 
     script:
     def input = params.single_end ? "-r \"${reads}\"" :  "-1 \"${reads[0]}\" -2 \"${reads[1]}\""
-    """
-    megahit -t "${task.cpus}" $input -o MEGAHIT --out-prefix "${name}"
-    """
+    if ( !params.megahit_fix_cpu_1 || task.cpus == 1 )
```
d4straub, line 907: I think this is not doing what it is supposed to do. I'd think what you want is
```
if ( params.megahit_fix_cpu_1 && task.cpus > 1 )
  error [...]
else
  """megahit [...]
```

main.nf
```diff
@@ -906,19 +940,22 @@ process spadeshybrid {
...
-        --pe1-1 ${sr[0]} \
-        --pe1-2 ${sr[1]} \
-        --nanopore ${lr} \
-        -o spades
-    mv spades/assembly_graph_with_scaffolds.gfa ${id}_graph.gfa
-    mv spades/scaffolds.fasta ${id}_scaffolds.fasta
-    mv spades/contigs.fasta ${id}_contigs.fasta
-    mv spades/spades.log ${id}_log.txt
-    """
+    if ( !params.spadeshyrid_fix_cpus || task.cpus == params.spadeshybrid_fix_cpus )
```
d4straub, line 943: Same as above.

main.nf
```diff
@@ -928,31 +965,35 @@ process spades {
...
-        --memory "$maxmem" \
-        --pe1-1 ${sr[0]} \
-        --pe1-2 ${sr[1]} \
-        -o spades
-    mv spades/assembly_graph_with_scaffolds.gfa ${id}_graph.gfa
-    mv spades/scaffolds.fasta ${id}_scaffolds.fasta
-    mv spades/contigs.fasta ${id}_contigs.fasta
-    mv spades/spades.log ${id}_log.txt
-    """
+    if ( !params.spades_fix_cpus || task.cpus == params.spades_fix_cpus )
```
d4straub, line 982: Same as above.

main.nf
```diff
@@ -928,31 +965,35 @@ process spades {
...
     input:
-    set id, file(sr) from trimmed_reads_spades  
+    set id, file(sr) from trimmed_reads_spades
 
     output:
     set id, val("SPAdes"), file("${id}_graph.gfa") into assembly_graph_spades
     set val("SPAdes"), val("$id"), file("${id}_scaffolds.fasta") into (assembly_spades_to_quast, assembly_spades_to_metabat)
     file("${id}_contigs.fasta")
     file("${id}_log.txt")
+    val("${task.cpus}") into ch_spades_cpus
```
d4straub, line 975: cant find where `ch_spades_cpus` is used. Probably this is supposed to be reported somewhere?

main.nf
```diff
@@ -800,6 +830,7 @@ process kraken2 {
     output:
     set val("kraken2"), val(name), file("results.krona") into kraken2_to_krona
     file("kraken2_report.txt")
+    val("${task.cpus}") into ch_kraken2_cpus
```
d4straub, line 833: Can't find where this is used `ch_kraken2_cpus`. Is that need for reproducibility as well?

main.nf
skrakau: ups sorry, this should have been deleted. Thanks!

docs/usage.md
skrakau: Maybe, wasn't sure because in the end it is an `int` parameter.

skrakau: thinking about it, I would still opt for "none" to distinguish it from boolean parameters.

main.nf
skrakau: well spotted :)

main.nf
skrakau: isn't that the same?

d4straub: Probably I misunderstood your aim here, but I thought you want to run megahit only when `params.megahit_fix_cpu_1=true` and `task.cpus=1` or when `params.megahit_fix_cpu_1=false` and `task.cpus>1`. In this case it does make a difference.


I tested with:
```
println "with &&:"
if ( megahit_fix_cpu_1 && cpus > 1 )
  println "error"
else
  println "megahit $cpus"
  
println "with ||:"
if ( !params.megahit_fix_cpu_1 || task.cpus == 1 )
  println "megahit $cpus"
else
  println "error"
```
with parameters:
```
megahit_fix_cpu_1=false
cpus=5
```
result:
```
with &&:
megahit 5
with ||:
megahit 5
```
with parameters:
```
megahit_fix_cpu_1=true
cpus=5
```
result:
```
with &&:
error
with ||:
megahit 5
```
--> here was the difference. And I thought that you aim for "error" here instead of running megahit.

with parameters:
```
megahit_fix_cpu_1=true
cpus=1
```
result:
```
with &&:
megahit 1
with ||:
megahit 1
```

docs/usage.md
d4straub: Makes sense, agreed.

main.nf
skrakau: Yes, indeed I aim for an error in this case. In your test with "||" you wrote `task.cpus` instead of `cpus`. Could you try changing that?

d4straub: Alright, sorry, my mistake!

Issue #77, opened 2020-07-17
BUSCO update 


skrakau:
BUSCO v3.0.2 is currently failing on some datasets, among others on the `test` profile when `scratch = false`. Somehow the results and thrown errors differ between `scratch = true` and `scratch = false` (on CFC), which we cannot not explain currently.  In the past such errors were ignored, which was changed now (see https://github.com/nf-core/mag/issues/68 and https://github.com/nf-core/mag/pull/72).
Moreover it seems that if no tblastn hits are found, this causes an error and is not handled properly.

To achieve more control, update BUSCO to v4.0.6 and handle the case of no tblastn hits. Use a parameter to pass over the path to an already downloaded db (test if this works with `--offline`), the name of the db for automatic download or the auto-lineage parameter

However, currently there is an issue with the download of the BUSCO databases: https://busco.ezlab.org/frames/bact.htm.
See also https://gitlab.com/ezlab/busco/-/issues/293. So I need to wait until this works again, to test this.

Additionally, for offline use, one can also download the whole dataset https://busco-data.ezlab.org/v4/data/ and add the path to the custom config file. I think this should work both with `--lineage_dataset bacteria_odb10` and `--auto-lineage`

d4straub:
If all possibilities would be implemented, I'd think the most requested cases would be `--lineage_dataset bacteria_odb10` >> `--auto-lineage` > `--lineage_dataset archea_odb10` > anything else. This is because usually the majority of genomes are bacteria and BUSCO evaluation results are best comparable if all bins are evaluated with the same reference data. Therefore, only making a single db available for now would be sufficient, I think. But the `--auto-lineage` option looks so tempting ;)

skrakau:
**UPDATE**: 

With BUSCO version 4.0.6 there are frequent, non-reproducible errors occurring, caused by a replace("faa", "fna") function corrupting nextflow filenames that contain the substring "faa" in their hash id. I prepared a fix for BUSCO (https://gitlab.com/ezlab/busco/-/issues/305). Waiting currently.

We need a new BUSCO release before preparing pipeline release.

ropolomx:
Hi @skrakau . Are you referring to this type of error? I am getting a lot of these with BUSCO version 3.0.2when running mag with revision: 8586c49436 [dev]

```
Aug-26 20:41:52.612 [Task monitor] ERROR nextflow.processor.TaskProcessor - Error executing process > 'busco (MEGAHIT-SRR9030455.61.fa)'

[...]

PR #81, opened 2020-07-30
Update to BUSCO 4.0.6 

skrakau:
Here is an update to BUSCO 4.0.6 (v4.1.2 is in conflict with the MAG python version 3.6.7). It contains only the basic functionality so far, without the automatic lineage selection and without automatic DB download which is provided since BUSCO 4 (which would require extra error handling among among others with the current BUSCO version).

I renamed the variable ${assembly} to ${bin} within the busco processes.

Additionally I changed the `busco_plot` process and the channel logic, so that the process now runs on each `assembler`-`sample` combination (instead of calling the busco script `generate_plot.py` for each combination within the process). This also fixes the grouping of BUSCO summaries for the plot, which was wrong in case one sample name was the prefix of another one. 

For the final busco summary for all bins I created an extra process.

Since current versions of the BUSCO `generate_plot.py` (v4.0.6-4.1.2) can not handle additional dots within the summary file names, I unfortunately had to replace also dots with underscores in the bin names, which will also appear as such in the BUSCO plots.

## PR checklist
 - [ ] `README.md` is updated

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -84,7 +84,7 @@ def helpMessage() {
     Bin quality check:
       --skip_busco [bool]                   Disable bin QC with BUSCO (default: false)
       --busco_reference [file]              Download path for BUSCO database, available databases are listed here: https://busco.ezlab.org/
-                                            (default: https://busco-archive.ezlab.org/v3/datasets/bacteria_odb9.tar.gz)
+                                            (default: https://busco-data.ezlab.org/v4/data/lineages/bacteria_odb10.2019-06-26.tar.gz)
 
     Reproducibility options:
       --megahit_fix_cpu_1 [bool]            Fix number of CPUs for MEGAHIT to 1. Not increased with retries (default: false)
@@ -154,9 +154,9 @@ if (params.megahit_fix_cpu_1 || params.spades_fix_cpus || params.spadeshybrid_fi
 if(!params.skip_busco){
     Channel
         .fromPath( "${params.busco_reference}", checkIfExists: true )
-        .set { file_busco_db }
+        .set { ch_file_busco_db }
 } else {
-    file_busco_db = Channel.from()
+    ch_file_busco_db = Channel.from()
 }
 
 if(params.centrifuge_db){
@@ -402,7 +402,7 @@ process get_software_versions {
     porechop --version > v_porechop.txt
     NanoLyse --version > v_nanolyse.txt
     spades.py --version > v_spades.txt
-    run_BUSCO.py --version > v_busco.txt
+    busco --version > v_busco.txt
     centrifuge --version > v_centrifuge.txt
     kraken2 -v > v_kraken2.txt
     CAT -v > v_cat.txt
@@ -1117,10 +1117,10 @@ process busco_download_db {
     tag "${database.baseName}"
 
     input:
-    file(database) from file_busco_db
+    file(database) from ch_file_busco_db
 
     output:
-    set val("${database.toString().replace(".tar.gz", "")}"), file("buscodb/*") into busco_db
+    set val("${database.toString().replace(".tar.gz", "")}"), file("buscodb/*") into ch_busco_db
 
     script:
     """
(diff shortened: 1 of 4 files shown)


skrakau:
And importantly it solves https://github.com/nf-core/mag/issues/77.

main.nf
```diff
@@ -1131,112 +1131,119 @@ process busco_download_db {
...
 
     script:
-    def assemblersampleunique = assemblersample.unique()
+    def name = "${assembler}-${sample}"
     """
-    #for each assembler and sample:
-    assemblersample=\$(echo \"$assemblersampleunique\" | sed 's/[][]//g')
-    IFS=', ' read -r -a assemblersamples <<< \"\$assemblersample\"
+    # replace dots in bin names within summary file names with underscores
+    # (currently, v4.1.2, generate_plot.py does not allow further dots)
```
d4straub, line 1217: Its v4.0.6, isnt it?

main.nf
```diff
@@ -1131,112 +1131,119 @@ process busco_download_db {
...
+    cp BUSCO/short_summary.specific.${db}.BUSCO.txt short_summary.specific.${db}.${bin}.txt
+
+    for f in BUSCO/run_${db}/busco_sequences/single_copy_busco_sequences/*faa; do
+        [ -e "\$f" ] && cat BUSCO/run_${db}/busco_sequences/single_copy_busco_sequences/*faa >${bin}_buscos.faa
+        break
+    done
+    for f in BUSCO/run_${db}/busco_sequences/single_copy_busco_sequences/*fna; do
+        [ -e "\$f" ] && cat BUSCO/run_${db}/busco_sequences/single_copy_busco_sequences/*fna >${bin}_buscos.fna
+        break
+    done
```
d4straub, line 1190: You removed the failsave `|| touch ${assembly}_buscos.fna`, does that mean its always there? In this case `[ -e "\$f" ] && ` might be not needed as well? I am unsure here.

main.nf
```diff
@@ -1131,112 +1131,119 @@ process busco_download_db {
...
+    mkdir dataset
+    mv ${db} dataset/
+
+    busco \
+        --mode genome \
+        --in ${bin} \
+        --lineage_dataset dataset/${db} \
+        --config \${config_file} \
+        --cpu "${task.cpus}" \
+        --out "BUSCO" > ${bin}_busco.log
```
d4straub, line 1179: `--blast_single_core \` in busco3 was added to prevent blast problems when using multi cores, is that not a risk any more?

main.nf
```diff
@@ -1131,112 +1131,119 @@ process busco_download_db {
...
     output:
-    file("short_summary_${assembly}.txt") into (busco_summary_to_multiqc, busco_summary_to_plot)
-    val("$assembler-$sample") into busco_assembler_sample_to_plot
-    file("${assembly}_busco_log.txt")
-    file("${assembly}_buscos.faa")
-    file("${assembly}_buscos.fna")
+    set val(assembler), val(sample), file("short_summary.specific.${db}.${bin}.txt") into (ch_busco_multiqc, ch_busco_to_summary, ch_busco_plot)
+    file("${bin}_busco.log")
+    file("${bin}_buscos.faa") optional true
+    file("${bin}_buscos.fna") optional true
```
d4straub, line 1151: when we are already thinking about making the result folder smaller in size, why not gzip these too? These files might be not large, however there might be many bins. This is optional obviously.

main.nf
```diff
@@ -1131,112 +1131,119 @@ process busco_download_db {
...
+    set val(assembler), val(sample), file(bin), val(db_name), file(db) from ch_metabat_db_busco
+ 
     output:
-    file("short_summary_${assembly}.txt") into (busco_summary_to_multiqc, busco_summary_to_plot)
-    val("$assembler-$sample") into busco_assembler_sample_to_plot
-    file("${assembly}_busco_log.txt")
-    file("${assembly}_buscos.faa")
-    file("${assembly}_buscos.fna")
+    set val(assembler), val(sample), file("short_summary.specific.${db}.${bin}.txt") into (ch_busco_multiqc, ch_busco_to_summary, ch_busco_plot)
+    file("${bin}_busco.log")
```
d4straub, line 1149: At least this filename changed, probably the output.md needs an update too?

main.nf
skrakau: I added `optional true` to the output, thought it might be nicer than having empty files in the results folder. 
`[ -e "\$f" ]`  is still needed, as it could be that the file does not exist, and than `cat` should not be applied

d4straub: Alright, sounds good.

main.nf
skrakau: yeah, that should be solved, it uses a blast version where it works

main.nf
skrakau: The summary output and also the plots etc. are yet listed, I will add them. For "${assembly}_buscos.faa" just the variable name changed, not that actual file name.

main.nf
d4straub: Great!

main.nf
d4straub: I meant `_busco_log.txt` -> `_busco.log`, and yes, thats nit-picky.

d4straub:
Hm, test_hybrid failed when busco found no relevant genes (Prodigal found no genes matching ref database) and therefore `short_summary` wasn't produced.

and also: 
```
WARN: Access to undefined parameter `spadeshyrid_fix_cpus` -- Initialise it to a default value eg. `params.spadeshyrid_fix_cpus = some_value`
```

skrakau:
yeah, will have a look

main.nf
skrakau: true (just also holds for newer version), will adjust

skrakau:
I can not reproduce the `test_hybrid_host_rm`  error, also not with nextflow 19.10.0 for which the error occurred (the other test run through). Need to dig deeper into it.

skrakau:
Will close this, since there remains the issue with BUSCO (https://gitlab.com/ezlab/busco/-/issues/305). Will create a new PR with a newer BUSCO version when this is fixed.

PR #116, opened 2020-10-07
Add Zenodo DOI to README

skrakau:
add Zenodo DOI.

## PR checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 [![install with bioconda](https://img.shields.io/badge/install%20with-bioconda-brightgreen.svg)](https://bioconda.github.io/)
 [![Docker](https://img.shields.io/docker/automated/nfcore/mag.svg)](https://hub.docker.com/r/nfcore/mag)
-[![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.4068515.svg)](https://doi.org/10.5281/zenodo.4068515)
+[![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.3589527.svg)](https://doi.org/10.5281/zenodo.3589527)
 [![Get help on Slack](http://img.shields.io/badge/slack-nf--core%20%23mag-4A154B?logo=slack)](https://nfcore.slack.com/channels/mag)
 
 ## Introduction
@@ -77,7 +77,7 @@ For further information or help, don't hesitate to get in touch on the [Slack `#
 
 ## Citation
 
-If you use nf-core/mag for your analysis, please cite it using the following doi: [10.5281/zenodo.4068515](https://doi.org/10.5281/zenodo.4068515)
+If you use nf-core/mag for your analysis, please cite it using the following doi: [10.5281/zenodo.3589527](https://doi.org/10.5281/zenodo.3589527)
 
 You can cite the `nf-core` publication as follows:
 


apeltzer (approved):

README.md
```diff
@@ -8,6 +8,7 @@
 
 [![install with bioconda](https://img.shields.io/badge/install%20with-bioconda-brightgreen.svg)](https://bioconda.github.io/)
 [![Docker](https://img.shields.io/docker/automated/nfcore/mag.svg)](https://hub.docker.com/r/nfcore/mag)
+[![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.4068515.svg)](https://doi.org/10.5281/zenodo.4068515)
```
ggabernet, line 11:
```suggestion
[![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.3589527.svg)](https://doi.org/10.5281/zenodo.3589527)
```

ggabernet (changes requested):
Hi Sabrina, we use the DOI that can match to all versions of the pipeline so that you don't need to update this every time that you do a release. Then it corresponds to the latest release.

README.md
MaxUlysse: Both actually resolve to the latest zenodo record: https://zenodo.org/record/4068515

MaxUlysse: but yeah, the second one should give out all the versions, so that's the one we want

README.md
```diff
@@ -76,8 +77,7 @@ For further information or help, don't hesitate to get in touch on the [Slack `#
 
 ## Citation
 
-<!-- TODO nf-core: Add citation for pipeline after first release. Uncomment lines below and update Zenodo doi. -->
-<!-- If you use  nf-core/mag for your analysis, please cite it using the following doi: [10.5281/zenodo.XXXXXX](https://doi.org/10.5281/zenodo.XXXXXX) -->
+If you use nf-core/mag for your analysis, please cite it using the following doi: [10.5281/zenodo.4068515](https://doi.org/10.5281/zenodo.4068515)
```
MaxUlysse, line 80:
```suggestion
If you use nf-core/mag for your analysis, please cite it using the following doi: [10.5281/zenodo.3589527](https://doi.org/10.5281/zenodo.3589527)
```

skrakau: Thanks @MaxUlysse , but I thought one would want here the DOI for the latest release, and not the generic one for all releases? Because this would not be resolved for the displayed DOI. Only when clicking on the link the user could figure out the DOI for the specific version.

skrakau:
@apeltzer how did you get the Zenodo badge link for the eager and bcellmagic pipelines, which automatically shows the DOI for the latest release (and not the generic one)? I can not find it on Zenodo...

README.md
skrakau: .. ok, maybe not for the latest, but for the actual release, as otherwise the READMEs for older version would be wrong. But still, if someone uses the pipeline, wouldn't it be encouraged to cite the actual version?

apeltzer:
You login, click on the `mag` entry  in your lsit of github repositories and the text above explains it a bit - if you want a specific release, you just need to click on the DOI and take the one that is there available. If you want the "latest" always, simply click on the DOI and use the one that is mentioned below: 

image

On the next page after clicking on the DOI:

image

Which means you can create your own badge now by replacing the URLs accordingly:

[DOI](https://doi.org/10.5281/zenodo.3589527)

README.md
apeltzer: You're technically right that this will always resolve to the latest version, e.g. if a user uses a version 1.0.0 at some point, this URL will point them to 1.1.0 (or whatever is the latest release for the pipeline) - which is going to be wrong at some point. However, as the DOI / badge above also links to an overview page that makes it easy via the navigation to get the version the user wants to obtain, I think its fine to use the latest url everywhere in order to make your life easier ;-)

ggabernet:
Ahh mistery solved! I thought that the one to be used was the one in `Cite all versions`

apeltzer:
I am confused everytime I look at that interface though - not really perfectly well clear and could benefit from an overview figure that explains all these 😆

skrakau:
Thanks @apeltzer ! I am still confused, as `10.5281/zenodo.3589527` is the one for all versions (while for bcellmagic and eager, the badge automatically shows the DOI for the latest version, which would be `10.5281/zenodo.4068515` for mag). 

But never mind, sounds as if anyway you agree to use the DOI for all version, so I will just use that :) (one could just adjust the text in the template accordingly, so that it does not seem that the user should use the displayed DOI and instead should look it up)

skrakau:
Could someone with commit rights for the dev branch commit the suggestions from above?

MaxUlysse (approved):

MaxUlysse:
> Could someone with commit rights for the dev branch commit the suggestions from above?

Done

skrakau:
Thank you @MaxUlysse !

ggabernet (approved):
Great!

PR #118, opened 2020-10-14
Fix seaborn version to avoid error with nanoplot

skrakau:
On some long read data, I got a `NanoPlot` error:

```
raise AttributeError(f"{type(self).__name__!r} object "
AttributeError: 'PathCollection' object has no property 'stat_func'
```
See alos https://github.com/wdecoster/NanoPlot/issues/201.

Seems there was recently a new `seaborn` version 0.11.0 released, which can cause problems with `NanoPlot`. They fixed the seaborn version in the current bioconda recipe. But since we use an older NanoPlot version, we need to fix this manually within the `environment.yml`.
 

## PR checklist

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -26,6 +26,7 @@ dependencies:
   - conda-forge::r-markdown=1.0
   - conda-forge::r-ggplot2=3.2.0
   - bioconda::nanoplot=1.26.3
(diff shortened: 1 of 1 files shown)


d4straub (approved):
LGTM!

Issue #131, opened 2020-11-13
Running the pipeline with singularity


ramakrishnas:
I had success running the test but when I tried to run the pipeline with my data it is not working and gives the message 
-[nf-core/mag] Pipeline completed with errors-

If some can take a look at my command below and let me know where I'm doing wrong that will be helpful.

nextflow run nf-core/mag -r 1.1.0 --input '/Users/rsompallae/*_R{1,2}.fastq' -profile singularity -with-singularity /Users/rsompallae/biotools/nfcore-mag-1.1.0.sif -c nfcore_ui_hpc_2.config --busco_reference /Users/rsompallae/biotools/bacteria_odb10.2020-03-06.tar.gz

Thanks
Rama

d4straub:
Hi, glad you got mag running. 
It would be extremely helpful to have a little more information, i.e. the error message. What did the pipeline complain about?

ramakrishnas:
based on the log file (.nextflow.log) I see following error. 

Nov-16 13:46:24.338 [main] WARN  nextflow.script.ScriptBinding - Access to undefined parameter `input_paths` -- Initialise it to a default value eg. `params.input_paths =
 some_value`
Nov-16 13:46:24.342 [main] WARN  nextflow.script.ScriptBinding - Access to undefined parameter `reads` -- Initialise it to a default value eg. `params.reads = some_value`
Nov-16 13:46:24.429 [PathVisitor-1] ERROR nextflow.Channel - null
java.lang.NullPointerException: null
        at nextflow.file.FileHelper.isGlobAllowed(FileHelper.groovy:233)
        at nextflow.file.PathVisitor.applyGlobPattern0(PathVisitor.groovy:91)
        at nextflow.file.PathVisitor.apply(PathVisitor.groovy:71)
        at nextflow.file.PathVisitor$_applyAsync_closure1.doCall(PathVisitor.groovy:77)
        at nextflow.file.PathVisitor$_applyAsync_closure1.call(PathVisitor.groovy)
        at groovy.lang.Closure.run(Closure.java:486)
        at java.util.concurrent.CompletableFuture.uniRun(CompletableFuture.java:719)
        at java.util.concurrent.CompletableFuture$UniRun.tryFire(CompletableFuture.java:701)
        at java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:456)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[...]

PR #133, opened 2020-11-18
Fix processing of input parameter

skrakau:
Fixed issue https://github.com/nf-core/mag/issues/131: internally still the deprecated `params.reads` where processed instead of `params.input` (sorry! must have overseen something when merging the template).

To avoid future confusions and avoid the warning 
"WARN: Access to undefined parameter `input_paths` -- Initialise it to a default value eg. `params.input_paths = some_value`" 
for users when not running test profiles, I also initialised the `input_paths` variable again (even though it was removed from the template).

## PR checklist


d4straub (approved):
LGTM!

PR #143, opened 2020-12-15
Changed 'manifest' format and merged into 'input' parameter

skrakau:
Changed manifest input format:
- `sampleId, lr, sr1, sr2` -> `sampleId, groupId, sr1, sr2 [, lr]`
- added groupId (not used yet)
- lr: long read file optional!

Needs to be TSV format now and has to be handed over also via `--input` parameter now.

I made a PR for the testate s well https://github.com/nf-core/test-datasets/pull/208. The current CI tests will fail.

If you agree to this, the `S3` manifest file needs to be updated as well.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## v1.2.0dev - [date]
 
+### `Changed`
+
+- [#143](https://github.com/nf-core/mag/pull/143) - Changed format of manifest file, has to be handed over via `--input` parameter as well now
+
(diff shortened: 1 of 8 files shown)


d4straub (approved):
Looks good, two change requests:
- removing `--manifest` so sudden is probably not so good, could you instead (1) raise a warning when `--manifest` is specified saying that it is deprecated (2) forward `--manifest` into `--input`
- add to the main.nf help & docu explicitely that the "tsv" extension is required for the mnaifest file.

skrakau:
Decided to return an error message when deprecated `--manifest` parameter is used and exit, since also the format changed.

PR #145, opened 2020-12-16
Rename FASTQ files and check sample ids

skrakau:
- rename all short read FASTQ files to ensure consistent names in reports from FastQC and allow non-unique read file base names
- added check for unique sample IDs, return error otherwise

## PR checklist

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -164,11 +164,11 @@ if(hasExtension(params.input, "tsv")){
                     exit 1, "Input TSV file contains row with ${row.size()} column(s). Expects 4 or 5."
                 }
             }
-        .into { ch_all_files_sr; ch_all_files_lr }
+        .into { ch_sample_validate; ch_all_files_sr; ch_all_files_lr }
     // prepare input for preprocessing
     ch_all_files_sr
         .map { row -> [ row[0], [ row[2], row[3] ] ] }
-        .into { ch_raw_short_reads }
+        .set { ch_raw_short_reads }
     ch_all_files_lr
         .map { row -> if (row.size() == 5) [ row[0], row[4] ] }
         .set { ch_raw_long_reads }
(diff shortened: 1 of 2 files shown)


main.nf
```diff
@@ -444,6 +450,28 @@ process get_software_versions {
...
+    set val(name), file(reads) from ch_raw_short_reads
+
+    output:
+    set val(name), path("*.fastq.gz", includeInputs: true) into (ch_raw_short_reads_fastqc, ch_raw_short_reads_fastp)
+
+    script:
+    if ( !params.single_end )
+        """
+        [ -f "${name}_R1.fastq.gz" ] || ln -s "${reads[0]}" "${name}_R1.fastq.gz"
+        [ -f "${name}_R2.fastq.gz" ] || ln -s "${reads[1]}" "${name}_R2.fastq.gz"
```
d4straub, line 467: To me this is not readable. What does this do? 

Guessing: If file  "${name}_R1.fastq.gz" doesnt exist than link reads[0} to "${name}_R1.fastq.gz"
If yes, could you write that in an easier understandable way? With an in if such? Like this its short but hard to read, in my opinion.
How does this interact with `includeInputs: true`, havent seen that one before, I guess it is important if the above link isnt created. But what happens is the link is actually created? Isnt than reads[0] also added to the output channel, provided it follows *.fastq.gz as well?

d4straub (approved):
LGTM!

main.nf
skrakau: Oh true, I forgot to add the corresponding glob pattern, thanks for reviewing!

`includeInputs` is important, because by default input files will not be used for the output (https://www.nextflow.io/docs/latest/process.html#output-path), even if the name matches. 
I didn't want to rename the input files, that's why I did it this way.

PR #149, opened 2021-02-01
Allow custom assembly options and export assembly QC

d4straub:
- added exit code 247 to bowtie2 retry, 247 is when it runs out of memory while index building
- added two parameters (`--spades_options` & `--megahit_options`) to allow custom SPAdes and MEGAHIT options, this is required to adjust e.g. k-mers or presets for more challenging assemblies
- export bowtie2 mapping statistics as another important QC. This is nicely integrated into the MultiQC report when assembling individual samples but not when assembling groups. Because it is here impossible to match assembly to sample and would create a huge mess.
- docs are updated

edit: Just figured out that MEGAHIT told `Process `megahit (groupall)` terminated with an error exit status (250)` when running into RAM issues. That should be added as well. Unfortunately, time limits resulted in an empty error code, very unfortunate.
edit2: Re-try MEGAHIT when exiting with code 250


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [#146](https://github.com/nf-core/mag/pull/146) - Add `--coassemble_group` parameter to allow group-wise co-assembly
 - [#146](https://github.com/nf-core/mag/pull/146) - Add `--binning_map_mode` parameter allowing different mapping strategies for binning (`all`, `group`, `own`)
+- Add two new parameters to allow custom SPAdes and MEGAHIT options (`--spades_options` and `--megahit_options`)
```
skrakau, line 12:
```suggestion
- [#149](https://github.com/nf-core/mag/pull/149) - Add two new parameters to allow custom SPAdes and MEGAHIT options (`--spades_options` and `--megahit_options`)
```

main.nf
```diff
@@ -62,6 +62,8 @@ def helpMessage() {
 
     Assembly:
       --coassemble_group [bool]             Co-assemble samples within one group, instead of assembling each sample separately (default: false)
+      --spades_options [string]             Additional custom options for SPAdes, e.g. "-k 21,33,55,77" or from https://github.com/ablab/spades#advanced-options
```
skrakau, line 65: important: other than `-t`, `-m` `-o` and `-out-prefix` !

skrakau: ups, this particular options were meant for megahit

d4straub: True, I'll add that info for megahit and spades

main.nf
```diff
@@ -1236,28 +1244,36 @@ if (params.binning_map_mode == 'all'){
...
     !params.skip_binning
 
     script:
-    def name = "${assembler}-${name}-${sampleToMap}"
+    def NAME = "${assembler}-${name}-${sampleToMap}"
     def input = params.single_end ? "-U \"${reads}\"" :  "-1 \"${reads[0]}\" -2 \"${reads[1]}\""
         """
         bowtie2-build --threads "${task.cpus}" "${assembly}" ref
-        bowtie2 -p "${task.cpus}" -x ref $input | \
+        bowtie2 -p "${task.cpus}" -x ref $input 2>"${NAME}.bowtie2-log" | \
```
skrakau, line 1265:
```suggestion
        bowtie2 -p "${task.cpus}" -x ref $input 2>"${NAME}.bowtie2.log" | \
```

main.nf
```diff
@@ -1236,28 +1244,36 @@ if (params.binning_map_mode == 'all'){
...
     tag "$assembler-$name"
+    publishDir "${params.outdir}/Assembly/${assembler}/${name}_QC", mode: params.publish_dir_mode,
+        saveAs: {filename -> filename.indexOf(".bowtie2-log") > 0 ? filename : null}
 
     input:
     set val(assembler), val(name), val(group), file(assembly), val(sampleToMap), val(sampleGroup), file(reads) from ch_bowtie2_input
 
     output:
     set val(assembler), val(name), file("${assembler}-${name}-${sampleToMap}.bam"), file("${assembler}-${name}-${sampleToMap}.bam.bai") into ch_assembly_mapping_for_metabat
+    set val(assembler), val(name), val(sampleToMap), file("*.bowtie2-log") into ch_assembly_mapping_stats
```
skrakau, line 1255:
```suggestion
    set val(assembler), val(name), val(sampleToMap), file("*.bowtie2.log") into ch_assembly_mapping_stats
```

skrakau: the file extension would be cleaner I think

main.nf
```diff
@@ -1236,28 +1244,36 @@ if (params.binning_map_mode == 'all'){
 
 process bowtie2 {
     tag "$assembler-$name"
+    publishDir "${params.outdir}/Assembly/${assembler}/${name}_QC", mode: params.publish_dir_mode,
+        saveAs: {filename -> filename.indexOf(".bowtie2-log") > 0 ? filename : null}
```
skrakau, line 1248:
```suggestion
        saveAs: {filename -> filename.indexOf(".bowtie2.log") > 0 ? filename : null}
```

main.nf
```diff
@@ -1236,28 +1244,36 @@ if (params.binning_map_mode == 'all'){
...
     output:
     set val(assembler), val(name), file("${assembler}-${name}-${sampleToMap}.bam"), file("${assembler}-${name}-${sampleToMap}.bam.bai") into ch_assembly_mapping_for_metabat
+    set val(assembler), val(name), val(sampleToMap), file("*.bowtie2-log") into ch_assembly_mapping_stats
 
     when:
     !params.skip_binning
 
     script:
-    def name = "${assembler}-${name}-${sampleToMap}"
+    def NAME = "${assembler}-${name}-${sampleToMap}"
```
skrakau, line 1261: good point, that was not nice! But I would suggest using a variable name in lower case, e.g. `map_name`, otherwise it seems as if there is another reason why is is suddenly in upper case.

main.nf
d4straub: This change might skew MultiQC reporting, have to test first.

main.nf
d4straub: Agreed, but than it might appear in MultiQC as separate line, I have to test whether it uses everything before a "." as sample name or only before the last ".". I would prefer a less clean file name than a messed up reporting.

main.nf
```diff
@@ -1236,28 +1244,36 @@ if (params.binning_map_mode == 'all'){
...
-        samtools index "${name}.bam"
+            samtools sort -@ "${task.cpus}" -o "${NAME}.bam"
+        samtools index "${NAME}.bam"
+        if [ ${NAME} = "${assembler}-${name}-${name}" ] ; then
+            mv "${NAME}.bowtie2-log" "${assembler}-${name}.bowtie2-log"
+        fi
         """
 }
 
+ch_assembly_mapping_stats_for_multiqc = ch_assembly_mapping_stats.map { assembler, name, sampleToMap ,file -> if (name == sampleToMap) {return [ file ]} }
```
skrakau, line 1275: So this would only report mappings, if no co-assembly is done and, additionally, only for the sample the assembly is created from. 

In case of `--binning_map_mode  all` this would mean that only a subset of the used mappings is reported, right?  This seems a bit confusing to me.

main.nf
```diff
@@ -1575,6 +1591,7 @@ process multiqc {
     file (fastqc_trimmed:'fastqc/*') from ch_fastqc_results_trimmed.collect().ifEmpty([])
     file (host_removal) from ch_host_removed_log.collect().ifEmpty([])
     file ('quast*/*') from ch_quast_results.collect().ifEmpty([])
+    file ('bowtie2log/*') from ch_assembly_mapping_stats_for_multiqc.collect().ifEmpty([])
```
skrakau, line 1594: Is this described within the MulitQC report, where these mappings actually come from? I didn't see any changes for this

main.nf
```diff
@@ -1236,28 +1244,36 @@ if (params.binning_map_mode == 'all'){
 
 process bowtie2 {
     tag "$assembler-$name"
+    publishDir "${params.outdir}/Assembly/${assembler}/${name}_QC", mode: params.publish_dir_mode,
```
skrakau, line 1247: this is not added to the `output.md` yet, if I see it right

main.nf
d4straub: All correct, I have written this in the opening post.

> This is nicely integrated into the MultiQC report when assembling individual samples but not when assembling groups. Because it is here impossible to match assembly to sample and would create a huge mess.

As it is with this setting, the MultiQC report shows an additional column with "% Aligned" for each assembly, how I would expect. I do not care if other samples map at this point.
If this isnt done, the MultiQC report shows (number of sample)² rows with only entries in "% Aligned", completely unacceptable in my opinion.

In case of --binning_map_mode all, this isnt shown at all because it makes no sense at all to show the mapping of just one sample, but I strongly dislike adding almost empty row to the report. I think we should eventually add this to the summary table "quast_and_busco_summary.tsv"

main.nf
d4straub: Yes, its in the same row. Perfectly aligned.

main.nf
d4straub: Done

skrakau:
Regarding the time limit and the missing exit code, can it be a Nextflow or slurm problem if this is not reported properly? See e.g. https://github.com/nextflow-io/nextflow/issues/1354 

Didn't it work in the past?

main.nf
d4straub: done

main.nf
d4straub: It actually works, done!

main.nf
d4straub: It actually works, done!

main.nf
d4straub: Done

main.nf
d4straub: done

skrakau:
Still, I think in general this way of allowing general custom options as strings is not ideal, we should fix those as soon as possible.

main.nf
skrakau: ok

d4straub:
>Regarding the time limit and the missing exit code, can it be a Nextflow or slurm problem if this is not reported properly? See e.g. nextflow-io/nextflow#1354
Didn't it work in the past?

Don't know, never happened to me in the past.

> Still, I think in general this way of allowing general custom options as strings is not ideal, we should fix those as soon as possible.

True, but I am not certain at this point which options might be most helpful.

skrakau (approved):

PR #152, opened 2021-02-04
Release 1.2.0

d4straub:
This is to prepare release 1.2.0, changes see CHANGELOG.md.

PR #154, opened 2021-02-04
Add some quotes and make group names nicer

d4straub:
Adressing comments in #152

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -1009,7 +1009,7 @@ if (params.coassemble_group) {
         set val(name), val(group), file(reads1), file(reads2) from ch_grouped_short_reads
 
         output:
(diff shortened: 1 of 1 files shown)


skrakau (changes requested):
Maybe it would make sense then also to change in `pool_short_reads` and `pool_long_reads`:
the use of of "group${group}" to ${name} (which is "group-${group}") ?

d4straub:
> Maybe it would make sense then also to change in pool_short_reads and pool_long_reads:
the use of of "group${group}" to ${name} (which is "group-${group}") ?

Absolutely, did I forget something?

skrakau (approved):
LGTM :)

PR #162, opened 2021-03-09
Dsl2 first draft

skrakau:
OK, here is a first running DSL2 version:

- [x] ...
- [x] Compared results to latest DSL1 version
- [x] Updated docs
- [x] Updated CHANGELOG
- [x] Added new `nf-core/test-datasets` in new format (https://github.com/nf-core/test-datasets/pull/226)
- [x] Replaced local `fastp` module by nf-core fastp module 
- [x] fix CAT db problem (built Biocontainer containing CAT and suitable Diamond version for latest CAT database)
- [x] solve S3 nf-amazon plugin problem (will be solved with next nextflow release)
- [x] make use of containers consistent (when only using tar etc.)
- [x] run and check full-size test
- [x] update required nextflow version (`nextflowVersion >= 20.10.0` for now, assuming release, although singularity requires `>=20.11.0-edge`)
- [ ] ~~fix non-reproducible centrifuge error  (on CFC and locally), any changed dependencies?~~

Use of local `Bowtie2` modules instead of using existing nf-core modules:
 1) `bowtie2_removal_build/align`:  different behaviour -> mapped reads are discarded for contamination removal
 2) `bowtie2_assembly_build/align`: since the index need to be build for all assemblies, I needed to hand over a `meta` map to the index building process as well to avoid complicated channel logic before and after calling the module

Use of local `quast` modules instead of using nf-core modules:
 1) `quast`, `quast_bins`: metaQUAST is used (not QUAST), with different processing for bins

I observed a non-reproducible centrifuge error (which disappears when using `-resume`) on different systems, which I did not observe with previous pipeline versions. However, since I could not reproduce it recently I will postpone this.

Since there are so many changes, I don't know what else to describe here. Let me know, if I should add something.

## PR checklist

 - [x] If necessary, also make a PR on the nf-core/mag _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.


modules/local/metabat2.nf
```diff
@@ -0,0 +1,51 @@
...
+    metabat2 -t "${task.cpus}" -i "${assembly}" -a depth.txt -o "MetaBAT2/${meta.assembler}-${meta.id}" -m ${params.min_contig_size} --unbinned --seed ${params.metabat_rng_seed}
+
+    # save unbinned contigs above thresholds into individual files, dump others in one file
+    split_fasta.py "MetaBAT2/${meta.assembler}-${meta.id}.unbinned.fa" ${params.min_length_unbinned_contigs} ${params.max_unbinned_contigs} ${params.min_contig_size}
+
+    mkdir MetaBAT2/discarded
+    mv "MetaBAT2/${meta.assembler}-${meta.id}.lowDepth.fa" MetaBAT2/discarded/
+    mv "MetaBAT2/${meta.assembler}-${meta.id}.tooShort.fa" MetaBAT2/discarded/
+    mv "MetaBAT2/${meta.assembler}-${meta.id}.unbinned.pooled.fa" MetaBAT2/discarded/
+    mv "MetaBAT2/${meta.assembler}-${meta.id}.unbinned.remaining.fa" MetaBAT2/discarded/
```
d4straub, line 44: These 4 files recently made ~50% of the size of the output folder, so it would be nice to have that zipped. Not urgent, but to keep in mind.

skrakau: https://github.com/nf-core/mag/issues/165

main.nf
```diff
@@ -14,201 +14,183 @@ Sabrina Krakau <sabrinakrakau@gmail.com>
...
 }
 
 // Ensure sample IDs are unique
-ch_sample_validate
-    .map { row -> row[0] }
+ch_input_rows
+    .map { id, group, sr1, sr2, lr -> id }
     .toList()
-    .map{ ids -> if( ids.size() != ids.unique().size() ) {exit 1, "ERROR: input contains duplicated sample IDs!" } }
+    .map { ids -> if( ids.size() != ids.unique().size() ) {exit 1, "ERROR: input samplesheet contains duplicated sample IDs!" } }
```
d4straub, line 189: This could go into a separate file.

main.nf
```diff
@@ -14,201 +14,183 @@ Sabrina Krakau <sabrinakrakau@gmail.com>
...
+include { NANOPLOT as NANOPLOT_FILTERED                       } from './modules/local/nanoplot'                    addParams( options: modules['nanoplot_filtered']          )
+include { CENTRIFUGE_DB_PREPARATION                           } from './modules/local/centrifuge_db_preparation'   addParams( options: [:]                                   )
+include { CENTRIFUGE                                          } from './modules/local/centrifuge'                  addParams( options: modules['centrifuge']                 )
+include { KRAKEN2_DB_PREPARATION                              } from './modules/local/kraken2_db_preparation'      addParams( options: [:]                                   )
+include { KRAKEN2                                             } from './modules/local/kraken2'                     addParams( options: modules['kraken2']                    )
+include { KRONA_DB                                            } from './modules/local/krona_db'                    addParams( options: [:]                                   )
```
d4straub, line 63: Here and others with empty options: why do you even add  `addParams( options: [:] )`? It seems to me it has no impact.

d4straub (commented):
Looks like great work!

Small points:
- I found twice optional outputs but with different syntax, not sure that works in both cases as intended: [modules/local/bowtie2_removal_align.nf](https://github.com/skrakau/mag/blob/00aae98353873c7434f018d2622ccff708dbc951/modules/local/bowtie2_removal_align.nf#L29) and [modules/local/busco.nf](https://github.com/skrakau/mag/blob/00aae98353873c7434f018d2622ccff708dbc951/modules/local/busco.nf#L28-L29)
- linting might break after the nf-core tools 1.13 rollout

skrakau:
The linting with tools 1.13.1 currently fails due to custom changes in `assets/multiqc_config.yaml` with:
```
╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ [✗] 1 Test Failed                                                                                                                                                               │
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ files_unchanged: assets/multiqc_config.yaml does not match the template                                                                                                         │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

`nf-core lint . --fix files_unchanged` would delete all my custom code. I don't know how to fix this right now.

main.nf
skrakau: I don't know, because others did it like this ;)

You as well https://github.com/d4straub/ampliseq/blob/9b70932af3045e0c3af9ce18456d23eabc1be593/workflows/ampliseq.nf#L159 
but seems one could omit it.

d4straub: Right, I could remove that I think. However here I did remove the addParams: https://github.com/d4straub/ampliseq/blob/9b70932af3045e0c3af9ce18456d23eabc1be593/workflows/ampliseq.nf#L149-L150

skrakau:
Thanks a lot for reviewing @d4straub ! The optional output seems to work in both cases so far, at least with the current Nextflow versions.

KevinMenden (approved):
Looks good to me!
I've been mainly just looking at the modules and DSL2 stuff. Most of the modules are local so you can basically write them however you want. Probably a good idea though to make them similar to nf-core modules, so it's easier to put them on nf-core/modules later.

Mostly I found that a process label is missing for I think basically all local modules, and sometimes emit statements are missing.

main.nf
skrakau: Hi @d4straub and @KevinMenden ,
were should this go, into a separate module file or should this be added to `functions.nf`?

modules/local/busco_db_preparation.nf
```diff
@@ -0,0 +1,33 @@
...
+    } else {
+        container "biocontainers/biocontainers:v1.2.0_cv1"
+    }
+
+    input:
+    path(database)
+
+    output:
+    path "buscodb/*", emit: db
+    path database
```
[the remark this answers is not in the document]
skrakau: in this case I think it doesn't make sense, since `database` is the input file. It is just used as output, because there is an option to publish this database file to `results`

main.nf
d4straub: I would put it into a separate file.

skrakau:
Hi @KevinMenden ,

I added the missing emit statements (apart from one exception where it doesn't make sense, see above).
 
Regarding the labels, we didn't use them for mag because it doesn't make sense to squeeze all those processes with very different requirements in these few categories. Thus I would also prefer adding the labels to the modules only when porting them to nf-core later.

main.nf
skrakau: should go into a proper `NF` script in `modules/local`

modules/local/bowtie2_assembly_build.nf
```diff
@@ -0,0 +1,31 @@
+// Import generic module functions
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+params.options = [:]
+options    = initOptions(params.options)
+
+process BOWTIE2_ASSEMBLY_BUILD {
+    tag "${meta.assembler}-${meta.id}"
+
```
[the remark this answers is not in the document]
skrakau: see answer below

modules/local/bowtie2_assembly_align.nf
```diff
@@ -0,0 +1,46 @@
+// Import generic module functions
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+params.options = [:]
+options    = initOptions(params.options)
+
+process BOWTIE2_ASSEMBLY_ALIGN {
+    tag "${assembly_meta.assembler}-${assembly_meta.id}-${reads_meta.id}"
+
```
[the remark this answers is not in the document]
skrakau: see answer below

skrakau:
I additionally changed some defaults in `nextflow.config` (and `nextflow_schema.json`) to clean up the params which are listed in the summary (-> only params != default).

KevinMenden:
Okay, but the modules are always the same processes, so requirements should be similar right?
Anyway it's only mandatory when putting them on nf-core/modules :) so no need to change  here

skrakau:
Thanks a lot for reviewing @KevinMenden ! 

Regarding labels: it is just that for mag there are for various processes error strategies applied for specific `exit codes`, different `maxRetries` applied, `exponential increases` of mem and time requirements applied when re-running, and in general the requirements where adjusted over time by people using the pipeline, and I don't want to mess with it right now (maybe later) :)

d4straub (approved):
Could't find any issues any more, looks good!

skrakau:
Parallelised GitHub CI test to address https://github.com/nf-core/mag/issues/170.

Issue #166, opened 2021-03-19
Process metabat requires too much time


d4straub:
## Check Documentation

I have checked the following places for your error:

- [x] [nf-core website: troubleshooting](https://nf-co.re/usage/troubleshooting)
- [x] [nf-core/mag pipeline documentation](https://nf-co.re/nf-core/mag/usage)

## Description of the bug

I perform currently a rather large assembly:
- 15 soil samples, Illumina only, `--coassemble_group --binning_map_mode "group"`
- SPAdes assembly not possible for me, estimated 3TB RAM required
- MEGAHIT assembly successful after 375h with 128 cpus and ~700GB RAM --> assembly: 14726955618 bytes ~15GB as fa.gz, 46.5 Gbases large, ~74 million contigs
- process metabat runs ages (>300h, 100 cores), toubleshooting:
  - metabat2 command requires ~12h, ~1200 MAGs produced
  - split_fasta.py requires too much time (>300h), uses only 1 cpu
  - split_fasta.py input is *.unbinned.fa that has a size of 5269544057 bytes ~ 5GB
  - split_fasta.py takes the following arguments: min_length_unbinned_contigs 100000 (instead of default 1000000), max_unbinned_contigs default, min_contig_size default

## Steps to reproduce

Well, apart from investing a lot of runtime, I'd not do that on purpose.

## Expected behaviour

This step should not take unreasonable long and would be best to be parallelized. With smaller assemblies that I performed with this pipeline the runtime was always reasonable.
Removing that step entirely seems not a good solution since #27 is still valid!

skrakau:
Solved in https://github.com/nf-core/mag/pull/175

PR #175, opened 2021-03-24
Fix split_fasta.py bug and improve runtime

skrakau:
There were two problems with the `split_fasta.py` script:

Bug:
 -  The sort function `sort_values()` was applied to `df`, but the result was nowhere stored. So I added an `inplace=True`.
 -  In any case, the index still refers to the position before sorting. Currently, after `max_sequences` it stops writing sequences to individual files, although the first `max_sequences` where not the longest and not necessarily have a `length >= length_threshold`. Thus sequences are missed. To address this I reset the index.

Runtime (see https://github.com/nf-core/mag/issues/166):
  - The `sort_values()` function is  applied to all sequences (O(n log n)). This is not necessary, since one can first separate all sequences below the `length_threshold`. Thus one can sort only the sequences `>= length_threshold`, take the longest `max_sequences` and add the remaining to the `pooled` list.
  - For the file with the unbinned sequences, which required > 300h this run through in a couple of minutes
 

## PR checklist

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -19,6 +19,34 @@ jobs:
         run: npm install -g markdownlint-cli
       - name: Run Markdownlint
         run: markdownlint ${GITHUB_WORKSPACE} -c ${GITHUB_WORKSPACE}/.github/markdownlint.yml
+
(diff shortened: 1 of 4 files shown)


d4straub (approved):
LGTM

PR #178, opened 2021-03-25
Add taxonomic classification for bins with gtdb-tk

skrakau:
I played with [GTDB-tk](https://ecogenomics.github.io/GTDBTk/index.html#) for taxonomic classification as an alternative for CAT:

- previous GTDB references stay accessible (CAT DBs only upon request currently)
- it will be clear which GTDB references will be compatible with which GTDB-tk version, i.e. it will be guaranteed that all pipelines version (with a specific GTDB-tk version) can be run at all times

Disadvantages:

- only uses marker genes, will not work for smaller bins

TODOs:

- [x] update BUSCO and make use of auto lineage functionality
- [ ] ~use `--force` to continue processing if an error occurs on a single genome (since `*.classify.tree` output would be interesting, parallelising over individual bins is likely not a good option)~
- [x] merge summary with Busco and quast summaries

(addresses https://github.com/nf-core/mag/issues/173)

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
 - [x] If you've added a new tool - add to the software_versions process and a regex to `scrape_software_versions.py`
 - [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/mag/tree/master/.github/CONTRIBUTING.md)
 - [ ] If necessary, also make a PR on the nf-core/mag _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core lint .`).
- [ ] Ensure the test suite passes (`nextflow run . -profile test,docker`).
- [x] Usage Documentation in `docs/usage.md` is updated.
- [x] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [x] `README.md` is updated (including new tool citations and authors/contributors).

PR #179, opened 2021-04-08
Update to busco 5.1.0 and enable automated lineage selection

skrakau:
Among others in preparation for adding GTDB-tk (https://github.com/nf-core/mag/pull/178).

This updates Busco from 4.1.4 to 5.1.0 and enables the use of the automated lineage selection.

- by default now the BUSCO parameter `--auto-lineage` is used and the data is downloaded automatically (also for tests)
- the lineage can still be specified by providing a lineage dataset via the mag parameter `--busco_reference` (uses BUSCO parameter `--lineage_dataset`). This still requires the download of a `file_versions.tsv` file, which is used to check if the newest lineage dataset is used (BUSCO warns if not).
- to run BUSCO in offline mode, the mag parameter `--busco_download_path` can be used (BUSCO: `--download_path` in currently only in combination with `--auto-lineage` or `--auto-lineage-prok`)
- additionally the mag parameter `--busco_auto_lineage_prok` can be used to ignore eukaryotes (BUSCO: `--auto_lineage_prok`)
- the mag parameter `--save_busco_reference` is also used to save the lineage datasets downloaded by BUSCO

Open todos:
- [ ] compress files when saving dowloaded lineage datasets
- [ ] update docs
- [ ] update to 5.1.2: maybe in another PR, since I had problems with the current singularity images

## PR checklist

### DIFF
--- a/lib/Completion.groovy
+++ b/lib/Completion.groovy
@@ -3,10 +3,13 @@
  */
 
 class Completion {
-    static void email(workflow, params, summary_params, projectDir, log, multiqc_report=[]) {
+    static void email(workflow, params, summary_params, projectDir, log, multiqc_report=[], busco_failed_bins = []) {
 
         // Set up the e-mail variables
         def subject = "[$workflow.manifest.name] Successful: $workflow.runName"
+        if (busco_failed_bins.size() > 0) {
+            subject = "[$workflow.manifest.name] Partially successful: For ${busco_failed_bins.size()} bin(s) the Busco analysis failed because no genes where found: $workflow.runName"
+        }
         if (!workflow.success) {
             subject = "[$workflow.manifest.name] FAILED: $workflow.runName"
         }
@@ -40,6 +43,7 @@ class Completion {
         email_fields['commandLine']  = workflow.commandLine
         email_fields['projectDir']   = workflow.projectDir
         email_fields['summary']      = summary << misc_fields
+        email_fields['busco_failed_bins'] = busco_failed_bins
         
         // On success try attach the multiqc report
         def mqc_report = null
@@ -111,8 +115,17 @@ class Completion {
         output_tf.withWriter { w -> w << email_txt }
     }
 
-    static void summary(workflow, params, log) {
+    static void summary(workflow, params, log, busco_failed_bins) {
         Map colors = Headers.log_colours(params.monochrome_logs)
+
+        if (busco_failed_bins.size() > 0) {
+            def failed_bins = ''
+            for (bin in busco_failed_bins) {
+                failed_bins += "    ${bin}\n"
+            }
+            log.info "-${colors.purple}[$workflow.manifest.name]${colors.red} For ${busco_failed_bins.size()} bin(s) the Busco analysis failed because no genes where found:\n${failed_bins}${colors.reset}-"
+        }
+
         if (workflow.success) {
             if (workflow.stats.ignoredCount == 0) {
                 log.info "-${colors.purple}[$workflow.manifest.name]${colors.green} Pipeline completed successfully${colors.reset}-"
--- a/modules/local/busco.nf
+++ b/modules/local/busco.nf
@@ -9,7 +9,7 @@ process BUSCO {
 
     publishDir "${params.outdir}",
         mode: params.publish_dir_mode,
-        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), publish_id:'') }
+        saveAs: { filename -> filename.indexOf("busco_downloads") == -1 ? saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), publish_id:'') : null }
 
     conda (params.enable_conda ? "bioconda::busco=5.1.0" : null)
     if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
@@ -25,6 +25,7 @@ process BUSCO {
 
     output:
     tuple val(meta), path("short_summary.specific.*.${bin}.txt"), optional:true , emit: summary
+    path('busco_downloads/'), optional:true                                     , emit: busco_downloads
     path("${bin}_busco.log")
     path("${bin}_buscos.faa.gz"), optional:true
     path("${bin}_buscos.fna.gz"), optional:true
--- a/modules/local/busco.nf
+++ b/modules/local/busco.nf
@@ -11,11 +11,11 @@ process BUSCO {
         mode: params.publish_dir_mode,
         saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), publish_id:'') }
 
-    conda (params.enable_conda ? "bioconda::busco=4.1.4" : null)
+    conda (params.enable_conda ? "bioconda::busco=5.1.0" : null)
     if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
-        container "https://depot.galaxyproject.org/singularity/busco:4.1.4--py_2"
+        container "https://depot.galaxyproject.org/singularity/busco:5.1.0--py_1"
     } else {
-        container "quay.io/biocontainers/busco:4.1.4--py_2"
+        container "quay.io/biocontainers/busco:5.1.0--py_1"
     }
 
     input:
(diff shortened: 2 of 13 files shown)


d4straub:
This is an awesome upgrade!

### Error handling:
> if run with --busco_reference the number of marker genes for the corresponding lineage is used and all output files just contain a "100% Missing" etc.

Sounds good

> if run with --auto-lineage number of marker genes is unknown and no busco results file is generated -> these contigs are missing in the MultiQC BUSCO report. In the final busco_summary.txt I put a NA since I still thought those contigs should be listed (could also be done differently) 

I guess you mean "bin" instead of "contig" here. Missing in the MultiQC report is fine. I absolutely agree that in `busco_summary.txt` bins with failed BUSCO analysis should appear anyhow. NA is ok, but if easily possible 100% Missing would be also great, e.g. by producing a dummy `"short_summary*" file` (and using NA only in the field for number of BUSCOs, i.e. "Total number"), but thats probably too much work for such a small improvement.

> since in this case the "short_summary*" file from busco is missing, I added additional output files for failed analyses which get processed in the downstream BUSCO_SUMMARY process:${bin}_busco.failed_bins.txt containing just the bin name

Not sure if this is needed since `busco_summary.txt` already contains that info, if I understood correctely? Nevertheless, one more file is fine as well. Just the bin name is fine.

> I added a Nextflow warning if busco analysis failed for bins due these error types (not for kept unbinned contigs though!)

I agree that this isn't really relevant for kept unbinned contigs. However, I'd hope that those still appear in the `busco_summary.txt`?

### Open todos:
> when using --busco_reference: add also a warning if for a binned genome 100% are missing?

Yes that would be great.

lib/Completion.groovy
```diff
@@ -111,8 +115,17 @@ class Completion {
...
-    static void summary(workflow, params, log) {
+    static void summary(workflow, params, log, busco_failed_bins) {
         Map colors = Headers.log_colours(params.monochrome_logs)
+
+        if (busco_failed_bins.size() > 0) {
+            def failed_bins = ''
+            for (bin in busco_failed_bins) {
+                failed_bins += "    ${bin}\n"
+            }
+            log.info "-${colors.purple}[$workflow.manifest.name]${colors.red} For ${busco_failed_bins.size()} bin(s) the BUSCO analysis failed because no genes where found or placements failed:\n${failed_bins}See ${params.outdir}/GenomeBinning/QC/BUSCO/[bin]_busco.err for further information.${colors.reset}-"
```
d4straub, line 126: Hm so this is also when `placements failed`, is that the same problem? I am uncertain.

modules/local/busco.nf
```diff
@@ -36,50 +40,70 @@ process BUSCO {
...
+        --out "BUSCO" > ${bin}_busco.log 2> ${bin}_busco.err; then
+
+        # get used db name
+        # (set nullgob: if pattern matches no files, expand to a null string rather than to itself)
+        shopt -s nullglob
+        summaries=(BUSCO/short_summary.specific.*.BUSCO.txt)
+        if [ \${#summaries[@]} -ne 1 ]; then
+            echo "ERROR: none or multiple 'BUSCO/short_summary.specific.*.BUSCO.txt' files found. Expected one."
+            exit 1
+        fi
```
d4straub, line 84: Curious: in what case can there be several specific summary files?

skrakau:
> I guess you mean "bin" instead of "contig" here. Missing in the MultiQC report is fine. I absolutely agree that in `busco_summary.txt` bins with failed BUSCO analysis should appear anyhow. NA is ok, but if easily possible 100% Missing would be also great, e.g. by producing a dummy `"short_summary*" file` (and using NA only in the field for number of BUSCOs, i.e. "Total number"), but thats probably too much work for such a small improvement.

Currently the `busco_summary.txt` looks like this when using auto selection:
```
GenomeBin       %Complete       %Complete and single-copy       %Complete and duplicated        %Fragmented     %Missing        Total number
SPAdes-test_minigut.1.fa        17.7    17.7    0.0     0.8     81.5    124
SPAdes-test_minigut.2.fa        13.7    13.7    0.0     3.2     83.1    124
MEGAHIT-test_minigut.1.fa       12.1    12.1    0.0     3.2     84.7    124
MEGAHIT-test_minigut.2.fa       18.5    18.5    0.0     0.8     80.7    124
SPAdes-test_minigut_sample2.unbinned.1.fa       0.0%    0.0%    0.0%    0.0%    100.0%  NA
SPAdes-test_minigut_sample2.unbinned.2.fa       0.0%    0.0%    0.0%    0.0%    100.0%  NA
...
```
I adjusted the summary.busco.py script for this.

> > since in this case the "short_summary*" file from busco is missing, I added additional output files for failed analyses which get processed in the downstream BUSCO_SUMMARY process:${bin}_busco.failed_bins.txt containing just the bin name
> 
> Not sure if this is needed since `busco_summary.txt` already contains that info, if I understood correctely? Nevertheless, one more file is fine as well. Just the bin name is fine.

`busco_summary.txt`  does not yet contain the info, because for those bins no summary file from BUSCO arrives. In theory the bin name would be sufficient yes, but I didn't get it done passing the bin names as `val` types, because `collect()` and passing it to the `BUSCO_SUMMARY` process only worked for files. Maybe there is a solution, but I didn't see a straight forward one. Do you know one?

> > I added a Nextflow warning if busco analysis failed for bins due these error types (not for kept unbinned contigs though!)
> 
> I agree that this isn't really relevant for kept unbinned contigs. However, I'd hope that those still appear in the `busco_summary.txt`?

sure :)

lib/Completion.groovy
skrakau: yes, in both cases bins would be listed here. 

The problem with the placement error is it says: 
`ERROR:  Placements failed. Try to rerun increasing the memory or select a lineage manually.`

In my case it was definitely not a memory problem. However, I am afraid that this could be caused by memory issues and then we would block retrying (such error messages do not really help...). That is why I pointed to the `${bin}_busco.err` file, so the user could at least discover this. It would of course not be nice, but I also don't know how else to handle this.

modules/local/busco.nf
skrakau: I don't know :D I think I just wanted to check if it is exactly 1

skrakau:
ok, finally another update :) roughly the following changes were done:

- added Nextflow warning that BUSCO did not find any matching genes also when using `--busco_reference`
- added used lineage dataset to `busco_summary.tsv`
- when run in auto lineage selection mode: results for both the selected domain and a more specific lineage are provided if available (`short_summary.domain.[lineage].[assembler]-[bin].txt`, `short_summary.specific_lineage.[lineage].[assembler]-[bin].txt`, `assembler]-[bin]_buscos.[lineage].fna.gz`, `[assembler]-[bin]_buscos.[lineage].faa.gz` and in `busco_summary.tsv`)
- `domain` results are now retrieved even if placement fails
- add 'Viruses' for domain in `busco_summary.tsv`
- fixed saving of datasets downloaded by BUSCO when using auto selection in online mode: since the complete reference data provided by BUSCO is > 100GB big (which could be used to ensure reproducibility), I wanted a setting which allows saving only the by BUSCO downloaded datasets. The downloaded data depends on the lineage selection. Therefore the process `BUSCO_SAVE_DOWNLOAD` is now run on a subset of the download folders (from different BUSCO runs), one for each selected specific lineage to make sure to retrieve all used datasets, while not overwriting already saved files. This now can be used to reproduce the BUSCO analysis (with same settings and same input data).
- regarding the saving of the downloaded files, I did not compress them yet as they did not take that much space yet, and like this the folder can directly be used as input for `--busco_download_path` 
- docs `output.md` and `usage.md` updated

and the last time I forgot to mention:
- the process `BUSCO_PLOT` does not output anymore the sample-wise summary files `[assembler]-[sample]-busco_summary.txt`, if this is needed I could easily add it again (generated with the `BUSCO_SUMMARY` process). Just wondered if they are used at all since they were also not mentioned in the `output.md` so far?

d4straub:
Really helpful changes!

> the process BUSCO_PLOT does not output anymore the sample-wise summary files [assembler]-[sample]-busco_summary.txt, if this is needed I could easily add it again (generated with the BUSCO_SUMMARY process). Just wondered if they are used at all since they were also not mentioned in the output.md so far?

Those are not used at all, because all this information is supposed to be in `busco_summary.tsv` anyway.

skrakau:
Thanks a lot @d4straub for all your feedback! Ready for re-review again

skrakau:
Correction:
unfortunately there are a few differences regarding the handling of `mollicutes` lineages and `failed placements` when running BUSCO in `online` and `offline` mode. Thus the saved datasets downloaded automatically by BUSCO can currently not be used to reproduce the analysis by specifying the already downloaded files via `--busco_download_path`.  I adjusted the docs accordingly.

d4straub (approved):
Looks good to me!

PR #202, opened 2021-05-26
Release 2.0.0

skrakau:
Main changes:
- DSL2 conversion
- BUSCO update and extension to use auto lineage selection
- Added GTDB-tk for taxonomic classification of bins as alternative to CAT
- Added process to generate CAT database
- Requires nextflow version `>= 21.04.0` 

see the CHANGELOG for details. The remaining open issues will be addressed in follow up releases, but this release is important as there is currently a problem with CAT and the available pre-built databases.

Since [semantic versioning](https://semver.org) says incompatible API changes are `MAJOR` releases, I increased the version to `2.0.0` due to the changed samplesheet format. Additionally, the default behaviour of `BUSCO` changed, i.e. use auto lineage selection now.

I checked the `DSL2 template` and the `viralrecon` pipeline as rough guides, please let me know if I missed useful or necessary changes.

## PR checklist


d4straub (approved):
Great!

conf/base.config
```diff
@@ -20,120 +20,106 @@ process {
...
+  withName: BOWTIE2_HOST_REMOVAL_BUILD {
+    cpus = { check_max (10 * task.attempt, 'cpus' ) }
+    memory = { check_max (20.GB * task.attempt, 'memory' ) }
+    time = { check_max (4.h * task.attempt, 'time' ) }
+  }
+  withName: BOWTIE2_HOST_REMOVAL_ALIGN {
+    cpus = { check_max (10 * task.attempt, 'cpus' ) }
+    memory = { check_max (10.GB * task.attempt, 'memory' ) }
+    time = { check_max (6.h * task.attempt, 'time' ) }
+  }
```
KevinMenden, line 32: I think this file is missing the standard process labels (process_high, process_low, ...) from the template, right?
Would be good to include (and use) them. Because all modules are build to use them.

lib/NfcoreSchema.groovy
```diff
@@ -0,0 +1,500 @@
...
+            }
+            // unexpected params
+            def params_ignore = params.schema_ignore_params.split(',') + 'schema_ignore_params'
+            def expectedParamsLowerCase = expectedParams.collect{ it.replace("-", "").toLowerCase() }
+            def specifiedParamLowerCase = specifiedParam.replace("-", "").toLowerCase()
+            if (!expectedParams.contains(specifiedParam) && !params_ignore.contains(specifiedParam) && !expectedParamsLowerCase.contains(specifiedParamLowerCase)) {
+                // Temporarily remove camelCase/camel-case params #1035
+                def unexpectedParamsLowerCase = unexpectedParams.collect{ it.replace("-", "").toLowerCase()}
+                if (!unexpectedParamsLowerCase.contains(specifiedParamLowerCase)){
+                    unexpectedParams.push(specifiedParam)
```
KevinMenden, line 121: as you're excluding this from the linting anyway, you could also exchange this file with the newest version from the template (nf-core/tools:dev). Fixed a little bug there.

modules/local/busco_db_preparation.nf
```diff
@@ -0,0 +1,33 @@
...
+
+    conda (params.enable_conda ? "conda-forge::sed=4.7" : null)
+    if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
+        container "https://containers.biocontainers.pro/s3/SingImgsRepo/biocontainers/v1.2.0_cv1/biocontainers_v1.2.0_cv1.img"
+    } else {
+        container "biocontainers/biocontainers:v1.2.0_cv1"
+    }
+
+    input:
+    path(database)
```
KevinMenden, line 22:
```suggestion
    path database
```

KevinMenden (approved):
Couldn't find any major things, looks really good! So it get's a ✅  from me.

The only thing I've found is that you don't seem to have the process labels in your `base.config`, but all modules use them.
On the same lines. your local could then use those labels. This way you could shrink your `base.config` file quite a bit, and the local modules are even more standardised.

conf/base.config
FriederikeHanssen: I agree. I am actually surprised that FASTP is not failing, since it uses `label 'process_medium' `. I had the weirdest error messages with missing labels in the base.config

FriederikeHanssen (approved):
Looks really nice. I am surprised the missing labels are not causing any problems. I had very weird and unrelated error messages when misspelling them for example. As Kevin mentioned, might be good to add them for the nf-core modules :)

skrakau:
there again occurred an issue with the aws full-size test, so I will check and fix this first ...

conf/base.config
skrakau: Added the labels to the `base.config` for the use with nf-core modules and to avoid errors in the future, but kept the `withName: XY` definitions for local modules.

skrakau:
Thanks a lot for reviewing @d4straub, @FriederikeHanssen and @KevinMenden !

PR #207, opened 2021-05-30
Fix GTDB-Tk summary process

skrakau:
So far `.size()` was used to check the number of input files within a process, which apparently does not work (or does not work anymore, not sure). When using `.sort.size()` (as in [modules/software/picard/mergesamfiles/main.nf](https://github.com/nf-core/modules/blob/68c678de935f9a2d26e55f70351d19358568f505/software/picard/mergesamfiles/main.nf#L31)) it works as expected, so I used this instead. @drpatelh since you added the `PICARD_MERGESAMFILES` process, do you think that's a reasonable solution here too or would there be a better way?

However, in [modules/software/picard/mergesamfiles/main.nf](https://github.com/nf-core/modules/blob/68c678de935f9a2d26e55f70351d19358568f505/software/sortmerna/main.nf#L35) the plain `.size()` is used. I am wondering, if this is working like this?

Additionally fixed handling of failed and filtered bins in `summary_gtdbtk.py` for files containing multiple or no bins. 

## PR checklist

### DIFF
--- a/bin/summary_gtdbtk.py
+++ b/bin/summary_gtdbtk.py
@@ -69,8 +69,9 @@ def main(args=None):
     filtered = []
     if args.filtered_bins:
         for file in args.filtered_bins:
-            with open(file) as infile:
-                bin_name = infile.readline().split("\t")[0]
+            df = pd.read_csv(file, sep='\t', names=["bin_name", "reason"])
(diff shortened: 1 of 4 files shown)


d4straub (approved):
Looks good to me!

PR #332, opened 2022-08-25
Post release bump

jfy133:
For dev

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,16 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## dev
+
+### `Added`
+
+### `Changed`
+
+### `Fixed`
+
+### `Dependencies`
+
 ## v2.2.1 - 2022/08/25
 
 ### `Added`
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## dev
```
d4straub, line 6:
```suggestion
## 2.3.0dev
```

nextflow.config
```diff
@@ -267,7 +267,7 @@ manifest {
     description     = 'Assembly, binning and annotation of metagenomes'
     mainScript      = 'main.nf'
     nextflowVersion = '!>=21.10.3'
-    version = '2.2.1'
+    version = '2.2.2dev'
```
d4straub, line 270:
```suggestion
    version = '2.3.0dev'
```

d4straub (approved):
Typically we hope to not need to release another bug fix, i.e. 2.2.2, but rather a feature release, i.e. 2.3.0, I guess. No idea how that is handled usually. Not so important anyway.

jfy133:
Ok, lets keep CHANGELOG as `dev` (as that's how it's done by edfault AFAIK), but bump to 2.3.0 fine with me.

I actually just remembered that nf-core is meant to use a special `patch` branch for actual branches...

Maybe we should use that in the future!

https://github.com/nf-core/nf-co.re/blob/1caada5226c0969154df26ffc56fafa7904e674d/markdown/developers/tutorials/nf_core_contributing_overview.md#branch-protection

PR #365, opened 2022-12-06
Add binning with CONCOCT

### LINKED ISSUE #339: Add CONCOCT as an alternative binner
### Description of feature

https://github.com/BinPro/CONCOCT

At the request of @alexhbnr

jfy133:
Closes #339 

Note CONCOCT seems to be quite permissive (doesn't itself produce any 'failed' bins, so produces _a lot_ of bins, even after SPLIT_FASTA). I'm not yet sure if this will cause an issue with our CI tests, but lets see.

## PR checklist


conf/modules.config
```diff
@@ -434,16 +435,33 @@ process {
         ext.prefix = { "${meta.assembler}-MaxBin2-${meta.id}" }
         // if no gene found, will crash so allow ignore so rest of pipeline
         // completes but without MaxBin2 results
-    }
+    }\
```
jfy133, line 438:
```suggestion
    }
```

subworkflows/local/binning.nf
```diff
@@ -100,8 +102,29 @@ workflow BINNING {
...
+
+                                [ meta_new, bins, bams, bais ]
+                            }
+                            .multiMap {
+                                meta, bins, bams, bais ->
+                                    bins: [ meta, bins ]
+                                    bams: [ meta, bams, bais ]
+                            }
+
+        FASTA_BINNING_CONCOCT ( ch_concoct_input ) // TODO UPDATE MODULE TO GZIP OUTPUT!
```
jfy133, line 120:
```suggestion
        FASTA_BINNING_CONCOCT ( ch_concoct_input )
```

docs/output.md
```diff
@@ -331,12 +331,34 @@ All the files in this folder contain small and/or unbinned contigs that are not
...
+
+Note that CONCOCT does not output what it considers 'unbinned' contigs, therefore no 'discarded' contigs are produced here. You may still need to do your own manual curation of the resulting bins.
+
 ### DAS Tool
 
 [DAS Tool](https://github.com/cmks/DAS_Tool) is an automated binning refinement method that integrates the results of a flexible number of binning algorithms to calculate an optimized, non-redundant set of bins from a single assembly. nf-core/mag uses this tool to attempt to further improve bins based on combining the MetaBAT2 and MaxBin2 binning output, assuming sufficient quality is met for those bins.
 
 DAS Tool will remove contigs from bins that do not pass additional filtering criteria, and will discard redundant lower-quality output from binners that represent the same estimated 'organism', until the single highest quality bin is represented.
 
+> ⚠️ If DAS Tool does not find any bins passing your selected threshold it will exit with an error files. Such an error is 'ignored' by nf-core/mag, therefore you will not find files in the `GenomeBinning/DASTool/` results directory for that particular sample.
```
d4straub, line 360: > If DAS Tool does not find any bins passing your selected threshold it will exit with an error files.

Wondering if that is a typo or will there be indeed an error file?

subworkflows/local/binning.nf
```diff
@@ -19,6 +19,8 @@ include { MAG_DEPTHS                            } from '../../modules/local/mag_
 include { MAG_DEPTHS_PLOT                       } from '../../modules/local/mag_depths_plot'          addParams( options: params.mag_depths_plot_options    )
 include { MAG_DEPTHS_SUMMARY                    } from '../../modules/local/mag_depths_summary'       addParams( options: params.mag_depths_summary_options )
```
d4straub, line 21:
```suggestion
```

subworkflows/local/binning_refinement.nf
```diff
@@ -41,30 +42,38 @@ workflow BINNING_REFINEMENT {
                                         .branch {
                                             metabat2: it[0]['binner'] == 'MetaBAT2'
                                             maxbin2:  it[0]['binner'] == 'MaxBin2'
+                                            concoct: it[0]['binner'] == 'CONCOCT'
```
d4straub, line 45:
```suggestion
                                            concoct:  it[0]['binner'] == 'CONCOCT'
```

subworkflows/local/binning_refinement.nf
```diff
@@ -41,30 +42,38 @@ workflow BINNING_REFINEMENT {
...
                                         meta, fastatocontig2bin ->
                                             def meta_new = meta.clone()
                                             meta_new.remove('binner')
                                             [ meta_new, fastatocontig2bin ]
                                     }
                                     .groupTuple(by: 0)
+                                    .dump(tag: "fastatocontigbin_for_dastool")
+                                    // TODO: contigs names might be truncated? https://github.com/cmks/DAS_Tool/issues/78#issuecomment-1086543966
+                                    // Error: Contigs of contig2bin files not found in assembly: 36
+                                    // /home/jfellows/Documents/git/nf-core/mag/testing/work/77/89fbe4f1d985f06ebfac3816b87560
```
d4straub, line 70: Is that still a problem?

workflows/mag.nf
```diff
@@ -561,7 +561,7 @@ workflow MAG {
         * DAS Tool: binning refinement
         */
 
-        if ( params.refine_bins_dastool && !params.skip_metabat2 && !params.skip_maxbin2 ) {
+        if ( params.refine_bins_dastool && !params.skip_metabat2 && !params.skip_maxbin2 && !params.skip_concoct ) {
```
d4straub, line 564: Do I understand this line correct that refinement with dastool will only be used when all three binners were not skipped? While I agree that it is a good idea to use all 3 binners for refinement, shouldnt it be possible to also use refinement with only 2 binners?

docs/output.md
jfy133, line 360:
```suggestion
> ⚠️ If DAS Tool does not find any bins passing your selected threshold it will exit with an error. Such an error is 'ignored' by nf-core/mag, therefore you will not find files in the `GenomeBinning/DASTool/` results directory for that particular sample.
```

Oops left over from a re-phrasing

subworkflows/local/binning_refinement.nf
jfy133: It wasn't an issue with our pipeline but a bug in DAS_Tool and in this case only effects CONCOCT:

https://github.com/cmks/DAS_Tool/issues/90

I'm waiting to see if there will be a quick patch release for it in the next day or so (I wanted the rest of this PR reviewed in the meantime), if not I'll go back and fix the header names coming out of CONCOCT so they don't have spaces in them.

I just noticed I forgot to put this to draft... will do that now

jfy133: While there is some activity on DAS_Tool I don't know how fast it will be. Instead I will 'fix' the MEGAHIT contig names instead

jfy133: I've added a 'fix' for the MEGAHIT contig names (which ended up being more complicated due to PROKKA contig length name restrictions...)

workflows/mag.nf
jfy133: Done

workflows/mag.nf
```diff
@@ -561,7 +561,8 @@ workflow MAG {
         * DAS Tool: binning refinement
         */
 
-        if ( params.refine_bins_dastool && !params.skip_metabat2 && !params.skip_maxbin2 ) {
+        // If any two of the binners are both skipped at once, do not run because DAS_Tool needs at least one
+        if ( params.refine_bins_dastool && !( ( params.skip_metabat2 && params.skip_maxbin2 ) || ( params.skip_metabat2 && params.skip_concoct ) || ( params.skip_maxbin2 && params.skip_concoct ) ) ) {
```
d4straub, line 565:
```suggestion
        if ( params.refine_bins_dastool && [ params.skip_metabat2, params.skip_maxbin2, params.skip_concoct ].count(false) > 1 ) {
```
I believe this would be better, not an exponential increase of tests, but linear. I tested it with nextflow console, but with variables instead of params, but I expect it would work?

d4straub (approved):
Looks good to me.

workflows/mag.nf
jfy133: Ooooh that's clever, I admittedly get stuck on how to do this! I will try to test locally before committing

jfy133:
@d4straub Thanks for the review! I checked and your method is indeed better! I noticed this check (and another I added earlier) were in the wrong place, I've moved that to the correct place now. Could you just check for this one more time (and if the error message is good), then I'll merge once tests pass

lib/WorkflowMag.groovy
```diff
@@ -85,14 +85,15 @@ class WorkflowMag {
...
+
         // Check if at least two binners were applied in order to run DAS Tool for bin refinment
         // (needs to be adjusted in case additional binners are added)
-        if (params.refine_bins_dastool && params.skip_metabat2 ) {
-            log.error 'Both --refine_bins_dastool and --skip_metabat2 are specified! Invalid combination, bin refinement requires MetaBAT2 and MaxBin2 binning results.'
-            System.exit(1)
-        }
-        if (params.refine_bins_dastool && params.skip_maxbin2 ) {
-            log.error 'Both --refine_bins_dastool and --skip_maxbin2 are specified! Invalid combination, bin refinement requires MetaBAT2 and MaxBin2 binning results.'
+        if ( params.refine_bins_dastool && [ params.skip_metabat2, params.skip_maxbin2, params.skip_concoct ].count(false) <= 1 ) {
```
d4straub, line 95: Doesnt `.count(false) <= 1` mean that 1 or less of the binners are run?
I had in mind to check whether 2 or more binners (> 1) are not skipped (false), following line 93 `Check if at least two binners were applied in order to run DAS Tool for bin refinment`:
```suggestion
        if ( params.refine_bins_dastool && [ params.skip_metabat2, params.skip_maxbin2, params.skip_concoct ].count(false) > 1 ) {
```

jfy133: Ah, yeah sorry. I flipped the use case here, as I want to throw an error if it is less than 1.

But using > 1 means you don't have to adjust in the future (I just need to take the opposite to throw the error), like you suggest :+1: 

Added a comment for myself as I'm very bad at thinking about these things abstractly.

jfy133: I've tested now @d4straub! If you're OK with it can you merge.

d4straub: Haha true, I had it wrong here! I am happy with the change!

jfy133: Ok, once tests pass I merge (ignoring the template lint stuff)

PR #366, opened 2022-12-06
Add CAT_SUMMARY process and offical_taxonomy param

prototaxites:
Added a CAT_SUMMARY process to summarise the output of CAT as a single final file, as currently there is a single output per assembly group (a lot if you have many single-sample assemblies!). I wanted to re-use the COMBINE_TSV process (as I essentially re-wrote it before realising it exists), but the output files from CAT are gzipped so I've cloned it and added an ungzipping line, rather than modify either the CAT output or push all the files through the GUNZIP process. I don't like duplicating code but this seemed like the simplest solution.

Also added an option to make CAT only output 'official' taxonomy, i.e. Kingdom, Phylum, etc. - I noticed this was available in the CAT documentation and thought it would be useful, as currently munging the taxonomy with many mismatched and empty fields is quite annoying! 

Haven't run any tests yet as I've got a pipeline running on our cluster where I have the database downloaded (I don't think CAT will run on Gitpod?), but will un-draft this once I've had a chance to run a test/if anyone else wants to run a quick test.

## PR checklist


prototaxites:
Finally got a chance to test this on our cluster - all working fine now.

conf/test.config
```diff
@@ -29,4 +29,5 @@ params {
     busco_reference               = "https://busco-data.ezlab.org/v5/data/lineages/bacteria_odb10.2020-03-06.tar.gz"
     busco_clean                   = true
     gtdb                          = false
+    cat_official_taxonomy         = true
```
d4straub, line 32: As far as I know CAT isnt run in the test profile anyway, because the db is just too large for CI tests.
So I propose removing this line.
But CAT is run in https://github.com/nf-core/mag/blob/master/conf/test_full.config, if you want you could activate cat_official_taxonomy there?

d4straub (approved):
Looks good to me.
I also restarted test runs because they seem to have failed due to connectivity problems. Linting is expected to fail because of the template update.
In case all test pass (except linting) but you cannot merge the PR, ping me and I'll do it.

conf/test.config
prototaxites: Ah, that's a good shout - I'll make that change!

prototaxites:
@d4straub All passing except linting as well - could you merge? Many thanks!

PR #373, opened 2022-12-20
Merging template updates 2.7.2

skrakau:
Template update 2.7.2

## PR checklist

### DIFF
--- a/modules/nf-core/multiqc/main.nf
+++ b/modules/nf-core/multiqc/main.nf
@@ -1,7 +1,7 @@
 process MULTIQC {
     label 'process_single'
 
-    conda (params.enable_conda ? 'bioconda::multiqc=1.13' : null)
+    conda "bioconda::multiqc=1.13"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
         'https://depot.galaxyproject.org/singularity/multiqc:1.13--pyhdfd78af_0' :
         'quay.io/biocontainers/multiqc:1.13--pyhdfd78af_0' }"
--- a/workflows/mag.nf
+++ b/workflows/mag.nf
@@ -222,6 +222,7 @@ workflow MAG {
         if ( params.clip_tool == 'fastp' ) {
             ch_clipmerge_out = FASTP (
                 ch_raw_short_reads,
+                [],
                 params.fastp_save_trimmed_fail,
                 []
             )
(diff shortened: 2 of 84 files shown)


modules/nf-core/multiqc/main.nf
```diff
@@ -1,7 +1,7 @@
 process MULTIQC {
     label 'process_single'
 
-    conda (params.enable_conda ? 'bioconda::multiqc=1.13' : null)
+    conda "bioconda::multiqc=1.13"
```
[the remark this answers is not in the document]
jfy133: Yes it should q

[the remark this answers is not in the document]
skrakau: Ups, will do

workflows/mag.nf
```diff
@@ -241,17 +241,8 @@ workflow MAG {
             ADAPTERREMOVAL_PE ( ch_adapterremoval_in.paired, [] )
             ADAPTERREMOVAL_SE ( ch_adapterremoval_in.single, [] )
 
-            // pair1 and 2 come for PE data from separate output channels, so bring
-            // this back together again here
-            ch_adapterremoval_pe_out = Channel.empty()
```
jfy133, line 246: This deletion may not work, as if you do merging of PE data you won't have the `truncated` output channel from the AR module (reviewed from my phone, so I'm not 💯 sure)

jfy133 (commented):
Overall looks straightforward to me, I'm not sure about the AR2 output channel refactor though - something to check

workflows/mag.nf
skrakau: Thanks, will check

skrakau:
I hope I didn't miss anything :D 

Maybe worth mentioning:
- updated nf-core modules with tool version updates: 
   - BCFtools  1.14 -> 1.16
   - Freebayes 1.3.5 -> 1.3.6
   - SAMtools 1.15 -> 1.16.1 
   
And I saw that `adapterremoval` says it's not deterministic when run with multiple threads, even when using the `--seed` parameter. Maybe it would be worth adding an option to run this in a deterministic manner.

d4straub (approved):
Looks good to me. Not sure about the channel problem above though.

workflows/mag.nf
jfy133: Actually ignore me, I confused myself - you don't want to do merging in this pipeline :sweat_smile: so it's irrelevent! You're correct , the output channel is just empty because the module wouldn't get executed.

jfy133 (approved):

jfy133:
> And I saw that `adapterremoval` says it's not deterministic when run with multiple threads, even when using the `--seed` parameter. Maybe it would be worth adding an option to run this in a deterministic manner.

This is only when you're doing read-merging (collapsing), which is not performed in the pipeline  so we should be fine :)

PR #374, opened 2022-12-21
Deactivate CONCOCT in most tests

jfy133:
Except bin refinement, where it is anyway most relevent.

This will allow most tests not take ages to complete. We could also further separate this so we do an extremely minimal run and only run CONCOCT (no other downstream steps), if these tests remain too length.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### `Added`
 
 - [#353](https://github.com/nf-core/mag/pull/353) - Added the busco_clean parameter to optionally clean each BUSCO directory after a successful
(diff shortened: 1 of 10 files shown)


d4straub (commented):
Looks good to me, but why is the `test_ancient_dna` test failing? Dont get it.

jfy133:
No, there is some weird corruption thing going on I think. @skrakau has also observed this.... Not sure why. I'll be looking into it next week when I'm back from holiday

jfy133:
Ok so the issue is for the aDNA run we bizarrely get a join mismatch:

https://github.com/nf-core/mag/actions/runs/3831704693/jobs/6609861357#step:4:271

But this is separate from the aDNA 'subworkflow'... 

(Cc @skrakau )

jfy133:
@d4straub @skrakau this is now working :) linting failure is because template has not been merged yet, but the rest should be OK if you're happy  with it

d4straub:
Hm, tests failed, I restarted them.

d4straub (approved):
Looks good to me

PR #385, opened 2023-01-30
Run CAT on unbinnned contigs as well as binned contigs

### LINKED ISSUE #125: Apply CAT also on unbinned contigs
Currently CAT is only applied to MetaBAT bins for taxonomic classification. Maybe it would be useful to apply it on whole assemblies / also on unbinned contigs.

jfy133:
Closes #125 

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [#353](https://github.com/nf-core/mag/pull/353) - Added the busco_clean parameter to optionally clean each BUSCO directory after a successful (by @prototaxites)
 - [#361](https://github.com/nf-core/mag/pull/361) - Added the skip_clipping parameter to skip read preprocessing with fastp or adapterremoval. Running the pipeline with skip_clipping, keep_phix and without specifying a host genome or fasta file skips the FASTQC_TRIMMED process (by @prototaxites)
-- [#365](https://github.com/nf-core/mag/pull/365) - Adds CONCOCT as an additional (optional) binning tool (by @jfy133)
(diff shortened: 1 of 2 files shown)


d4straub:
Did you test that? Maybe I misunderstood, but doesnt your change mean that also _all_ unbinned contigs (potentially _millions_) are classified by CAT? If I am right, CAT will be unusable for larger (and not very well assembled & binned) assemblies because it will take around >=1000x longer. This would need to be optional or only contigs above a specific length threshold (that would be adjustable) or such.

jfy133:
> 
Did you test that? Maybe I misunderstood, but doesnt your change mean that also all unbinned contigs (potentially millions) are classified by CAT? If I am right, CAT will be unusable for larger (and not very well assembled & binned) assemblies because it will take around >=1000x longer. This would need to be optional or only contigs above a specific length threshold (that would be adjustable) or such.

On the test data yes, not on real data (I dont have any at hand). Note that the 'unbinned' means the contigs considered 'good' from [split_fasta](https://github.com/nf-core/mag/blob/4b215b7939f25919d7a9dda23cdd4f793b5859a9/subworkflows/local/binning.nf#L197-L198), so there is a certain level of filtering going on there. It's not just any random contig.

I could make it optional, but maybe it would be to discuss on #125 then before I continue, if this is not sufficient?

d4straub:
I see, thanks, I didnt make the connection! Indeed, then "only" contigs that are above `--min_length_unbinned_contigs` are added, i.e. by default >=1.000.000 bp, that seems good!

d4straub (approved):
LGTM!

jfy133:
> I see, thanks, I didnt make the connection! Indeed, then "only" contigs that are above --min_length_unbinned_contigs are added, i.e. by default >=1.000.000 bp, that seems good!

Correct! Basically the same input for e.g. BUSCO and other dowsntream tools :)

PR #395, opened 2023-02-23
Add domain-level classification for bins

prototaxites:
edit: ready to go now!

This one is still a work-in-progress, but I'm putting this up now as it passes `-profile test`, with all bins from the minigut being classified as prokarya, as expected. If anyone has a dataset which produces eukaryotic MAGs, I'd be keen to hear if it works for you.

Aside from adding the domain classification subworkflow, I've split off the coverage/MAG_DEPTHS processes to a new subworkflow, which allows for a set of final bins to be determined (following refinement, classification, etc.) before calculating coverage. 

## PR checklist

  - [x] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/mag/tree/master/.github/CONTRIBUTING.md)
  - [x] If necessary, also make a PR on the nf-core/mag _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.


prototaxites:
Created a PR for test data with eukaryotic reads to the mag branch of the test-datasets repo: https://github.com/nf-core/test-datasets/pull/782

conf/test.config
```diff
@@ -26,7 +26,7 @@ params {
     skip_krona                    = true
     min_length_unbinned_contigs   = 1
     max_unbinned_contigs          = 2
-    busco_reference               = "https://busco-data.ezlab.org/v5/data/lineages/bacteria_odb10.2020-03-06.tar.gz"
+    //busco_reference               = "https://busco-data.ezlab.org/v5/data/lineages/bacteria_odb10.2020-03-06.tar.gz"
```
d4straub, line 29: Any reason why this is commented? 
The param below (busco_clean) is related...

conf/test_domain_classification.config
```diff
@@ -11,24 +11,22 @@
 */
 
 params {
-    config_profile_name        = 'Test profile for skipping all short read preprocessing'
-    config_profile_description = 'Minimal test dataset to check pipeline function when all short read preprocessing is skipped.'
+    config_profile_name        = 'Test profile for domain classificaiton of bins.'
```
d4straub, line 14:
```suggestion
    config_profile_name        = 'Test profile for domain classification of bins.'
```
But isnt that rather `'Test profile for domain classification of bins and skipping all short read preprocessing.'`?

conf/test.config
prototaxites: Whoops! Had that commented out for testing as it conflicts with --skip_binqc, but didn't mean to actually commit the change...

conf/test_domain_classification.config
prototaxites: Yes, I think that's probably better - thought the latter was less relevant but best to be explicit.

prototaxites:
The binrefinement test seems to be failing because it's running out of space:

```
  docker: failed to register layer: ApplyLayer exit status 1 stdout:  stderr: write /usr/local/lib/R/library/methods/help/methods.rdb: no space left on device.
```

I don't think the outputs of the test should have changed based on the changes in the PR - anyone have any thoughts on what's up?

d4straub:
As far as I know there is only a specific amount of storage space available on github tests, maybe that was now exceeded. I just restarted the failed test, maybe it was just a temporal hiccup.

prototaxites:
No luck! Looking at the binrefinement config, I notice that the busco_clean parameter isn't set. A BUSCO run can take up quite a lot of space, and with CONCOCT adding ~129 bins (many of them single contig), this might be mean the test is using a lot more storage than is necessary, resulting in failure.

It might be worth enabling it within the test config?

prototaxites:
It works! 😅 Reviews welcome 😊

subworkflows/local/depths.nf
```diff
@@ -0,0 +1,67 @@
+params.mag_depths_options                           = [:]
+params.mag_depths_plot_options                      = [:]
+params.mag_depths_summary_options                   = [:]
+
+include { MAG_DEPTHS                            } from '../../modules/local/mag_depths'               addParams( options: params.mag_depths_options         )
+include { MAG_DEPTHS_PLOT                       } from '../../modules/local/mag_depths_plot'          addParams( options: params.mag_depths_plot_options    )
+include { MAG_DEPTHS_SUMMARY                    } from '../../modules/local/mag_depths_summary'       addParams( options: params.mag_depths_summary_options )
```
d4straub, line 7: `addParams` shouldnt be there any more afaik.
```suggestion
include { MAG_DEPTHS                            } from '../../modules/local/mag_depths' 
include { MAG_DEPTHS_PLOT                       } from '../../modules/local/mag_depths_plot'
include { MAG_DEPTHS_SUMMARY                    } from '../../modules/local/mag_depths_summary'
```

workflows/mag.nf
```diff
@@ -581,45 +603,70 @@ workflow MAG {
...
+                // bins which cannot be classified ('unknown') get fed into binning refinement
+                // but also could putatively be eukaryotic so keep separate as well
+                ch_prokarya_bins_dastool = ch_binning_results_bins
+                    .filter { meta, bins ->
+                        meta.domain in ["bacteria", "archaea", "prokarya", "organelle", "unknown"]
+                    }
+
+                ch_eukarya_bins_dastool = ch_binning_results_bins
+                    .filter { meta, bins ->
+                        meta.domain in ["eukarya", "unknown"]
```
d4straub, line 618: Wont there now be two copies of bins with `meta.domain == "unknown"` in `ch_refined_bins` because both, `ch_prokarya_bins_dastool` and `ch_eukarya_bins_dastool`, contain "unknown" and will be mixed in `ch_refined_bins = ch_eukarya_bins_dastool.mix(BINNING_REFINEMENT.out.refined_bins)`?

workflows/mag.nf
```diff
@@ -718,14 +791,23 @@ workflow MAG {
         /*
          * Prokka: Genome annotation
          */
-        ch_bins_for_prokka = ch_input_for_postbinning_bins_unbins.transpose()
+
```
d4straub, line 794:
```suggestion
```

workflows/mag.nf
```diff
@@ -761,7 +843,6 @@ workflow MAG {
     ch_multiqc_files = ch_multiqc_files.mix(ch_quast_multiqc.collect().ifEmpty([]))
     ch_multiqc_files = ch_multiqc_files.mix(ch_bowtie2_assembly_multiqc.collect().ifEmpty([]))
     ch_multiqc_files = ch_multiqc_files.mix(ch_busco_multiqc.collect().ifEmpty([]))
-
```
d4straub, line 764: ? why?

conf/modules.config
```diff
@@ -261,23 +261,14 @@ process {
         ]
     }
 
-    withName: 'MAG_DEPTHS_PLOT|MAG_DEPTHS_SUMMARY|MAG_DEPTHS_PLOT_REFINED' {
+    withName: 'MAG_DEPTHS_PLOT|MAG_DEPTHS_SUMMARY|' {
```
d4straub, line 264:
```suggestion
    withName: 'MAG_DEPTHS_PLOT|MAG_DEPTHS_SUMMARY' {
```

d4straub (commented):
Nice work! 
The PR does make quite some changes, also in files I am not particular familiar with (the remodeled binning), so I'd appreciate another opinion, but it seems fine to me.

workflows/mag.nf
prototaxites: I think that depends on the outcome of DASTool - here, we assume that an 'unknown' bin could be putatively eukaryotic or prokaryotic. On the assumption that it's eukaryotic, we keep it as is, and on the assumption it's prokaryotic, it gets fed into DASTool. Whether we get the same bin out at the end of the DASTool process depends on what's in the prokaryotic bins - I presume a bin that's 'unknown' is probably pretty poor quality anyway, so it might get chopped up and moved around?

workflows/mag.nf
prototaxites: accidental deletion I think!

prototaxites:
Does anyone else have any comments? I'd really like to work on adding a eukaryotic annotation step to the pipeline, but ideally this would build on top of this PR. Otherwise, I could submit a separate PR which adds this to run on all bins regardless of domain, if that is preferable?

bin/domain_classification.R
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env Rscript
```
jfy133, line 1: Please add license information here - if you've created this script specifically for the pipeline (looks like it could be), MIT is fine. Else please  use the license where the code has been copied from.

Nice example/tempalte when taken from elsewhere: https://github.com/nf-core/viralintegration/blob/master/bin/chimeric_contig_evidence_analyzer.py

Example when written for the pipeline: https://github.com/nf-core/eager/blob/dev/bin/filter_bam_fragment_length.py#L3-L4

conf/modules.config
```diff
@@ -283,23 +283,14 @@ process {
...
 
-    withName: 'MAG_DEPTHS_SUMMARY_REFINED' {
-        ext.prefix = "bin_refined_depths_summary"
-        publishDir = [
-            path: { "${params.outdir}/GenomeBinning/depths/bins" },
-            mode: params.publish_dir_mode,
-            pattern: "*.{tsv}"
-        ]
-    }
-
```
jfy133, line 302: Reason why both have deleted (make sense to collapse into one `withName` though, if that was the original purpose)?

conf/modules.config
```diff
@@ -666,6 +661,42 @@ process {
...
+
+    withName: TIARA_CLASSIFY {
+        publishDir = [
+            path: { "${params.outdir}/GenomeBinning/Tiara" },
+            mode: params.publish_dir_mode,
+            pattern: "*.binclassification.tsv"
+        ]
+        ext.args = { "--join_prokaryotes --assembler ${meta.assembler}" }
+        ext.prefix = { "${meta.assembler}-${meta.binner}-${meta.bin}-${meta.id}" }
+    }
```
jfy133, line 689: Maybe make this optional (and by default turned off) given you have `TIARA_SUMMARY` vthat presumably has all the info for you just in one place.

docs/output.md
```diff
@@ -392,6 +392,21 @@ By default, only the raw bins (and unbinned contigs) from the actual binning met
 
 ⚠️ Due to ability to perform downstream QC of both raw and refined bins in parallel (via `--postbinning_input)`, bin names in DAS Tools's `*_allBins.eval` file will include `Refined`. However for this particular file, they _actually_ refer to the 'raw' input bins. The pipeline renames the input files prior to running DASTool to ensure they can be disambiguated from the original bin files in the downstream QC steps.
 
+### Tiara
+
+Tiara is a contig classifier that identifies the domain (prokarya, eukarya) of contigs within a file. This is used in this pipeline to find the most likely domain classification of each bin or unbin based on its contig identities.
```
jfy133, line 397: I wonder if it to make it clear it's in the same scope as DAS_Tool by explicitly writing that it 'refines' bins along classification terms?

I also wonder if we should now have different subdirectories now to also guide users that not everything in  GenomeBinning is the same thing? E.g.

```
GenomeBinning/
    Binning/
        MaxBin
        MetaBat
        Concoct
    BinRefinement/
        DASTool
        Tiara
```

Or something - thoughts (also at @d4straub @skrakau)?

docs/output.md
```diff
@@ -392,6 +392,21 @@ By default, only the raw bins (and unbinned contigs) from the actual binning met
...
+<summary>Output files</summary>
+
+- `GenomeBinning/Tiara/`
+  - `[assembler]-[sample/group].tiara.txt` - Tiara output classifications (with probabilities) for all contigs within the specified sample/group assembly
+  - `[assembler]-[binner]-[bin_type]-[sample/group].binclassification.tsv` - Final classification of each bin/unbin based on the classification probabilities of each contig within the bin.
+  - `log/log_[assembler]-[sample/group].txt` - log file detailing the parameters used by the Tiara model for contig classification.
+- `GenomeBinning/tiara_summary.tsv` - Summary of Tiara domain classification for all bins.
+
+</details>
+
```
jfy133, line 409: I always like to have a sentence or two after the list of files of what you could use the subsequent files for

e.g. 'Typically you would use `tiara_summary.tsv` as the primary file to allow you to pull out XYZ, whereas only (`binclassification.tsv`) if you need XYZ additional information'

workflows/mag.nf
```diff
@@ -611,6 +613,26 @@ workflow MAG {
             )
         }
 
+        if ( params.bin_domain_classification ) {
+            DOMAIN_CLASSIFICATION ( ch_assemblies, BINNING.out.bins, BINNING.out.unbinned )
+            ch_binning_results_bins = DOMAIN_CLASSIFICATION.out.classified_bins
+            ch_binning_results_unbins = DOMAIN_CLASSIFICATION.out.classified_unbins
+            ch_versions = ch_versions.mix(DOMAIN_CLASSIFICATION.out.versions)
```
jfy133, line 620:
```suggestion
            ch_binning_results_bins   = DOMAIN_CLASSIFICATION.out.classified_bins
            ch_binning_results_unbins = DOMAIN_CLASSIFICATION.out.classified_unbins
            ch_versions               = ch_versions.mix(DOMAIN_CLASSIFICATION.out.versions)
```

subworkflows/local/domain_classification.nf
```diff
@@ -0,0 +1,143 @@
...
+            meta_new.bin = 'unbins'
+            [meta_new, unbins]
+        }
+
+    ch_tiara_input = bins.mix(unbins)
+
+    TIARA_TIARA ( assemblies )
+    ch_versions = ch_versions.mix(TIARA_TIARA.out.versions.first())
+    // Need contig2bin file for each bin group
+
```
jfy133, line 38:
```suggestion
    
    // Need contig2bin file for each bin group
```

subworkflows/local/domain_classification.nf
```diff
@@ -0,0 +1,143 @@
+/*
+* Domain classification with Tiara
+*/
+
+include { TIARA_TIARA                                                  } from '../../modules/nf-core/tiara/tiara/main'
+include { TIARA_CLASSIFY                                               } from '../../modules/local/tiara_classify'
+include { DASTOOL_FASTATOCONTIG2BIN as DASTOOL_FASTATOCONTIG2BIN_TIARA } from '../../modules/nf-core/dastool/fastatocontig2bin/main'
+include { COMBINE_TSV as TIARA_SUMMARY                                 } from '../../modules/local/combine_tsv'
+
+workflow DOMAIN_CLASSIFICATION {
```
jfy133, line 10: I wonder if the future users may request alternative domain classification tools, maybe it would be worth future-proofing and pushing most of the TIARA sections into a Tiara-speciufic sub workflow, and pull that into here with a if/else statement to allow user choice of tool (Currently only TIara)?

workflows/mag.nf
```diff
@@ -676,8 +723,18 @@ workflow MAG {
             /*
             * CheckM subworkflow: Quantitative measures for the assessment of genome assembly
             */
+
+            if (params.bin_domain_classification){
+                ch_input_bins_for_checkm = ch_input_bins_for_qc
+                    .filter { meta, bins ->
+                        meta.domain in ["bacteria", "archaea", "prokarya", "organelle", "unknown"]
```
jfy133, line 730: Coudl simplify/be more permissive to exclude eukarya instead?

nextflow.config
```diff
@@ -100,6 +100,8 @@ params {
     skip_metabat2                   = false
     skip_maxbin2                    = false
     skip_concoct                    = false
+    bin_domain_classification       = false
+    tiara_min_length                = 3000
```
jfy133, line 104: maybe make parameter more explicit: `domain_classification_tiara_min_length` or something, as I guess for some people come into the pipeline might not know exactly what `tiara` is (even if thye are interesting is in domain classification)

jfy133 (changes requested):
Overall really nice, and some good optimisations - thanks @prototaxites !

Mostly minor changes and questions:

- As you're adding enw functioanlity, you need to update the workflow diagram SVG!
- Do you need to add `${meta.domains}` to more of the `prefixes` in the post-domain classification steps in `modules.conf`, in cases once it's turned on (basically for any downstream step that consumes `ch_input_for_postbinning_bins_*` on input? (may not be necssary, but I want to double check).

conf/modules.config
prototaxites: I moved all the MAG_DEPTHS processes to a new subworkflow that runs after bin refinement, and operates depending on which bins are kept as defined in `params.postbinning_input`, so there's no separate calculation of refined bin depths any more (maybe there should be?). Otherwise, yes, just collapsing into one withName!

docs/output.md
d4straub: > Alternatively, Tiara might go in the Taxonomy directory as it's technically broadly classifying bins taxonomically?

This sounds good to me! Then it would be together in a folder with centrifuge, kraken2, CAT and GTDBTK taxonomy assignments (the latter 2 on bins, the former 2 on reads).

nextflow.config
d4straub: I disagree because we have also stuff like `freebayes_ploidy` or `pydamage_accuracy` in there that also do not specify what the tool is doing.

conf/modules.config
prototaxites: Good point - might suggest this is actually entirely unnecessary to publish these files as tiara_summary.tsv should just be all of these concatenated together.

subworkflows/local/domain_classification.nf
prototaxites: I think this is a good idea - my only problem is that I'm not sure what files other classifiers might want in future. Currently the DOMAIN_CLASSIFICATION workflow takes the assembles, bins, and unbins, classifies the contigs for each assembly, and then uses the custom script to split the assembly contig classifications back into their bins. Other classifiers might work differently and require different files - if that's a problem we want to tackle if and when in future, happy to split Tiara into a separate workflow now, otherwise we could wait and see what a future tool requires?

workflows/mag.nf
prototaxites: `!(meta.domain in ["eukarya"])`

or 

`meta.domain !in ["eukarya"]`?

Not sure which is syntactically/stylistically correct!

docs/output.md
jfy133: OK fair enough! Then yes, a separate directory works :)

subworkflows/local/domain_classification.nf
jfy133: I would say tackle it when it comes later! CAn always add more input  channels to the subwf! :)

nextflow.config
jfy133: Fair enough, I guess would require larger overhaul which is out of scope.

workflows/mag.nf
d4straub: Both fine, I prefer the former.
But `meta.domain != "eukarya"` also does the trick and seems simpler

prototaxites: of course... 🤦‍♂️

conf/modules.config
jfy133: Ah yes good point! Nope, logic makes sense :+1:

prototaxites:
Also, a heads up that the new test_domain_classification.config test depends on the following PR in the test datasets repo:

https://github.com/nf-core/test-datasets/pull/782

jfy133 (approved):

jfy133:
As far as  I can tell LGTM !

PR #433, opened 2023-05-05
tax_classification_taxids is the channel that needs collect for CAT_summary

maxibor:
When running mag with the CAT contig taxonomic annotation tool, I stumbled upon the following error:

```bash
 Workflow execution completed unsuccessfully

Error executing process > 'NFCORE_MAG:MAG:CAT_SUMMARY'

Caused by:
  Process `NFCORE_MAG:MAG:CAT_SUMMARY` input file name collision -- There are multiple input files for each of the following file names: MEGAHIT-DASTool-SZG018.ORF2LCA.names.txt.gz, MEGAHIT-DASTool-SZG023.ORF2LCA.names.txt.gz, MEGAHIT-DASTool-SZG023.bin2classification.names.txt.gz, MEGAHIT-DASTool-SZG017.ORF2LCA.names.txt.gz, MEGAHIT-DASTool-SZG024.ORF2LCA.names.txt.gz, MEGAHIT-DASTool-SZG017.bin2classification.names.txt.gz, MEGAHIT-DASTool-SZG022.bin2classification.names.txt.gz, MEGAHIT-DASTool-SZG025.ORF2LCA.names.txt.gz, MEGAHIT-DASTool-SZG020.ORF2LCA.names.txt.gz, MEGAHIT-DASTool-SZG018.bin2classification.names.txt.gz, MEGAHIT-DASTool-SZG025.bin2classification.names.txt.gz, MEGAHIT-DASTool-SZG020.bin2classification.names.txt.gz, MEGAHIT-DASTool-SZG024.bin2classification.names.txt.gz, MEGAHIT-DASTool-SZG022.ORF2LCA.names.txt.gz
```

Looking more closely at the source of this error, I believe it because the CAT_SUMMARY process only expects `.bin2classification.names.txt.gz` while there are both `.ORF2LCA.names.txt.gz` and `.bin2classification.names.txt.gz` in the `tax_classification` channel. The `tax_classification_taxids` is the one containing only the `.bin2classification.names.txt.gz` files.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## v2.3.1dev - [date]
+## 2.4.0dev - [date]
 
 ### `Added`
 
@@ -38,6 +38,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 | -------- | ---------------- | ----------- |
 | BCFtools | 1.16             | 1.17        |
 | SAMtools | 1.16.1           | 1.17        |
+
+## v2.3.2 - [2023-06-23]
+
+### `Fixed`
+
+- [#461](https://github.com/nf-core/mag/pull/461) - Fix full-size AWS test profile paths (by @jfy133)
+- [#461](https://github.com/nf-core/mag/pull/461) - Fix pyDamage results being overwritten (reported by @alexhbnr, fix by @jfy133)
+
+## v2.3.1 - [2023-06-19]
+
+### `Fixed`
+
+- [#458](https://github.com/nf-core/mag/pull/458) - Correct the major issue in ancient DNA workflow of binning refinement being performed on uncorrected contigs instead of aDNA consensus recalled contigs (issue [#449](https://github.com/nf-core/mag/issues/449))
+- [#451](https://github.com/nf-core/mag/pull/451) - Fix results file overwriting in Ancient DNA workflow (reported by @alexhbnr, fix by @jfy133, and integrated by @maxibor in [#458](https://github.com/nf-core/mag/pull/458) )
 
 ## v2.3.0 - [2023/03/02]
 
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -493,8 +493,8 @@ process {
     }
 
     withName: BCFTOOLS_VIEW {
-        ext.args   = "-v snps,mnps -i 'QUAL>=${params.bcftools_view_high_variant_quality} || (QUAL>=${params.bcftools_view_medium_variant_quality} && FORMAT/AO>=${params.bcftools_view_minimal_allelesupport})'"
         ext.prefix = { "${meta.assembler}-${meta.id}.filtered" }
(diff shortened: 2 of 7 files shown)


maxibor:
The content of the two channels:

https://github.com/maxibor/mag/blob/3b031177c22c9314e8776b3c22341dc9a35da70f/modules/local/cat.nf#L13-L20

jfy133 (approved):
LGTM, but I've not touched CAT before. Will give some time for @d4straub or @skrakau to shout if something is wrong before merging :)

d4straub:
Hm, there seems to be a difference between
`[assembler]-[binner]-[sample/group].bin2classification.names.txt.gz: Taxonomy classification of the genome bins, with full lineage names`
and 
`[assembler]-[binner]-[sample/group].bin2classification.txt.gz: Taxonomy classification of the genome bins`
according to https://nf-co.re/mag/2.3.0/output#taxonomic-classification-of-binned-genomes
I do not remember though whether that is relevant by changing for example the summary files. Did you check?
If it does change resulting files, maybe the two different `*names* `files could be put into separate output channels?

jfy133 (commented):
Config should be reactivated now (see [here](https://github.com/nf-core/configs/pull/507)), but otherwise LGTM, 

I do notice that `ORF2LCA` and `bin2classification` is not any more picked up from `raw/`, - but I assume you've tested this

CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#439](https://github.com/nf-core/mag/pull/445) - Fix bug in assembly input (by @prototaxites)
 - [#447](https://github.com/nf-core/mag/pull/447) - Remove `default: None` from parameter schema (by @drpatelh)
```
jfy133, line 32:
```suggestion
```

tillenglert:
If I may add:

> Looking more closely at the source of this error, I believe it because the CAT_SUMMARY process only expects .bin2classification.names.txt.gz while there are both .ORF2LCA.names.txt.gz and .bin2classification.names.txt.gz in the tax_classification channel. The tax_classification_taxids is the one containing only the .bin2classification.names.txt.gz files.

I tracked the problem not to the duality of the input (.ORF2LCA.names.txt.gz and .bin2classification.names.txt.gz) but instead to the naming of the DASTool output/input for CAT. The unbinned input for CAT resulting from DASTool refinement is named [Assembler]-DASTool-[prefix].*, which is the same name for the binned input. Therefore creating not only the collision for `CAT_SUMMARY` but also overwriting the output in the results dir.

The log from CAT includes the number of bins (which should be always 1 for DASTool, I think?) and one could probably grep this number to rename the output of CAT accordingly.

I hope this helps! :)

jfy133:
> DASTool refinement is named [Assembler]-DASTool-[prefix].*, which is the same name for the binned input.

@tillenglert coudl you confirm that the `RENAME_POSTDASTOOL` module was executed?

If I understand what you identified, this _should_ be addressed by [this](https://github.com/nf-core/mag/blob/master/modules/local/rename_postdastool.nf) module...?

jfy133:
Ahh got you! Ok maybe that indeed makes sense, sorry:

```

    withName: CAT {
        publishDir = [
            path: { "${params.outdir}/Taxonomy/CAT/${meta.assembler}/${meta.binner}" },
            mode: params.publish_dir_mode,
            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
        ]
    }
    ```
    
  We would need to investigate if we ebed into the meta whether teh files are unbinned or not I guess, and then can insert that into the file name for CAT... 
  
  I see CAT currently doesn't use a prefix, but rather hard codes the file names in the module (another local module..) so this makes much more sense now :face_exhaling:

PR #435, opened 2023-05-08
Adds support for run merging for multi-run samples

### LINKED ISSUE #358: Allow to provide multiple FastQs per sample
### Description of feature

In case, there is more than one pair of FastQ files per sample, e.g. because there are multiple libraries of a sample or the same library of a sample was sequenced multiple times, it would be great if these could be automatically be merged prior to starting the assembly during the pre-processing and host removal steps.

This could be solved by introducing an additional column with a library identifier in the sample sheet that is used for concatenating the different libraries of a sample.

jfy133:
resolves #358 

## PR checklist


docs/usage.md
```diff
@@ -27,38 +27,39 @@ Please note the following additional requirements:
...
 
 ```bash
-sample,group,short_reads_1,short_reads_2,long_reads
-sample1,0,data/sample1_R1.fastq.gz,data/sample1_R2.fastq.gz,data/sample1.fastq.gz
-sample2,0,data/sample2_R1.fastq.gz,data/sample2_R2.fastq.gz,data/sample2.fastq.gz
-sample3,1,data/sample3_R1.fastq.gz,data/sample3_R2.fastq.gz,
+sample,run,group,short_reads_1,short_reads_2,long_reads
+sample1,1,0,data/sample1_R1.fastq.gz,data/sample1_R2.fastq.gz,data/sample1.fastq.gz
+sample2,1,0,data/sample2_R1.fastq.gz,data/sample2_R2.fastq.gz,data/sample2.fastq.gz
+sample3,1,1,data/sample3_R1.fastq.gz,data/sample3_R2.fastq.gz,
```
jfy133, line 46:
```suggestion
sample3,,1,data/sample3_R1.fastq.gz,data/sample3_R2.fastq.gz,
```

jfy133:
@alexhbnr any chance you could test this?

```bash
nextflow run jfy133/nf-core-mag -r run-merging
```

d4straub (commented):
Why do you need a column "run" when it would be easier (from a user perspective) to just use twice the same "sample" id?
I am not into that topic of multiple files merging per sample, so forgive my naive question.

jfy133:
> Why do you need a column "run" when it would be easier (from a user perspective) to just use twice the same "sample" id? I am not into that topic of multiple files merging per sample, so forgive my naive question.

A run could also potentially be used to indicate separate libraries, and/or one library sequenced across multiple sequencing flow cells.

In both cases, you may want to evaluate the sequencing quality and adpater removal performance independently in case a flowcell/library failed for some reason.

d4straub:
I understand your reasoning (and agree that this seems helpful) but I dislike the additional required column that in many cases will be useless. If someone else thinks thats the way its done best and wants to give an ok, I can live with it but I would prefer a non-required column "run".

jfy133:
I sort of agree, but one could argue the same with the groups column ;).

I'll look to see if I can make the column optional first but if not I will cite eager as a precedent 😬.

d4straub:
Ampliseq allows optional column "run" and "reverseReads" in [parse_input.nf](https://github.com/nf-core/ampliseq/blob/master/subworkflows/local/parse_input.nf) documented [here](https://nf-co.re/ampliseq/2.5.0/usage#samplesheet-input). Not sure if that is helpful in this case though.

jfy133:
> Ampliseq allows optional column "run" and "reverseReads" in [parse_input.nf](https://github.com/nf-core/ampliseq/blob/master/subworkflows/local/parse_input.nf) documented [here](https://nf-co.re/ampliseq/2.5.0/usage#samplesheet-input). Not sure if that is helpful in this case though.


Oh that looks really nice though! I'll definitely look into that and borrow!

jfy133:
@d4straub inspired by ampliseq (thanks for that!), this is ready for re-review
@alexhbnr please test!

jfy133:
Should also have option to save the run-merged FASTQ files

alexhbnr (approved):
I tested this PR and it performs the concatenation of multiple FastQ files prior to the assembly step as intended.

docs/usage.md
```diff
@@ -53,12 +54,22 @@ sample1,0,data/sample1.fastq.gz,,
 sample2,0,data/sample2.fastq.gz,,
 ```
 
+or to additionally to perform run merging of two runs of sample1:
+
+```bash
```
prototaxites, line 59: The documentation above mentions using a `run` column, but the example here excludes it - is it accidentally misnamed `group`? (it varies within the sample). Does group need to be additionally specified as well?

docs/usage.md
```diff
@@ -53,12 +54,22 @@ sample1,0,data/sample1.fastq.gz,,
...
+sample,group,short_reads_1,short_reads_2,long_reads
+sample1,1,data/sample1_R1.fastq.gz,data/sample1_R2.fastq.gz,data/sample1.fastq.gz
+sample1,2,data/sample1_R1.fastq.gz,data/sample1_R2.fastq.gz,data/sample1.fastq.gz
+sample2,0,data/sample2_R1.fastq.gz,data/sample2_R2.fastq.gz,data/sample2.fastq.gz
+sample3,1,data/sample3_R1.fastq.gz,data/sample3_R2.fastq.gz,
```
jfy133, line 64:
```suggestion
sample,run,group,short_reads_1,short_reads_2,long_reads
sample1,1,0,data/sample1_R1.fastq.gz,data/sample1_R2.fastq.gz,data/sample1.fastq.gz
sample1,2,0,data/sample1_R1.fastq.gz,data/sample1_R2.fastq.gz,data/sample1.fastq.gz
sample2,0,0,data/sample2_R1.fastq.gz,data/sample2_R2.fastq.gz,data/sample2.fastq.gz
sample3,1,0,data/sample3_R1.fastq.gz,data/sample3_R2.fastq.gz,
```

docs/usage.md
jfy133: Oops, good call. Forgot to include the _additional_ run column

prototaxites:
Tested and it works well - runs merged happily in the test profile. 

There are conflicts now since adding the assembly input option. At the moment the pipeline skips read all pre-processing when supplying assemblies as input - do we want to assume that a user using this option would also supply merged-run fastq files (likely if they've already done the pre-processing), or is it preferable to allow run merging in this case (QC would still have to be skipped!)?

jfy133:
> Tested and it works well - runs merged happily in the test profile.
> 
> There are conflicts now since adding the assembly input option. At the moment the pipeline skips read all pre-processing when supplying assemblies as input - do we want to assume that a user using this option would also supply merged-run fastq files (likely if they've already done the pre-processing), or is it preferable to allow run merging in this case (QC would still have to be skipped!)?

I would consider if you've already got to the assembly level, you would already have merged run your reads, so I think we can make our lives easier here ;). I've added a docs note to that effect.

jfy133:
Conflicts resolved and lightly tested (ran with and withour run-merging samplesheet), and documented the QC skipping @prototaxites !

prototaxites:
test_binnning_entry fails: `No such variable: ch_short_reads`

Looks like you've dropped the 

```} else {
        ch_short_reads = ch_raw_short_reads
    }```

from the `!assembly_input` if block!

**Comment by prototaxites** [comment]
https://github.com/nf-core/mag/blob/dev/conf/test_binning_entry.config

Should be available with -profile test_binning_entry!

**Comment by jfy133** [comment]
I've added it to the CI tests (now we have more capacity with nf-core),

but now reached the dreaded 

```
groovy.lang.MissingMethodException: No signature of method: Script_84f765ce$_runScript_closure1$_closure2$_closure11.call() is applicable for argument types: (ArrayList) values: [[test_minigut, null, 0, /nf-core/test-datasets/raw/mag/test_data/test_minigut_R1.fastq.gz, ...]]
```

error :facepalm:

jfy133:
Self note in case I don't have time to fix: OK, I've identified the issue. In the preprocessnig subworkflow we remove the `run` column at some point so everything during/after assembly does not have this. But if we are running with binning entry, we skip the preprocessing step entirely so the removal of that metadata column doesn't happen. Need to update the assignment so that meta entry gets dropped

jfy133:
Now it's because the `ch_short_reads` includes `false` at the end, so I guess somewhere the long-reads bit of the channel isn't being removed

prototaxites:
LGTM!

prototaxites (approved):

PR #436, opened 2023-05-09
Make GTDB downloading skippable using standard parameter name

jfy133:
To explicitly skip GTDB downloading, you would have to supply `--gtdb false`. This is not inline with other skipping parameters in the pipeline, and uses a nextflow specific method (supplying `false` to an otherwise string parameter) that may not be logical to a user.

This PR separates the 'activation' (or rather skipping) of the module from the URL,  so one can explicitly turn skip that step. This is benefitial as the GTDB file can be large and take a long time to download (78GB from a server in Australia, which can be quite far for many users).

## PR checklist


jfy133:
It appears there is something going wrong where the GTDB DB isn't being passed to the module and so the summary file is always empty...

investigateign

jfy133:
I'm hitting lots of problems now e.g., when skipping things, I think I might start from scratch but need fresh eyes

jfy133:
Should also make sure all download steps are slipped when that whole section of workflow is skipped (s.g.  aria2/checkM and binning)

jfy133:
I'm going to start from scratch given the conflicts with the dev bvranch

Issue #431, opened 2023-05-04
Entry point to pipeline at binning stage


prototaxites:
### Description of feature

Assembly can be a time and resource-intensive process, and if returning to an old pipeline run (or using an assembly supplied by someone else), it would be useful to be able to skip the assembly stage altogether and use existing files. This would allow users to use the updated binning workflows from newer versions of the pipeline on assemblies generated in older pipeline runs.

Could implement this by allowing the user to supply a CSV of assemblies to the pipeline, as well as the standard input sheet which should contain the reads (post-QC, host removal, etc.) that were used to build the assemblies. The assembly CSV should have a group field which should match with the group identifiers listed in the reads sheet.

artur-matysik:
It would be fantastic to have this! 

As a quick and dirty workaround I was using something like:
- added `params.assembly_done`. If true, workflow is expecting `params.inputAssemblies`
- assembly enclosed in `if (!params.assembly_done) { ... }`
- and if assembly is done already (`else {...}`) assemblies channels `ch_assemblies` is defined as below, using provided
`params.inputAssemblies`

```
// Extract sample ID
        def get_id = { it ->
            def startDelimiter = it.split("-")[0] + "-"
            def endDelimiter = "_scaffolds.fasta"
            def startIndex = it.indexOf(startDelimiter) + startDelimiter.length()
            def endIndex = it.indexOf(endDelimiter)
            def result = it.substring(startIndex, endIndex)
            return result
        }
    
        ch_assemblies = Channel
            .fromPath(params.inputAssemblies)
            .map { row -> tuple ( row.getBaseName(), row )} 
            .map { row -> tuple ( [id: get_id(row[0]), group: 0, single_end: false, assembler: row[0].split("-")[0]], row[1] ) }
```

PR #439, opened 2023-05-10
Add entry point to pipeline at binning stage

prototaxites:
Resolves https://github.com/nf-core/mag/issues/431. Hope the documentation is clear! Will create a separate PR in the test-datasets repo to add the assemblies and assembly samplesheet for the test config.

## PR checklist

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -78,15 +78,38 @@ Where `id` is the ID of the assembly, `assembler` is the assembler used to produ
 
 The value of the `id` column should be set depending on the desired `--binning_map_mode`. If `--binning_map_mode` is set to "own", then the values should match those in the `id` column of the `--input` CSV. If `--binning_map_mode` is set to "group", then the values should match those in the `group` column of a `--input` CSV.
 
-Valid examples could look like the following:
+For the below `--input` CSV, valid options could be:
+
+```bash
+sample,group,short_reads_1,short_reads_2,long_reads
+sample1,group0,data/sample1_R1.fastq.gz,data/sample1_R2.fastq.gz,data/sample1.fastq.gz
+sample2,group0,data/sample2_R1.fastq.gz,data/sample2_R2.fastq.gz,data/sample2.fastq.gz
+sample3,group1,data/sample3_R1.fastq.gz,data/sample3_R2.fastq.gz,
+```
+
+If using `--binning_map_mode "own"`:
 
 ```bash
 id,assembler,fasta
-0,MEGAHIT,MEGAHIT-test_minigut.contigs.fa.gz
-0,SPAdes,SPAdes-test_minigut_contigs.fasta.gz
-0,MEGAHIT,MEGAHIT-test_minigut_sample2.contigs.fa.gz
-0,SPAdes,SPAdes-test_minigut_sample2_contigs.fasta.gz
+sample1,MEGAHIT,MEGAHIT-sample1_contigs.fa.gz
+sample1,SPAdes,SPAdes-sample1_contigs.fasta.gz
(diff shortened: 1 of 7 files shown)


jfy133 (approved):
I think this looks good to me (a few documentation clarifications).

The code is getting rather complex now with many nested if/else statements, I think alot of this needs to be refactorered into more subworkflows as it's getting hard to read now... but that's for another time! Thanks @prototaxites - It hink this is going to be really popular functionality :)

workflows/mag.nf
```diff
@@ -423,118 +433,122 @@ workflow MAG {
...
-            .filter { it[0].single_end }
-            .map { meta, reads -> [ meta, [ reads ], [] ] }
-            .mix (
-                ch_short_reads_assembly
-                    .filter { ! it[0].single_end }
-                    .map { meta, reads -> [ meta, [ reads[0] ], [ reads[1] ] ] }
-            )
-        ch_long_reads_grouped = ch_long_reads
-    }
+    if ( !params.assembly_input ) {
```
d4straub, line 436: Here the steps are skipped with if statements. That is clean, because the terminal output will not show those skipped processes, but it leads to such convoluted if else stuff. While I typically prefer the if statements as well, here it might be better to work with empty channels to skip steps. I didnt think the idea through, but if it would make the code much easier to read, I would rather prefer to use here empty-channel-skipping than if-skipping.

subworkflows/local/input_check.nf
```diff
@@ -66,13 +66,75 @@ workflow INPUT_CHECK {
...
+                        def assembler = row.assembler ?:  false
+                        def assembly  = row.fasta ? file(row.fasta, checkIfExists: true) : false
+                        // Check if given combination is valid
+                        if (!assembly) exit 1, "Invalid input assembly samplesheet: fasta can not be empty."
+                        if (!assembler) exit 1, "Invalid input assembly samplesheet: assembler can not be empty."
+                        return [ id, assembler, assembly ]
+                    } else {
+                        exit 1, "Input samplesheet contains row with ${row.size()} column(s). Expects 3."
+                    }
+                }
```
d4straub, line 88: it seems to me some indents are not right, e.g. line 72 might need one more, line 88 one less?

docs/usage.md
```diff
@@ -66,6 +66,30 @@ Please note the following requirements:
...
+
+Where `id` is the ID of the assembly, `assembler` is the assembler used to produce the assembly (one of `MEGAHIT`, `SPAdes`, or `SPAdesHybrid`), and `fasta` is the path to the assembly fasta file.
+
+The value of the `id` column should be set depending on the desired `--binning_map_mode`. If `--binning_map_mode` is set to "own", then the values should match those in the `id` column of the `--input` CSV. If `--binning_map_mode` is set to "group", then the values should match those in the `group` column of the `--input` CSV.
+
+Valid examples could look like the following:
+
+```bash
+id,assembler,fasta
+0,MEGAHIT,MEGAHIT-test_minigut.contigs.fa.gz
```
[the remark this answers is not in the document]
prototaxites: That would be for `group` - perhaps a better idea here would be to supply the example test `--input` CSV, and then show how to do both `own` and `group`. Probably should also include a line saying that for `all`, it doesn't matter if they match!

subworkflows/local/input_check.nf
prototaxites: Had another look, looks correct to me - happy to be proven wrong though!

workflows/mag.nf
d4straub: Alright. I wont have time to review that current mess though, so go with it, you have James' ok.

Issue #449, opened 2023-05-22
Genome binning is performed on uncorrected contigs when selecting the ancient DNA subworkflow


alexhbnr:
### Description of the bug

When selecting the ancient DNA sub-workflow using `--ancient_dna`, a correction of the consensus sequence reported by the assembler is performed. This step should remove artefacts that were wrongly called by the assembler due to the presence of ancient DNA damage.

However, for genome binning, nf-core/mag doesn't select these corrected contigs for the binning but uses the non-corrected contigs instead. This defeats the purpose of enabling the `--ancient_dna` sub-workflow.

### System information

- version of nf-core/mag: 2.3.0

jfy133:
@alexhbnr could you provide more information - as far as I can see what you want is already implemented, as in the BCFtools consensus output (presumably with the corrected bases) is then passed to `binning` via the 'contigs_recalled` channel name:

![image](https://github.com/nf-core/mag/assets/17950287/f4a64e43-ad41-41d2-9d5b-ef3a13cb0609)
![image](https://github.com/nf-core/mag/assets/17950287/80e490f3-8d23-4305-b857-83edbe90fe24)

Can you provide an example somehow (or a reprex) where the wrong contigs were used downstream?

alexhbnr:
In case, it's already in the `dev` branch and it seems to work, then that's what I want. In the major release v2.3.0 that I used the contig sequences found in the final MAGs were identical to the contigs found in the uncorrected samples. That's why I raised this issue.

jfy133:
That's concerning, I'm pretty sure when I looked earlier the code looked identical. Which mags were you looking at in the results directory (presumably)

alexhbnr:
Here is the code that I used to start the pipeline:
```
nextflow run nf-core/mag -r 2.3.0 \
            -profile eva,archgen \
            --input 04-analysis/Zape2/nfcore_mag_samplesheet.csv \
            --outdir 04-analysis/Zape2/assembly \
            --skip_clipping \
            --skip_prodigal \
            --binning_map_mode own \
            --min_contig_size 1000 \
            --gtdb false \
            --binqc_tool checkm \
            --save_checkm_data \
            --refine_bins_dastool \
            --postbinning_input both \
            --run_gunc \
            --gunc_save_db \
            --ancient_dna
```

[...]

PR #459, opened 2023-06-16
Make ancient DNA damage correction optional

### LINKED ISSUE #456: Split ancientDNA sub-workflow into separate sub-workflows for running damage analysis and consensus correction
### Description of feature

Currently, enabling the `ancientDNA` workflow automatically enables the correction of the consensus sequence reported by the *de novo* assembler and the analysis of the presence of ancient DNA damage. The first step is mainly driven by the issue that we observed that MEGAHIT generates contig sequences that are riddled with nonsense mutations or stop codons when ancient DNA damage is present but were not supported by the short-read data.

When assembling with metaSPAdes, we do not observe such a behaviour and therefore the consensus correction might not be necessary. However, we still would like to run the evaluation of the amount of aDNA damage present using pyDamage.

It would be very helpful, if the two modules in this sub-workflow could be separated from each other, so they can be switched on and off separately.

jfy133:
This makes the damage correction subworkflow optional, i.e. ytou can run just pyDamage, and optionally turn on damage correction using freebayes and bcftools.

@alexhbnr please can you test, and let me know if you would refer to have it opt in (rather than opt out).

Also note this _does not yet_ include the hotfix in the patch for #449 , I will add that later as it requires more thinking due to the more complex inputs to bin refinement since domain classification has come in.

fixes #456

## PR checklist

### DIFF
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -521,7 +521,7 @@ process {
     withName: PYDAMAGE_ANALYZE {
         ext.prefix = { "${meta.assembler}-${meta.id}" }
         publishDir = [
-            path: {"${params.outdir}/Ancient_DNA/pydamage/analyze/" },
+            path: {"${params.outdir}/Ancient_DNA/pydamage/analyze/"${meta.assembler}-${meta.id}" },
             mode: params.publish_dir_mode
         ]
     }
@@ -530,7 +530,7 @@ process {
         ext.prefix = { "${meta.assembler}-${meta.id}" }
         ext.args   = "-t ${params.pydamage_accuracy}"
         publishDir = [
-            path: {"${params.outdir}/Ancient_DNA/pydamage/filter/" },
+            path: {"${params.outdir}/Ancient_DNA/pydamage/filter/"${meta.assembler}-${meta.id}" },
             mode: params.publish_dir_mode
         ]
     }
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -521,7 +521,7 @@ process {
     withName: PYDAMAGE_ANALYZE {
         ext.prefix = { "${meta.assembler}-${meta.id}" }
         publishDir = [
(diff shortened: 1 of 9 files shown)


alexhbnr (changes requested):
The skipping of the contig correction itself works well. However, when activating `--skip_ancient_damagecorrection` pyDamage doesn't export the files into a sample specific sub-folder any longer. The usual way was that pyDamage has sample-specific sub-folders in `Ancient_DNA/pydamage/analyze` and `Ancient_DNA/pydamage/filter`. At the moment, for every sample the same file is used and the results of the previous run sample are overwritten.

conf/modules.config
```diff
@@ -521,7 +521,7 @@ process {
     withName: PYDAMAGE_ANALYZE {
         ext.prefix = { "${meta.assembler}-${meta.id}" }
         publishDir = [
-            path: {"${params.outdir}/Ancient_DNA/pydamage/analyze/" },
+            path: {"${params.outdir}/Ancient_DNA/pydamage/analyze/"${meta.assembler}-${meta.id}" },
```
jfy133, line 524:
```suggestion
            path: {"${params.outdir}/Ancient_DNA/pydamage/analyze/${meta.assembler}-${meta.id}" },
```

conf/modules.config
```diff
@@ -530,7 +530,7 @@ process {
         ext.prefix = { "${meta.assembler}-${meta.id}" }
         ext.args   = "-t ${params.pydamage_accuracy}"
         publishDir = [
-            path: {"${params.outdir}/Ancient_DNA/pydamage/filter/" },
+            path: {"${params.outdir}/Ancient_DNA/pydamage/filter/"${meta.assembler}-${meta.id}" },
```
jfy133, line 533:
```suggestion
            path: {"${params.outdir}/Ancient_DNA/pydamage/filter/${meta.assembler}-${meta.id}" },
```

jfy133:
Fix applied, waiting from results on tests as discussed on slack.

alexhbnr (approved):

PR #489, opened 2023-08-08
Fix for `NFCORE_MAG:MAG:CAT_SUMMARY input file name collision`

maxibor:
This PR brings a fix to what was first described in #433 
The issue happened when using `--postbinning_input both` in combination with domain annotation, bin_refinement and CAT profiling.
This is because bins annotated as eukaryotic are sent to the refined bins channels as is, bypassing dastool, and hence keeping the same name.
This fix adds a "refined" entry to the meta, and includes it in the new prefix for the CAT process.

## PR checklist


subworkflows/local/binning_refinement.nf
```diff
@@ -68,6 +68,7 @@ workflow BINNING_REFINEMENT {
     ch_versions = ch_versions.mix(DASTOOL_FASTATOCONTIG2BIN_MAXBIN2.out.versions.first())
     ch_versions = ch_versions.mix(DASTOOL_FASTATOCONTIG2BIN_CONCOCT.out.versions.first())
 
+
```
jfy133, line 71:
```suggestion
```

workflows/mag.nf
```diff
@@ -736,6 +736,15 @@ workflow MAG {
...
         */
 
+        ch_binning_results_bins = BINNING.out.bins
+                .map { meta, bins ->
+                    [meta + ['refined':'unrefined'], bins]
+                }
+            ch_binning_results_unbins = BINNING.out.unbinned
+                .map { meta, bins ->
+                    [meta + ['refined':'unrefined'], bins]
+                }
```
jfy133, line 746:
```suggestion
        ch_binning_results_unbins = BINNING.out.unbinned
            .map { meta, bins ->
                [meta + ['refined':'unrefined'], bins]
            }
```

Indendentation guessed, please double check :sweat_smile:

jfy133 (commented):
A few minor comments, but a more general one, won't most of the downstream steps that can take in output from bin refinement also need update to their prefixes (E.g. in `modules.conf`)?

CarsonJM (approved):
🚀

nextflow.config
```diff
@@ -208,7 +208,7 @@ try {
 
 profiles {
     debug {
-        dumpHashes             = true
+        dumpHashes             = false
```
tillenglert, line 211: Was this change intended? If yes I'm not sure if this general debugging option should be changed hidden away in another topics PR!

workflows/mag.nf
```diff
@@ -736,6 +736,15 @@ workflow MAG {
         * DAS Tool: binning refinement
         */
 
+        ch_binning_results_bins = BINNING.out.bins
+                .map { meta, bins ->
+                    [meta + ['refined':'unrefined'], bins]
```
tillenglert, line 741: I think as someone rather new to the pipeline code a meta entry: 'refined':'unrefined' is rather confusing. Maybe the entry should be named something like 'refinement' or 'refinement_tag' to make it more verbose.

tillenglert (approved):
LGTM! Had some minor comments/ideas though. Thanks for the implementation and Bugfix!

nextflow.config
jfy133: No it shouldn't be changed,it's from template... I had also left a comment in my review on this but apparently didnt' save(?!)

workflows/mag.nf
jfy133: Agreed! Pr maybe 'bin-status' or if you prefer `un/refined` as a string? If you want to keep hte meta ID as `refined` it could be made boolean (but this will be more complicated for inserting into text

jfy133:
@maxibor (or anyone else) any thoughts on my original feedback:

> <...>  but a more general one, won't most of the downstream steps that can take in output from bin refinement also need update to their prefixes (E.g. in modules.conf)?

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#449](https://github.com/nf-core/mag/pull/447) - Fix results file overwriting in Ancient DNA workflow (reported by @alexhbnr, fix by @jfy133)
 - [#470](https://github.com/nf-core/mag/pull/470) - Fix binning preparation from running even when binning was requested to be skipped (reported by @prototaxites, fix by @jfy133)
 - [#480](https://github.com/nf-core/mag/pull/480) - Improved `-resume` reliability through better meta map preservation (reported by @prototaxites, fix by @jfy133)
+- [#489](https://github.com/nf-core/mag/pull/481) - Fix `NFCORE_MAG:MAG:CAT_SUMMARY input file name collision`
```
jfy133, line 41:
```suggestion
- [#489](https://github.com/nf-core/mag/pull/481) - Fix `NFCORE_MAG:MAG:CAT_SUMMARY input file name collision` (reported and fixed by @maxibor )
```

maxibor:
Update: this is actually a more generalized issue with processes that perform a collect: CAT, but also Quast, and GTDB-Tk

jfy133:
> @maxibor (or anyone else) any thoughts on my original feedback:
> 
> > <...>  but a more general one, won't most of the downstream steps that can take in output from bin refinement also need update to their prefixes (E.g. in modules.conf)?

I guess this was correct then???

I wonder why this issue hasn't come up before...?!

maxibor:
It's a new issue because samples can now appear multiple times in downstream channels for the following reasons:
- Split by domain (new, only happends when using domain classification)
- Split by refined/not refined (only happens when using bin refinement)
- Split by binned/unbinned (only happens when using `--postbinning_input both`)

When they're collected at the final steps (Quast, GTDB-Tk, or CAT), bins that were not renamed by the way they were treated now appear multiple times with same name, and triggers the input file name collision.

Which in addition was particularly challenging to debug because of a misreporting of the actual error https://github.com/nextflow-io/nextflow/issues/3828

prototaxites:
I’ve been thinking about this and I’m not sure we need the refined/unrefined label to fix this - this might specifically just be a bug in the code that assigns bins post-refinement.

If we think about the original sources of bins, we have bins coming from the three binners - metabat, maxbin, and concoct. Each of these bins should have a unique numbered Coleman’ file name. If refinement is enabled, these bins are fed into DAS tool and a new set of bins are returned - that are not necessarily the same as the input bins. These should also be uniquely named, differently to the original bins because their binner is listed as DAS tool.

If domain classification is enabled with bin refinement, then undefined eukaryotic bins are passed as “refined” at this stage as otherwise they will be lost, unless ‘both’ is set for post-binning input. I think this is the only time that a duplicate bin can slip through. So don’t we just have to fix the code so that eukaryotic bins are only fed to the refined bin channel if ‘both’ is not selected? Otherwise, every bin should be unique and only be able to enter each process downstream exactly once, because they might get split into separate channels, and recombined, but never duplicated?

maxibor:
A small schematic to illustrate the issue.
The bins do not need to be renamed, only the results of GTDB-Tk, Quast, and CAT. Otherwise the `.collect()` for the summary process of these three tools will fail with an input file name collision.
mag meta flowthrough_230828_152417

I've now update the meta's accordingly, and passed them through when needed.

maxibor:
There are now 4 possible values for the `refinement` meta:
- `unrefined` for bins that won't be refined
- `unrefined_unbinned` for the contigs that were not binned and will not be refined
- `dastool_refined` for bins that were refined with DASTool
- `dastool_refined_unbinned` for the `unbinned` processed with DASTool

Note that the `input file name collision` was not only be triggered by the lack of refinment tracking, but also by the lack of domain tracking.

jfy133:
That's a really helpful diagram @maxibor thank you very much!

subworkflows/local/depths.nf
```diff
@@ -23,17 +23,18 @@ workflow DEPTHS {
     main:
     ch_versions = Channel.empty()
 
+
```
jfy133, line 26:
```suggestion
```

subworkflows/local/gtdbtk.nf
```diff
@@ -82,8 +82,10 @@ workflow GTDBTK {
     GTDBTK_SUMMARY (
         ch_filtered_bins.discarded.map{it[1]}.collect().ifEmpty([]),
         GTDBTK_CLASSIFYWF.out.summary.collect().ifEmpty([]),
-        GTDBTK_CLASSIFYWF.out.filtered.collect().ifEmpty([]),
-        GTDBTK_CLASSIFYWF.out.failed.collect().ifEmpty([])
+        [],
+        // GTDBTK_CLASSIFYWF.out.filtered.collect().ifEmpty([]),
+        []
+        // GTDBTK_CLASSIFYWF.out.failed.collect().ifEmpty([])
```
jfy133, line 88:
```suggestion
        GTDBTK_CLASSIFYWF.out.filtered.collect().ifEmpty([]),
        GTDBTK_CLASSIFYWF.out.failed.collect().ifEmpty([])
```

Still needed, no? I assume this was from debugging

jfy133 (commented):
@maxibor 

- For GTDBK: you can't update nf-core modules without adding a patch. However in this case I think we can move the rename to `ext.prefix` definition to `modules.conf`.   This is the cause of the linting error. We need to fix the module though... (I'll tryand do that now)
- Missing docs update

But otherwise looks OK to me. I'm concerned we may have same collision errors e.g. with BUSCO, CheckM, Prokka... could that be the case? Or do you think it's safe?

The rule of thumb is: everything that happening at the sample level (directly, or with a .collect() after processing each bin) will likely trigger the same error. Everything that is happening only at the bin level should be fine.
Checkm is taken care of in `modules.config`, Busco is running per bin, Prokka, I haven't tested yet (I use https://github.com/maxibor/corephylo)

subworkflows/local/gtdbtk.nf
maxibor: In the later/st release of GTDB-Tk, everying is now reported in the summary file.

maxibor:
> @maxibor
> 
>     * For GTDBK: you can't update nf-core modules without adding a patch. However in this case I think we can move the rename to `ext.prefix` definition to `modules.conf`.   This is the cause of the linting error. We need to fix the module though... (I'll tryand do that now)
> 
>     * Missing docs update
> 
> 
> But otherwise looks OK to me. I'm concerned we may have same collision errors e.g. with BUSCO, CheckM, Prokka... could that be the case? Or do you think it's safe ?


The rule of thumb is: everything that happening at the sample level (directly, or with a .collect() after processing each bin) will likely trigger the same error. Everything that is happening only at the bin level should be fine.
Checkm is taken care of in `modules.config`, Busco is running per bin, as well as Prokka and MetaEuk (there's a `.transpose()`)

jfy133:
Watiing on: https://github.com/nf-core/modules/pull/3791

prototaxites:
I’m on annual leave and don’t have time to do a review (at least for a couple of weeks), but just wanted to follow up on this:

>The rule of thumb is: everything that happening at the sample level (directly, or with a .collect() after processing each bin) will likely trigger the same error. Everything that is happening only at the bin level should be fine.

In the case of bin-level processes, should we potentially be worried that some bins will be duplicated and the same process will run multiple times for the same bin?

maxibor:
> I’m on annual leave and don’t have time to do a review (at least for a couple of weeks), but just wanted to follow up on this:
> 
> > The rule of thumb is: everything that happening at the sample level (directly, or with a .collect() after processing each bin) will likely trigger the same error. Everything that is happening only at the bin level should be fine.
> 
> In the case of bin-level processes, should we potentially be worried that some bins will be duplicated and the same process will run multiple times for the same bin?

The only "duplication" that I can think of is when bith refined and unrefined bins are processed further. But arguably, refined bins are not the same anymore (at least in the filename).

jfy133 (commented):
With the docs update: Assuming that TIARA and QUAST gets the output file names based on the input files, I think this looks almost good to me now! Just need to get the working update GTDBTK module :)

subworkflows/local/gtdbtk.nf
jfy133: the commetned out bits can be removed (I guess once the module is implemnented :+1: )

jfy133:
I am currently downloading the latest GTDB database, then I'll do a manual 'full test' and if all looks good then we can merge!

jfy133:
- [x] CheckM looks good:  summary seems to be raw bins (`MEGAHIT-MetaBAT2-test_minigut_sample2.unbinned.1`) but per CheckM run looks good
- [x] QUAST looks good: summary seems to be raw bins (`MEGAHIT-CONCOCT-test_minigut_67.fa`) but per QUAST run looks good
- [x] CAT looks good: summary seems to be raw bins (`MEGAHIT-CONCOCT-test_minigut_sample2_24.fa`) but per CAT run output has correct names
- [ ] GTDBTK looks good: couldn't test due to db issue

This can be merged once I've done one test without domain classsifcation, and I'll follow up with a few other things I noticed with a separate PR

jfy133:
I propose merge this, patch release, then I'll make issues and a further patch release of the other more minor things I found

jfy133:
I'm happy with this now after some testing! Just need pull upstream from dev and CHANGELOG update :)

jfy133:
Ok, there are multiple issues but not related directly to problem this PR is solving.

We will break this down bit-by-bit so merging this now, and fixes will come in over time :+1: 

Thanks again @maxibor !

jfy133:
@nf-core-bot fix linting

subworkflows/local/depths.nf
```diff
@@ -19,21 +19,34 @@ workflow DEPTHS {
...
         .transpose()
+        .map {
+            meta_combine, meta, bins, depth ->
+            def meta_new = meta - meta.subMap('domain','refinement')
+            [meta_new, bins, depth]
+        }
         .groupTuple(by: [0,2])
+        .map {
+            meta, bins, depth ->
+            [meta, bins.unique(), depth]
```
jfy133, line 46:
```suggestion
            [meta, bins.unique().flatten(), depth]
```

jfy133:
Just need to remove some dumps, but hopefully the `.flatten()` of the bins going into depth fixes the bug

subworkflows/local/busco_qc.nf
```diff
@@ -65,6 +65,14 @@ workflow BUSCO_QC {
...
     }
 
+    busco_summary_domain = BUSCO.out.summary_domain.collect()
+    busco_summary_specific = BUSCO.out.summary_specific.collect()
+    busco_failed_bin = BUSCO.out.failed_bin.collect()
+
+    busco_summary_domain.dump(tag: 'busco_summary_domain', pretty: true)
+    busco_summary_specific.dump(tag: 'busco_summary_specific', pretty: true)
+    busco_failed_bin.dump(tag: 'busco_failed_bin', pretty: true)
+
```
jfy133, line 75:
```suggestion
```

subworkflows/local/depths.nf
```diff
@@ -19,21 +19,34 @@ workflow DEPTHS {
     main:
     ch_versions = Channel.empty()
 
+
+    bins_unbins.dump(tag: 'bins_unbins', pretty: true)
```
jfy133, line 23:
```suggestion
```

subworkflows/local/depths.nf
```diff
@@ -19,21 +19,34 @@ workflow DEPTHS {
...
+            def meta_new = meta - meta.subMap('domain','refinement')
+            [meta_new, bins, depth]
+        }
         .groupTuple(by: [0,2])
+        .map {
+            meta, bins, depth ->
+            [meta, bins.unique().flatten(), depth]
+        }
+
+    ch_depth_input.dump(tag: 'ch_depth_input', pretty: true)
```
jfy133, line 49:
```suggestion
```

subworkflows/local/gtdbtk.nf
```diff
@@ -46,6 +47,9 @@ workflow GTDBTK {
             }
     }
 
+    ch_bin_metrics.dump(tag: 'ch_bin_metrics', pretty: true)
+    bins.transpose().map { meta, bin -> [bin.getName(), bin, meta]}.dump(tag: 'ch_filtered_bins', pretty: true)
```
jfy133, line 51:
```suggestion
```

maxibor:
Long time in the making update:
That'll would be the pattern for better solution regarding the input file name collision

```nextflow
new_channel = channel_name .collectFile(keepHeader: true) {
        meta, f ->
        [f.getBaseName(), f]
    }
WHATEVER_SUMMARY_PROCESS(new_channel.collect())
```

This pattern takes all the files that share the same name and concatenates them into one (and `keepHeader` makes sure only header is written).
It's the equivalent of 

```bash
cat a/test.tsv  b/test.tsv > test.tsv
```

jfy133:
> Long time in the making update: That'll would be the pattern for better solution regarding the input file name collision
> 
> ```nextflow
> new_channel = channel_name.collectFile(keepHeader: true) {
>         meta, f ->
>         [f.getBaseName()+'.tsv', f]
>     }
> WHATEVER_SUMMARY_PROCESS(new_channel.collect())
> ```
> 
> This pattern takes all the files from the `channel_name` channel that share the same name and concatenates them into one (and `keepHeader` makes sure only header is written). It's the equivalent of
> 
> ```shell
> cat a/test.tsv  b/test.tsv > test.tsv
> ```

Hm interesting... but do we want to collapse those bins back into a single bin again? I'm pretty sure of the tools rely on the file name to disambiguate between bins so I'm worried we would then not be able to distinguish between eukaryotic vs non-eukaryotic bins in the summary tools.

Furthermore, does it support gzipped files? The original error you reported are gzipped output from CAT and I don't see reference to that in the NXF docs on `collectFile`

That said, I may be being overly paranoid there. So I suggest that  I merge this current PR as a conservative approach (making sure everything is very explicitly independent). But then I will make an issue describing what you're suggesting as an enhancement - as other than my (maybe unfounded) worry above - that would indeed be an elegant solution :+1:

PR #497, opened 2023-08-29
Adds --krona_db parameter to point at a local copy of the db needed for Krona. Suggested in Issue #404. 

### LINKED ISSUE #404: Add support to supplying local copy of Krona taxonomy database
### Description of feature

Currently either you skip Krona entirely, or you're forced to run a Krona  download step when using containers.

The latter causes the pipeline to fail on no-internet-on-compute-node clusters.

We should give the option to supply a pre-downloaded krona taxonomy/database and supply this e.g. like `--checkm_db` and `--cat_db` etc.

willros:
closes #404 

The user can now point at a local copy of the the taxonomy.tab file needed for running Krona. 

## PR checklist

- [x ] This comment contains a description of changes (with reason).
- [x ] Make sure your code lints (`nf-core lint`).
- [ x] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [x ] Usage Documentation in `docs/usage.md` is updated.
- [ x] Output Documentation in `docs/output.md` is updated.
- [ x] `CHANGELOG.md` is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Added`
 
+- [#X](https://github.com/nf-core/mag/pull/X) - Adds support for pointing at a local db for krona, using the parameter `--krona_db` (by @willros).
 - [#395](https://github.com/nf-core/mag/pull/395) - Adds support for fast domain-level classification of bins using Tiara, to allow bins to be separated into eukaryotic and prokaryotic-specific processes.
 - [#422](https://github.com/nf-core/mag/pull/422) - Adds support for normalization of read depth with BBNorm (added by @erikrikarddaniel and @fabianegli)
 - [#439](https://github.com/nf-core/mag/pull/439) - Adds ability to enter the pipeline at the binning stage by providing a CSV of pre-computed assemblies (by @prototaxites)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Added`
 
-- [#X](https://github.com/nf-core/mag/pull/X) - Adds support for pointing at a local db for krona, using the parameter `--krona_db` (by @willros).
+- [#497](https://github.com/nf-core/mag/pull/#497) - Adds support for pointing at a local db for krona, using the parameter `--krona_db` (by @willros).
 - [#395](https://github.com/nf-core/mag/pull/395) - Adds support for fast domain-level classification of bins using Tiara, to allow bins to be separated into eukaryotic and prokaryotic-specific processes.
 - [#422](https://github.com/nf-core/mag/pull/422) - Adds support for normalization of read depth with BBNorm (added by @erikrikarddaniel and @fabianegli)
 - [#439](https://github.com/nf-core/mag/pull/439) - Adds ability to enter the pipeline at the binning stage by providing a CSV of pre-computed assemblies (by @prototaxites)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Added`
 
-- [#497](https://github.com/nf-core/mag/pull/#497) - Adds support for pointing at a local db for krona, using the parameter `--krona_db` (by @willros).
+- [#497](https://github.com/nf-core/mag/pull/497) - Adds support for pointing at a local db for krona, using the parameter `--krona_db` (by @willros).
 - [#395](https://github.com/nf-core/mag/pull/395) - Adds support for fast domain-level classification of bins using Tiara, to allow bins to be separated into eukaryotic and prokaryotic-specific processes.
 - [#422](https://github.com/nf-core/mag/pull/422) - Adds support for normalization of read depth with BBNorm (added by @erikrikarddaniel and @fabianegli)
 - [#439](https://github.com/nf-core/mag/pull/439) - Adds ability to enter the pipeline at the binning stage by providing a CSV of pre-computed assemblies (by @prototaxites)
--- a/modules/local/krona.nf
+++ b/modules/local/krona.nf
@@ -15,7 +15,7 @@ process KRONA {
     path "versions.yml"             , emit: versions
 
     script:
-    taxonomy_folder = taxonomy_file.toRealPath().getParent()
+    taxonomy_folder = taxonomy_file.getParent()
     """
     ktImportTaxonomy "$report" -tax ${taxonomy_folder}
 
--- a/modules/local/krona.nf
+++ b/modules/local/krona.nf
@@ -8,16 +8,17 @@ process KRONA {
 
     input:
     tuple val(meta), path(report)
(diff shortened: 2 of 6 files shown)


modules/local/krona.nf
d4straub, line 1: Eventually it might be good to go to https://nf-co.re/modules/krona_ktimporttaxonomy (which seems to do the same), but I agree that `--krona_db` parameter is in any case a good addition.

modules/local/krona.nf
```diff
@@ -8,15 +8,16 @@ process KRONA {
...
-    path "*.html"       , emit: html
-    path "versions.yml" , emit: versions
+    tuple val(meta), path("*.html") , emit: html
+    path "versions.yml"             , emit: versions
 
     script:
+    taxonomy_folder = taxonomy_file.toRealPath().getParent()
     """
-    ktImportTaxonomy "$report" -tax taxonomy
+    ktImportTaxonomy "$report" -tax ${taxonomy_folder}
```
jfy133, line 20: 'm a bit concerned by 'toRealPath()': what does this do exactly - does it try to replace a symlink with a 'real' path? This _may_ be problematic  cloud situations (I'm not an expert but has been brought up with me in different cases along similar lines). 

Why not leave it as passing the file directly and running just `.getParent()`? or does that not work with symlinks?

Could you just not supply the directory to the parameter rather than the `.tab` file itself?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Added`
 
+- [#497](https://github.com/nf-core/mag/pull/#497) - Adds support for pointing at a local db for krona, using the parameter `--krona_db` (by @willros).
```
jfy133, line 10:
```suggestion
- [#497](https://github.com/nf-core/mag/pull/497) - Adds support for pointing at a local db for krona, using the parameter `--krona_db` (by @willros).
```

nextflow_schema.json
```diff
@@ -488,6 +488,11 @@
                     "description": "Database for taxonomic binning with kraken2.",
                     "help_text": "The database file must be a compressed tar archive that contains at least the three files `hash.k2d`, `opts.k2d` and `taxo.k2d`. E.g. ftp://ftp.ccb.jhu.edu/pub/data/kraken2_dbs/minikraken_8GB_202003.tgz."
                 },
+                "krona_db": {
+                    "type": "string",
+                    "description": "Database for taxonomic binning with krona",
+                    "help_text": "Local `taxonomy.tab` file for Krona, instead of downloading. Point at the `.tab` file."
```
d4straub, line 494: This doesnt need to be local, a path to a remote source can work as well.
```suggestion
                    "help_text": "Path to `taxonomy.tab` file for Krona, instead of downloading the default file. Point at the `.tab` file."
```

d4straub (commented):
Looks good to me.
Is there a way to include that into tests? Potentially with files from the [test](https://github.com/nf-core/modules/tree/master/tests/modules/nf-core/krona/ktimporttaxonomy) of the [nf-core module ktimporttaxonomy](https://github.com/nf-core/modules/blob/master/modules/nf-core/krona/ktimporttaxonomy/main.nf)?

jfy133 (commented):
Overall LGTM thanks @willros !

Just a question about the implementation with the file rather than a directory (containing the file). If there is an issue with symlinks :+1: I have an alternative suggestion

modules/local/krona.nf
willros: ndeed, I agree. I will eliminate the usage of `toRealPath()`.

The rationale behind allowing users to indicate a file rather than directly selecting a folder is that the output from the Krona DB download process generates a file. Hence, I attempted to maintain consistency in this approach.

modules/local/krona.nf
willros: I will look into how the test is implemented in https://nf-co.re/modules/krona_ktimporttaxonomy.

modules/local/krona.nf
jfy133: OK that's reasonable!

jfy133 (approved):

jfy133:
Hm, auto download didn't work....

```
james@bionb103:~/git/nf-core/mag/tests/work/5e/93d9d90d1c6bd699d78b6c9d1e4f58 (local-krona-db)$ cat .command.sh 
#!/bin/bash -euo pipefail
ktImportTaxonomy "results.krona" -tax null

cat <<-END_VERSIONS > versions.yml
"NFCORE_MAG:MAG:KRONA":
    ktImportTaxonomy: $(ktImportTaxonomy 2>&1 | sed -n '/KronaTools /p' | sed 's/^.*KronaTools //; s/ - ktImportTaxonomy.*//')
END_VERSIONS
```

jfy133 (changes requested):
The KRONA module received `null` rather than the `.getParent()` output for autodownloading...

jfy133:
Ah I think this is because what is literally given to the module is `taxonomy.tab` not the whole path itself, so `getParent()` returns `null` as there is no parent of the string provided.

modules/local/krona.nf
```diff
@@ -8,15 +8,16 @@ process KRONA {
...
-    path "*.html"       , emit: html
-    path "versions.yml" , emit: versions
+    tuple val(meta), path("*.html") , emit: html
+    path "versions.yml"             , emit: versions
 
     script:
+    taxonomy_folder = taxonomy_file.getParent()
     """
-    ktImportTaxonomy "$report" -tax taxonomy
+    ktImportTaxonomy "$report" -tax ${taxonomy_folder}
```
jfy133, line 20:
```suggestion
    ktImportTaxonomy "$report" -tax .
```

If the `.tab` file is provided, it is staged directly in the working directory. We can just indicate `./` as the directory for `-tax`, and gets around all of these problems! (tested and works)!

jfy133:
After thinking about it, my suggestion is to follow what is done in various `bwa` and `Bowtie2` modules, i.e., 

https://github.com/nf-core/modules/blob/868d4c3dc7a3db39d36184173e4fe3484499396e/modules/nf-core/bwa/mem/main.nf#L28

Supply the whole directory containing the taxonomy files, then search for tab, and resolve the path that way.

willros:
Sure, you can solve it that way. The main thing is to make sure it matches the output from the `KRONA_DB` process. Perhaps `getPath()` method could come in handy here?

Looking deeper into the krona module of nf-core, why not use the process implemented there? https://github.com/nf-core/modules/blob/868d4c3dc7a3db39d36184173e4fe3484499396e/modules/nf-core/krona/ktimporttaxonomy/main.nf

jfy133 (approved):
All works for me now, thank you very much @willros this is VERY helpful (as I'm having to do a manual 'full test' of the AWS runs on my offline cluster 😅, which obiovusly wasn't working before...)

jfy133:
> Looking deeper into the krona module of nf-core, why not use the process implemented there? https://github.com/nf-core/modules/blob/868d4c3dc7a3db39d36184173e4fe3484499396e/modules/nf-core/krona/ktimporttaxonomy/main.nf
> 
> Anyways, I implemented the same solution in the local process for this pipeline.

We also need to do a big pass through the pipeline replaceing local modules with now-official moduels... but this will requrie a big overhaul in the future! So thanks again for contribution and we hope you will keep contributing!

PR #508, opened 2023-09-15
2.4 review comments

jfy133:
## PR checklist

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/mag/tree/master/.github/CONTRIBUTING.md)- [ ] If necessary, also make a PR on the nf-core/mag _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [ ] Make sure your code lints (`nf-core lint`).
- [ ] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

PR #509, opened 2023-09-15
[DO NOT MERGE] Registry update for all modules

jfy133:
Updates the URL to remove the registeries in all teh modules (also repairing a couple of patches)

Should only be merged _after_ https://github.com/nf-core/mag/pull/508

TODO

- [x] Fix aDNA SAMTOOLS_FAIDX
- [x] Update or patch GTDBTK here? Or leave that to @maxibor 's https://github.com/nf-core/mag/pull/489

## PR checklist


d4straub (approved):
LGTM

jfy133:
Testing...

jfy133:
Confirm pass!

I note that GTDB-TK in the full test does not end up being executed because now the completeness and contamination values do not reach the default thresholds for any of the bins... but this is a separate issue. I assume this occurred since the change of the GTDB database version/tool update a couple of versions ago.

@d4straub do you see any reason why I should not just set those two values to very low for the `test_full`. At least that way the tool will execute in the full test (even if one wouldn't usually do this).

I guess this was not noticed previously due to the full test failures because of the SPAdes incompatibility with fusion..

d4straub:
> @d4straub do you see any reason why I should not just set those two values to very low for the test_full. At least that way the tool will execute in the full test (even if one wouldn't usually do this).

As far as I remember GTDB failed with too incomplete bins. GTDB uses marker genes to classify bins and when no marker genes are found it clashed. But that might have changed over time and software updates.

PR #543, opened 2023-11-20
Added ability to auto-create samplesheet for nf-core/phageannotator

CarsonJM:
## PR checklist

- [X] This comment contains a description of changes (with reason).
- [X] If you've fixed a bug or added code that should be tested, add tests!
- [X] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/mag/tree/master/.github/CONTRIBUTING.md)
- [X] Make sure your code lints (`nf-core lint`).
- [X] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).


nextflow_schema.json
```diff
@@ -52,6 +52,10 @@
                     "type": "string",
                     "description": "MultiQC report title. Printed as page header, used for filename if not otherwise specified.",
                     "fa_icon": "fas fa-file-signature"
+                },
+                "nf_core_pipeline": {
+                    "type": "string",
+                    "description": "Create a samplesheet for the specified nf-core pipeline"
```
jfy133, line 58: Missing longer helptext describing a bit more what it does; also missing the `enum` to ensure the dropdown/validation is correct.

jfy133 (commented):
Ran out of time before baby came back so couldn't test, but first pass: very clean code thank you very much @CarsonJM !

Couple of things  missing (docs atm):

- Missing changelog update
- Missing update to README
- Missing input/output docs

nextflow.config
```diff
@@ -13,6 +13,7 @@ params {
     input                                = null
     single_end                           = false
     assembly_input                       = null
+    nf_core_pipeline                     = null
```
jfy133, line 16: I'm not sure about this parameter name... I prefer things to be more active:

```
--generate_downstream_samplesheet
```

What about something like that?

subworkflows/local/samplesheet_creation.nf
```diff
@@ -0,0 +1,115 @@
...
+                    def meta_new = meta.subMap('id')
+                [ meta_new, fasta ]
+            }
+            .groupTuple()
+
+        //
+        // MODULE: Combine all assemblies from a sample into one FastA file
+        //
+        ch_combined_assemblies = CAT_CAT ( ch_assemblies ).file_out
+        ch_versions = ch_versions.mix( CAT_CAT.out.versions )
```
jfy133, line 25: I guess this maybe would not apply to all pipelines? I wonder if you should move much of this into a pipeline specific creation subworkflow, so this file would have 

```
if { nf_core_pipeline == phage_annotator } {
   CREEATE_ PHAGEANNOTATOR_SAMPLESHEET (xyz)
} (nf_core_pipeline == funcscan {
  CRERATE_FUNCSCAN_SAMPLESHEET
}
```

This would then make it more flexible for other pipelines

subworkflows/local/samplesheet_creation.nf
```diff
@@ -0,0 +1,115 @@
...
+                }.join ( ch_combined_assemblies_remap )
+                .map {
+                    id, group, single_end, fastq, fasta ->
+                        def reads   = fastq instanceof List ? fastq.flatten() : [ fastq ]
+                        def meta    = [:]
+
+                        meta.id         = id
+                        meta.group      = group
+                        meta.single_end = single_end
+                        meta.fastq_1    = reads[0] ? reads[0] : ''
```
jfy133, line 53: Shouldn't this _always_ be just `reads[0]` in this case? would there be a case the reads are not passed here/phageannotator doesn't need the reads?

subworkflows/local/samplesheet_creation.nf
```diff
@@ -0,0 +1,115 @@
...
+                        meta.fastq_2    = reads[1] && !meta.single_end ? reads[1] : ''
```
jfy133, line 54:
```suggestion
                        meta.fastq_2    = reads[1] && !meta.single_end ? reads[1] : ''
```

Sort of the sam as above, if should simply be `!meta.single_end ? reads[1] : ''`

subworkflows/local/samplesheet_creation.nf
```diff
@@ -0,0 +1,115 @@
...
+
+                        return meta
+                }
+                .set { ch_mag_metadata }
+        } else {
+            // if coassembly was used, join FastQ and FastA by group
+            ch_combined_assemblies_remap = ch_combined_assemblies
+                .map {
+                    meta, fasta ->
+                        def group = meta.id.split('-')
```
jfy133, line 65: Will this alwasy work? what if the sample name has `-` in it?

(Or the group name... I can't remember which order it goes)

conf/modules.config
```diff
@@ -741,6 +741,22 @@ process {
...
+        publishDir = [
+            path: { "${params.outdir}/samplesheet" },
+            enabled: false
+        ]
+    }
+
+    withName: 'MAG_MERGE_SAMPLESHEET' {
+        ext.prefix = "${params.nf_core_pipeline}"
+        publishDir = [
+            path: { "${params.outdir}/samplesheet" },
```
jfy133, line 754: `samplesheet` seems a bit generic, we spit out a 'validated' samplesheet in `piepline_info`. What about `downstream_samplesheet` or something... I know that's a bit ugly but I'm open to suggestions

modules/local/mag_merge_samplesheet.nf
```diff
@@ -0,0 +1,28 @@
...
+    path ('samplesheets/*')
+
+    output:
+    path "*_samplesheet.csv", emit: samplesheet
+    path "versions.yml"   , emit: versions
+
+    script:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    head -n 1 `ls ./samplesheets/* | head -n 1` > ${prefix}_samplesheet.csv
```
jfy133, line 18: This worries me a little; because hacky awk.

Why not using the native `.collectFile()` operator: https://www.nextflow.io/docs/latest/operator.html#collectfile 

 would that be able to serve a similar purpose? Same applies to `mag_to_samplesheet` now I think about it... `collectFile` might serve both purposes in one, as long as the channel strucure is already correct.

jfy133 (commented):
Slightly more lucid feedback, sorry didn't think of these before

CarsonJM:
@nf-core-bot fix linting

subworkflows/local/create_phageannotator_samplesheet.nf
```diff
@@ -0,0 +1,104 @@
...
+                        meta.fastq_2    = !meta.single_end ? reads[1] : ''
+                        meta.fasta      = fasta ? fasta : ''
+
+                        return meta
+                }
+                .set { ch_mag_metadata }
+        }
+
+        // Create samplesheet for each sample using meta information
+        ch_mag_id_samplesheets = ch_mag_metadata.collectFile() { meta ->
```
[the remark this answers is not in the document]
jfy133: One thing I did not consider with this: the paths of the FASTA/FASTQs point to the working directory, not the results directory! Probably something to cleanup (`${params.outdir}/Assembly/${meta.assembler}/` + fasta.name )` or something?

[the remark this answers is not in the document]
jfy133: (Noted when testing)

```
sample,group,fastq_1,fastq_2,fasta
test_minigut,0,/workspace/mag/testing/work/3c/41f57f9f4eb72e0fe3e2bfb22c17d9/test_minigut_1.merged.fastq.gz,/workspace/mag/testing/work/3c/41f57f9f4eb72e0fe3e2bfb22c17d9/test_minigut_2.merged.fastq.gz,/workspace/mag/testing/work/73/4738167448abe5f370a53e638514c6/test_minigut.fa
test_minigut_sample2,0,/nf-core/test-datasets/raw/mag/test_data/test_minigut_sample2_R1.fastq.gz,/nf-core/test-datasets/raw/mag/test_data/test_minigut_sample2_R2.fastq.gz,/workspace/mag/testing/work/fc/187d2e7655cfdc6202a903bdca8904/test_minigut_sample2.fa
```

jfy133 (commented):
Non-blocking thoughts:

- Q: I wonder if we should put this into an umbrella workflow rather than the if/else statement
- Q: The `test` profile should ideally be `test_generatesamplesheet` rather than just `test_samplesheet` (someone might think it's to use an input samplesheet)
- Q: Are you definitely sure people would always want to collapse all assemblies per sample (no cases they would want to keep it separate)

subworkflows/local/create_phageannotator_samplesheet.nf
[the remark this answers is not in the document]
CarsonJM: So you think it would be better to have them point to the results directory? Definitely can change that if you think it would be best

CarsonJM:
Regarding your questions: 
1) I think it would be a good idea to create an umbrella workflow for generating samplesheets for any downstream nf-core pipeline 
2) Great suggestion, I will make that change
3) It is possible that they would want to keep them separate, so I can make that an option!

jfy133 (commented):
OK overall this looks good to me now - the only thing I don't like is the paths to `work` directories for the files - I strongly feel that it should be the relevant subdirectory of `params.outdir` + filename.

docs/output.md
```diff
@@ -706,6 +707,20 @@ Because of aDNA damage, _de novo_ assemblers sometimes struggle to call a correc
...
+<summary>Output files</summary>
+
+- `samplesheet/`
+  - `[generate_downstream_samplesheet].csv`: a samplesheet in CSV format that can be directly used as input for the specified nf-core pipeline
+
+</details>
+
+Currently, samplesheets for the following nf-core pipelines can be automatically generated:
+
+- [phageannotator](https://github.com/nf-core/phageannotator): a pipeline for identifying, annotation, and quantifying phage sequences in (meta)-genomic sequences.
```
CarsonJM, line 722:
```suggestion
- [phageannotator](https://github.com/nf-core/phageannotator): a pipeline for identifying, annotation, and quantifying phage sequences in (meta)-genomic sequences. Utilizes quality controlled reads and contigs generate by nf-core/mag
```

nextflow_schema.json
```diff
@@ -884,6 +884,24 @@
...
+            "properties": {
+                "generate_downstream_samplesheet": {
+                    "type": "string",
+                    "description": "Create a samplesheet for the specified nf-core pipeline",
+                    "help_text": "Automatically generate a samplesheet in CSV format that can be directly used as input for the specified nf-core pipeline.",
+                    "enum": ["phageannotator"]
+                },
+                "samplesheet_combine_assemblers": {
+                    "type": "boolean",
+                    "description": "Combine assembler outputs into one assembly file"
```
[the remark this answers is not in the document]
CarsonJM:
```suggestion
                    "description": "Combine all contigs from all assemblies of a given sample into a single FASTA file"
```

subworkflows/local/generate_downstream_samplesheet.nf
```diff
@@ -0,0 +1,117 @@
...
+                    .set { ch_mag_metadata }
+            }
+
+            // Create samplesheet for each sample using meta information
+            ch_mag_id_samplesheets = ch_mag_metadata.collectFile() { meta ->
+                [ "${meta.id}_phageannotator_samplesheet.csv", "sample,group,fastq_1,fastq_2,fasta" + '\n' + "${meta.id},${meta.group},${meta.fastq_1},${meta.fastq_2},${meta.fasta}" + '\n' ]
+            }
+
+            // Merge samplesheet across all samples for the pipeline
+            ch_mag_id_samplesheets.collectFile(name: "phageannotator_samplesheet.csv", keepHeader:true, skip:1, storeDir:"${params.outdir}/downstream_samplesheets/")
```
CarsonJM, line 112: @jfy133 Just to clarify, are you thinking I should add the `storeDir:"${params.outdir}/downstream_samplsheets/"` to the first `collectFile()`?

jfy133: No, more don't bind just `fasta` to the string on the line above.

`params.outdir + <path I don't remember> + fasta.name` or something

jfy133: Or even just validate that if you turn on samplesheet creation, then you must also save the fastqs.

But yes, one way or another - running from explicitly saved fsstqs in outdir is preferable.

If you have `cleanup = true` in a config, something many do but the user never sees, they won't have any fastq files in the work dir to point to in the sample sheet 😬. This is harder to control for rather than directing the user to make sure to save the fsstqs

subworkflows/local/generate_downstream_samplesheet.nf
```diff
@@ -103,10 +104,35 @@ workflow GENERATE_DOWNSTREAM_SAMPLESHEET {
...
 
-            // Create samplesheet for each sample using meta information
-            ch_mag_id_samplesheets = ch_mag_metadata.collectFile() { meta ->
-                [ "${meta.id}_phageannotator_samplesheet.csv", "sample,group,fastq_1,fastq_2,fasta" + '\n' + "${meta.id},${meta.group},${meta.fastq_1},${meta.fastq_2},${meta.fasta}" + '\n' ]
-            }
+                // Create samplesheet for each sample using meta information
+                ch_mag_id_samplesheets = ch_mag_metadata.collectFile() { meta ->
+                    // Save reads and assemblies to outdir so that they are in a stable location
+                    file(meta.fastq_1.toUriString(), checkIfExists: true).copyTo("${params.outdir}/downstream_samplesheets/fastq/${meta.fastq_1.name}")
+                    file(meta.fasta, checkIfExists: true).copyTo("${params.outdir}/downstream_samplesheets/fasta/${meta.fasta.name}")
```
jfy133, line 111: Asking for some advise about this, this is indeed very clean, but given it's outside a process/publishdir I'm a bit unsure. E.g. some people use symlinks from work to results dir rather than copy, so this would violate this.

Secondly this woudl potentially result in two copies of the same read files if one of the `--save_*` parameters are given.

It maybe we have to come up with some complicated logic that instead picks the right 'final' directory for the proceessing reads, and it's that that you append to the beginning, and then add.

I think this would come with two steps:

1. Make sure one of `--save_clipped_reads`, `--save_hostremoved_reads`, `--save_phixremoved_reads`, `--save_bbnorm_reads` are selected if `params.generate_samplesheet` is true

    So some input validation code like:
    ```
    if ( params.generate_samplesheet && ![params.saved_clipped, save.hostremoved <...>].any() ) { Nextflow.error('[nf-core/mag] ERROR: must at least save one XYX if --generate_samplesheet <...>') }
     ```
     
 2.  Then have to have some complicated logic to select which directory gets appended to the beginning of the file name (probably best in a case when but w/e), e.g.
 
     ```
     if (params.save_clipped_reads && !params.save_phix && !params.save_host_removed && !params._save_bbnorm) { 
    samplehseet_reads_dir = /subdir/clipedreads/saved/in
       } else if ( params.save_phix && !params.save_host_removed && !params._save_bbnorm){
       samplesheet_reads_dir = /subdir/phixreads/saved/in
       }
     ```

CarsonJM: Totally see what you're saying and was curious what thoughts would be about doing this outside a process. I am more than happy change it to what you're suggesting!

The reason I was giving this approach a try was I was hoping to find a generalizable approach that would work across pipelines/updates even if the save input logic you're talking about gets modified (also, I selfishly wanted to find a quicker approach haha) Also, I was hoping to force a copy because in the past I've symlinked to publishDir thinking I copied and had my reads disappear because my workDir is in a scratch directory 😅.

Maybe we could make a module that just collects/copies the files, then at least were getting inputs and outputs checked for existence and it would take care of issue #1 with this approach.

Still, I definitely have though about what you're saying and will make that change if you think it's best!

PR #620, opened 2024-05-23
Added test_nothing nf-test and updated CI

CarsonJM:
## PR checklist


prototaxites:
Not knowing much (yet) about nf-test - I think this looks great! 

I had a poke around in the failing tests - here, the kreport test fails because the test reads the mag `conf/modules.conf` and sets the prefix as `ext.prefix = { "${meta.id}_kreport" }`, whereas the test snapshot expects the output file to just be called `${meta.id}.txt`. I suspect this could be a problem for any test where we're manually setting a prefix and adding custom strings to it which aren't hard-coded in the process itself. Off the top of my head this may cause problems with the read QC filtering, phiX removal, etc. as these add prefixes like "_host_removed" to separate out fastq files which otherwise would have the same names.

jfy133:
Replaced with https://github.com/nf-core/mag/pull/827

PR #652, opened 2024-08-20
Update nextflow_schema.json

amizeranschi:
Remove "=" sign from parameter value assignment

Previous instructions may lead to issues when the parameter initialized with an "=" sign is precedeted by a Boolean parameter. See the discussion here:
https://nfcore.slack.com/archives/CE9MS66BS/p1724141829294639?thread_ts=1724053876.586899&cid=CE9MS66BS


prototaxites (approved):
LGTM 👍

Issue #663, opened 2024-09-05
convert_depths for co-assembly taking only average depth and not sample wise read depths for maxbin2 binning with solution


uel3:
### Description of the bug

I ran into the issue of not generating high quality bins for a known bacteria present in my samples (0.25 by DASTool scoring) when using nf-mag for generating MAGs from a metagenomic co-assembly. This resulted in the known bacteria bins not being included in the final refined DASTool bins using default parameters. I was able to generate a high quality (0.95 by DASTool scoring) bin of the known bacteria present in my samples when I ran MaxBin2, MetaBat2, and DASTool in a separate mNGS pipeline using the same parameters but instead passing reads_list into MaxBin2 instead of the abund_file like in nf-mag. When I looked at the input and output files for the nf-mag processes, I noticed far less depth information being used to generate bins with MaxBin2, only the total depth from METABAT2_JGISUMMARIZE output was being passed as the -abund_file for MaxBin2 instead of sample wise or individual read depth for each contig.   
 
I believe the issue lies here in line 21 of CONVERT_DEPTHS process used in the BINNING subworkflow: 
```console
bioawk -t '{ { if (NR > 1) { { print \$1, \$3 } } } }' ${depth.toString() - '.gz'} > ${prefix}_mb2_depth.txt
```
I figured out how to change the process to provide sample wise depths and generate my missing high quality bin. Instead of passing the abund_file that comes from CONVERT_DEPTHS output, the mNGS reads can be directly passed via the -reads or -reads_list flag as I did in my separate mNGS pipeline. Using this approach nf-mag generates the high quality bins for my known pathogen but requires more time and resources to do so. My fix is to use still use the depth information generated by METABAT2_JGISUMMARIZE but use the sample-wise depth information for all contigs and pass as -abund_list which is the solution I offer below. 

### Command used and terminal output

```console
$ nextflow run nf-core/mag --coassemble_group --binning_map_mode 'group' --refine_bins_dastool --postbinning_input 'refined_bins_only'
```

### Relevant files

Command.sh from CONVERT_DEPTHS with my data:
```console
#!/bin/bash -euo pipefail
gunzip -f MEGAHIT-group-Col-depth.txt.gz
bioawk -t '{ { if (NR > 1) { { print $1, $3 } } } }' MEGAHIT-group-Col-depth.txt > group-Col_mb2_depth.txt
 
cat <<-END_VERSIONS > versions.yml
[...]

PR #690, opened 2024-10-11
Fix MaxBin2 abundance input

d4straub:
Fix https://github.com/nf-core/mag/issues/663

I adopted for `CONVERT_DEPTHS` largely the solution by @uel3 but changes the `MaxBin2` module so that it accepts depth files via `-abund`, I think this is a more elegant solution. Additionally the `CONVERT_DEPTHS` module was changed so that the abundance files are named by samples, that makes the `*.abundance` files by MaxBin2 include the sample names, i.e. they are meaningful now.
The nf-core module maxbin2 was updated separately and is used here.
Linting with nf-core tools version 2.14.1 succeeds.

## PR checklist


jfy133:
I would rather these to go in the official module @d4straub but I'm around for fast review for that too

modules/nf-core/maxbin2/main.nf
```diff
@@ -27,7 +28,15 @@ process MAXBIN2 {
...
-    def associate_files = reads ? "-reads $reads" : "-abund $abund"
+    def associate_files = ""
+    if ( reads ) {
+        associate_files = "-reads $reads"
+    } else if ( abund instanceof List ) {
+        associate_files = "-abund ${abund[0]}"
+        for (i in 2..abund.size()) { associate_files += " -abund$i ${abund[i-1]}" }
+    } else {
+        associate_files = "-abund $abund"
+    }
```
jfy133, line 39: Maybe also add an error check that `reads` and `abund` cannot be supplied together

> (Warning: Please do NOT specify abundance and reads that BELONG TO THE SAME METAGENOME together. MaxBin will treat them as two different information and thus will count this metagenome TWICE!)

From the docs

jfy133 (commented):
Looking good so far, just I would request porting this to the official module (like I said above, I'm available for rapid review on that)

My last question is (being naive but want to make sure this is recorded for future me):

I was just thinking @uel3 mentioned this issue in the context of co-assembly,  do we need to distinguish the mechanism being added here  between standalone or co-assembly?

Reading the minimal Maxbin2 docs I think it doesn't really matter - so we can always do it in this way, but I just want to double check as you likely have more experience than I have with the tool.

modules/nf-core/maxbin2/main.nf
d4straub: But only ever reads or abund is used in the module, never both, so that isnt neccessary, if I am not mistaken?

d4straub:
> Looking good so far, just I would request porting this to the official module (like I said above, I'm available for rapid review on that)

Thasnks, PR to modules is open.

> I was just thinking @uel3 mentioned this issue in the context of co-assembly, do we need to distinguish the mechanism being added here between standalone or co-assembly?

Not sure I understand the question. In my tests this works with and without co-assembly, it depends rather on `--binning_map_mode`, and it works in my tests as expected.

jfy133:
> > Looking good so far, just I would request porting this to the official module (like I said above, I'm available for rapid review on that)
> 
> Thasnks, PR to modules is open.

OK will look now!

> > I was just thinking @uel3 mentioned this issue in the context of co-assembly, do we need to distinguish the mechanism being added here between standalone or co-assembly?
> 
> Not sure I understand the question. In my tests this works with and without co-assembly, it depends rather on `--binning_map_mode`, and it works in my tests as expected.

That answers my question basically :+1:

jfy133:
@d4straub module can be installed in this PR now :)

jfy133:
Just going to run a test with current version and your changes to compare, but code looks good :+1:

jfy133:
@d4straub I had a look and it seems to run through, however when I run with `test_virus_identification` which does co-assembly, the `depths/bin_summary.tsv` only reports CONCOCT... do you see that in your test run also?

d4straub:
@jfy133 I checked the code and `CONVERT_DEPTHS` is only channeled into `ch_maxbin2_input` which only goes into `MAXBIN2`. As far as I see, of the many files that are available from `MAXBIN2.out`, summary file isnt used, and that file is the only file that changes in this PR (MaxBin2's summary file looses column that has abundance with the fix).

Works as expected for me (this runs on this modified branch):
```
nextflow run mag -profile test_bbnorm,singularity --skip_binning false --centrifuge_db false --kraken2_db false --outdir results_test_bbnorm_3-1-0dev -resume --skip_concoct
nextflow run mag -profile test_bbnorm,singularity --skip_binning false --centrifuge_db false --kraken2_db false --outdir results_test_bbnorm_3-1-0dev_concoct -resume
```
`test_virus_identification` uses
```
    skip_binning                = true
    skip_metabat2               = true
    skip_maxbin2                = true 
```
so I would not expect any info from binners...
(my test with `test_virus_identification` continuously fails because singularity pull doesnt work reliably since a few days for me...)

jfy133:
OK that's strange, maybe I just had a weird test configuration (using the `test_virus_identification` profile, and extra flags), but it looks like you've tested in multiple configurations and it always looks OK :+1:

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#674](https://github.com/nf-core/mag/pull/674) - Make longread preprocessing a subworkflow (added by @muabnezor)
 - [#674](https://github.com/nf-core/mag/pull/674) - Add porechop and filtlong logs to multiqc (added by @muabnezor)
 - [#674](https://github.com/nf-core/mag/pull/674) - Change local filtlong module to the official nf-core/filtlong module (added by @muabnezor)
+- [#690](https://github.com/nf-core/mag/pull/690) - MaxBin2 is using the abundance information from different samples now as expected (reported by @uel3 and fixed by @d4straub)
```
jfy133, line 21:
```suggestion
- [#690](https://github.com/nf-core/mag/pull/690) - MaxBin2 now using the abundance information from different samples rather than an average (reported by @uel3 and fixed by @d4straub)
```

jfy133 (approved):

jfy133:
@d4straub I will merge in and then do the template merge

PR #674, opened 2024-09-20
Move long read preprocessing into a subworkflow

muabnezor:
This PR moves long read preprocessing into a subworkflow. It also adds the option to run porechop-abi, instead of porechop, for long read adaptertrimming. 

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/mag/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/mag _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core lint`).
- [x] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [x] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [x] `CHANGELOG.md` is updated.
- [x] `README.md` is updated (including new tool citations and authors/contributors).

PR #707, opened 2024-10-27
Bin QC Improvements

### LINKED ISSUE #607: Replace CheckM with CheckM2

dialvarezs:
This PR adds:
- CheckM2 as an alternative for  bin qc
- Updates CheckM and GUNC modules
- A new `BIN_QC` subworkflow, integrating  CheckM, CheckM2, BUSCO and GUNC

Closes #607.

## PR checklist

- [x] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/mag/tree/master/.github/CONTRIBUTING.md)
- [x] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).


jfy133:
@nf-core-bot fix linting

jfy133:
Before you continue (sorry this is a bit late):

I generally don't like to deprecate old version of tools for a while, but rather keep them as alternative tools.

In some cases people want to stick with the original version for compatibility with previous runs

Could you 'revert' (or reinstall) the old checkm module and wrap it in an if/else statement (but within the subworkflow :) ) 

@muabnezor did a similar thing when adding porechop_ABI hree: https://github.com/nf-core/mag/pull/674

dialvarezs:
@jfy133 that makes sense, I will revert the CheckM removal. Bad for me for not asking before 😅.

dialvarezs:
Sorry, I didn' catch your last comment about merging both tools in a single workflow. In that sense, does it make sense to include BUSCO as well, and just make a "bin_qc" subworkflow?

jfy133:
> Sorry, I didn't catch your last comment about including both tools in a single workflow. With that in mind, would make sense to include BUSCO as well, and just make a "bin_qc" subworkflow?

Yes that would be perfect! We need to subworkflow the sh*t out of this monster :sweat_smile:  thank you!!!

dialvarezs:
It's ready for review now, @jfy133. I’ve added GUNC to the subworkflow to keep mag.nf cleaner.

The only issue I found is with GUNC_MERGECHECKM outputs not being saved correctly. Since the process runs per bin but uses the sample as ID, the output files are overwritten. However, once [this PR](https://github.com/nf-core/modules/pull/6920) is merged, we’ll be able to run GUNC with all bins simultaneously, which should resolve the issue.

jfy133:
Hi @dialvarezs just FYI I'm at another conference since yesterday and didn't have a chance beginning of this week to look at your PRs, sorry about that! 

My response time should speed up from next week :)

prototaxites:
I think this looks good, though I don't have the latitude to run any manual tests of it myself at the moment - so probably wait for someone else to sign off on it!

My only question is whether it might be good to move all the binqc database preparation steps (e.g. CHECKM2_DATABASEDOWNLOAD, and the bits that initialise the files/channels) inside the BINQC workflow where they are consumed? I think keeping conceptually-related code together will probably help with maintenance down the line, and the main mag.nf workflow is already over-full. Plus, there's less subworkflow inputs and outputs to keep track of. Might be a job for another PR though!

dialvarezs:
@prototaxites Absolutely, that makes sense to me. That would help to simplify the mag workflow a bit. What do you think about this @jfy133 ?

jfy133:
I've pondered that a few times, however many of these download steps take a very long time, and thus from a user PoV I think it makes sense to have it triggered right at the beginning of the pipeline so by the time assembly and binning is done it's ready to go rather than getting all the way to binning and then having to wait the same length of time again before you can start the binning QC.

In my mind to clear up the code I would rather have a dB download subworkflow for all DB Downloadsto make the code clearer. Maybe from a related module PoV it's not as efficient but functionally they are related

dialvarezs:
In most cases, database downloads don’t depend on anything, so, if I'm not wrong, the processes should start immediately regardless of where they are in placed in the code. Or am I missing something?

prototaxites:
> In most cases, database downloads don’t depend on anything, so, if I'm not wrong, the processes should start immediately regardless of where they are in placed in the code. Or am I missing something?

Yes, that's my understanding - Nextflow processes kick off as soon as any inputs are available, so processes beginning with file/URL input from parameters should start as soon as the pipeline begins, no matter how many subworkflows deep they are (there might be a latency hit if you go 10,000 subworkflows deep...). 

Also not opposed to a "database download" subWF - but that seems a little more limited in scope. In particular I'm thinking about all the steps that are like `"if(params.some_db) { db = Channel.value(file(path)) } else { db = Channel.empty }"` - there are a lot of these input databases, often just a zip or fasta file, and building these in a single WF would make something with a lot of outputs. Hence why I think it may be more maintainable to keep that code near to where the outputs are consumed (set the PhiX fasta in the "short read preprocessing" subWF, lambda in the "long read preprocessing", etc.).

jfy133:
> > In most cases, database downloads don’t depend on anything, so, if I'm not wrong, the processes should start immediately regardless of where they are in placed in the code. Or am I missing something?
> 
> Yes, that's my understanding - Nextflow processes kick off as soon as any inputs are available, so processes beginning with file/URL input from parameters should start as soon as the pipeline begins, no matter how many subworkflows deep they are (there might be a latency hit if you go 10,000 subworkflows deep...).

Hmm fair. I might be feeling over defensive due to the huge number of conditions mag has... so maybe this is indeed the case.

> Also not opposed to a "database download" subWF - but that seems a little more limited in scope. In particular I'm thinking about all the steps that are like `"if(params.some_db) { db = Channel.value(file(path)) } else { db = Channel.empty }"` - there are a lot of these input databases, often just a zip or fasta file, and building these in a single WF would make something with a lot of outputs. Hence why I think it may be more maintainable to keep that code near to where the outputs are consumed (set the PhiX fasta in the "short read preprocessing" subWF, lambda in the "long read preprocessing", etc.). But either way cleaning up the main WF should help - this all just may be beyond the scope of this PR.

Fair point.

Ok - if you're feeling to up to it @dialvarezs go ahead and move the relevant database downloads  to the BINQC subworklow(s) :)

Note you don't need a separate CheckM download CI check -> I would rather you just include an UNTAR step in the pipeline isntead.

I added the seaprte CheckM CI tests originally due to instable downloads when connecting to the servers in Australia, but as these databases are now on Zenodo this should rarely happen.

dialvarezs:
@jfy133 I'll get back to this on Friday or Saturday.

> Ok - if you're feeling to up to it @dialvarezs go ahead and move the relevant database downloads to the BINQC subworklow(s) :)

I think this is a step in the right direction to improve the modularity of mag, so I will.

> Note you don't need a separate CheckM download CI check -> I would rather you just include an UNTAR step in the pipeline isntead.

Got it, I will remove the checkm2 ci test I added. Should I remove the checkm ci test as well?

jfy133:
> @jfy133 I'll get back to this on Friday or Saturday.
> 
> > Ok - if you're feeling to up to it @dialvarezs go ahead and move the relevant database downloads to the BINQC subworklow(s) :)
> 
> I think this is a step in the right direction to improve the modularity of mag, so I will.
> 
> > Note you don't need a separate CheckM download CI check -> I would rather you just include an UNTAR step in the pipeline isntead.
> 
> Got it, I will remove the checkm2 ci test I added. Should I remove the checkm ci test as well?

Yes you probably can! Thank you!

conf/modules.config
```diff
@@ -364,9 +368,30 @@ process {
...
+            mode: params.publish_dir_mode,
+            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
+        ]
+    }
+
+    withName: CHECKM2_DATABASEDOWNLOAD {
+        publishDir = [
+            path: { "${params.outdir}/GenomeBinning/QC/CheckM2/checkm2_downloads" },
+            mode: params.publish_dir_mode, overwrite: false,
+            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }, enabled: params.save_checkm2_data
```
jfy133, line 384:
```suggestion
            saveAs: { filename -> filename.equals('versions.yml') ? null : filename },
            enabled: params.save_checkm2_data
```

docs/output.md
```diff
@@ -566,6 +567,31 @@ If the parameter `--save_checkm_reference` is set, additionally the used the Che
 
 </details>
 
+#### CheckM2
+
+[CheckM2](https://github.com/chklovski/CheckM2) is atool for assessing the quality of metagenome-derived genomes. It uses a machine learning approach to predict the completeness and contamination of a genome regardless of its taxonomic lineage.
```
jfy133, line 572:
```suggestion
[CheckM2](https://github.com/chklovski/CheckM2) is a tool for assessing the quality of metagenome-derived genomes. It uses a machine learning approach to predict the completeness and contamination of a genome regardless of its taxonomic lineage.
```

docs/output.md
```diff
@@ -566,6 +567,31 @@ If the parameter `--save_checkm_reference` is set, additionally the used the Che
...
+
+<details markdown="1">
+<summary>Output files</summary>
+
+- `GenomeBinning/QC/CheckM2/`
+  - `[assembler]-[binner]-[domain]-[refinement]-[sample/group]/quality_report.tsv`: Detailed statistics about bins informing completeness and contamamination scores. This should normally be your main file to use to evaluate your results.
+  - `[assembler]-[binner]-[domain]-[refinement]-[sample/group]/`: Intermediate files for CheckM2 results, including CheckM2 generated annotations, log, and Diamond alignment results.
```
jfy133, line 579:
```suggestion
  - `[assembler]-[binner]-[domain]-[refinement]-[sample/group]/`: Intermediate files for CheckM2 results, including CheckM2 generated annotations, log, and DIAMOND alignment results.
```

modules/local/bin_summary.nf
```diff
@@ -18,18 +17,17 @@ process BIN_SUMMARY {
...
     def cat_summary    = cat_sum.sort().size() > 0 ?    "--cat_summary ${cat_sum}" : ""
     """
     combine_tables.py --depths_summary ${bin_depths} \
-                    $busco_summary \
-                    $checkm_summary \
+                    $binqc_summary \
                     $quast_summary \
                     $gtdbtk_summary \
                     $cat_summary \
+                    --binqc_tool ${params.binqc_tool} \
```
jfy133, line 30: To try and conform to nf-core module standards as much as possible, can you instead pass this via an input channel as it's presumably a mandatory option?

subworkflows/local/bin_qc.nf
```diff
@@ -0,0 +1,182 @@
...
+                .map { _lin, downloads -> downloads[0] }
+                .toSortedList()
+                .flatten()
+            BUSCO_SAVE_DOWNLOAD(ch_downloads)
+        }
+
+        BUSCO(ch_input_bins_for_qc, ch_db_for_busco)
+
+        BUSCO_SUMMARY(
+            BUSCO.out.summary_domain.collect { v -> v[1] }.ifEmpty([]),
```
jfy133, line 76: Please replace `f` with more human descriptive variables :)

subworkflows/local/bin_qc.nf
```diff
@@ -0,0 +1,182 @@
...
+                .groupTuple()
+                .map { _lin, downloads -> downloads[0] }
+                .toSortedList()
+                .flatten()
+            BUSCO_SAVE_DOWNLOAD(ch_downloads)
+        }
+
+        BUSCO(ch_input_bins_for_qc, ch_db_for_busco)
+
+        BUSCO_SUMMARY(
```
jfy133, line 75: Missing versions file being mixed into `ch_versions` for this process, the SAVE_DOWNLOAD doesn't have one, although given that process is entirely empty, shouldn't that be replaced by a `publishDir` specification in `modules.conf` :thinking: 

(I know you didn't write this, but might be worth fixing this now :) )

subworkflows/local/bin_qc.nf
```diff
@@ -0,0 +1,182 @@
...
+
+        ch_checkmqa_input = CHECKM_LINEAGEWF.out.checkm_output
+            .join(CHECKM_LINEAGEWF.out.marker_file)
+            .map { meta, dir, marker ->
+                [meta, dir, marker, []]
+            }
+
+        CHECKM_QA(ch_checkmqa_input, [])
+
+        COMBINE_BINQC_TSV(CHECKM_QA.out.output.collect { v -> v[1] })
```
jfy133, line 112: same again, what type of file is the 'v' referring to? I like to be ultra verbose as we have quite diverse developers contributing to the same pipeline :)

subworkflows/local/bin_qc.nf
```diff
@@ -0,0 +1,182 @@
...
+         */
+        CHECKM2_PREDICT(ch_input_bins_for_qc.groupTuple(), ch_checkm2_db)
+
+        COMBINE_BINQC_TSV(CHECKM2_PREDICT.out.checkm2_tsv.collect { v -> v[1] })
+
+        qc_summary = COMBINE_BINQC_TSV.out.combined
+        ch_versions = ch_versions.mix(CHECKM2_PREDICT.out.versions.first())
+    }
+
+    if (params.run_gunc) {
```
jfy133, line 129: Can you check whether GUNC is supported in whatever the `qc_summary` output channel goes into?

jfy133 (commented):
This is looking good to me! 

I just completed a comparison test between checkm/checkm2 and just need to compare the results.

Then I'll run with BUSCO/GUNC next week (have to stay home with the kiddo tomorow as well unfortunately) to make sure they aren't broken and then I think we are good to go.

Do you want to include the moving of the database downloading in this PR @dialvarezs or in a follow up one?

jfy133:
No worries!

No stress if you can't make it ATM, but I think we are close to a release once this is in!

dialvarezs:
If you have time, can you review this PR? https://github.com/nf-core/modules/pull/6920
I also plan to update the GUNC/CHEKM modules in this PR, and also the GUNC PR should solve the problem with gunc outputs.

jfy133:
> If you have time, can you review this PR? [nf-core/modules#6920](https://github.com/nf-core/modules/pull/6920) I also plan to update the GUNC/CHEKM modules in this PR, and also the GUNC PR should solve the problem with gunc outputs.

Done!

jfy133:
My kid is sick again :roll_eyes: but if you can update GUNC in this PR before then, then I will be working on mag Thursday so can finish reviewing this PR again :) (but no stress!)

subworkflows/local/bin_qc.nf
dialvarezs: Right now GUNC summary it's not being used anywhere down in the pipeline. Maybe we can integrate the results in the  bin summary, but that should be for another PR.

subworkflows/local/bin_qc.nf
dialvarezs: By the comment on that module, the problem with simply uisng `publishDIr` is that the same files would get written multiple times. I couldn't think of a more elegant solution, so I will add the version, unless you have a better idea.

subworkflows/local/bin_qc.nf
dialvarezs: TBH, I was using `v` because the Nextflow documentation no longer encourages the use of `it`, and instead uses `v` in the examples:: https://www.nextflow.io/docs/stable/script.html#closures. But yeah, you're absolutely right, it's a very lazy thing to do.

dialvarezs:
I made all the important updates. I will run a complete test with every QC tool in out HPC to check if everything is alright, and with that done it should be ready.

dialvarezs:
It should be ready now. There is a last minor issue that should be solved by this PR: https://github.com/nf-core/modules/pull/7119

jfy133:
Running the failed test one more time, the conoct test has failed a few times before too right?

dialvarezs:
From the [ci logs](https://github.com/nf-core/mag/actions/runs/12074273648/job/33671976673#step:6:353) it seems to be lack of space in the runner.

dialvarezs:
Oh, I realized something now. I totally forgot to skip Bin QC if `params.skip_binqc` is set.

jfy133:
Glad you worked it out! 

Unfortunately since yesterday now have both kids home sick so I will only have time to test next week again (I'm really really sorry for this, apparently it's being a really bad year in my city in the kindergartens and schools for viruses and stuff 😣)

jfy133 (commented):
A few more last comments, but otherwise all I'm doing for the rest of the day is running the tests!

One major thing missing now though: We've removed the standalone checkm test in the `ci.yml`, but this has not been replaced by an alternative test, so now we have no checkM test... we should make sure all three are executed at least one across all our tests.

I suggest:

- `test.config`: run BUSCO (default, no change)
- `test_adapterremoval.config`: run checkm
- `test_bbnorm.config`: run checkm2

Does that make sense?

jfy133:
Checked on your PR vs dev:

- [x] Normal test (BUSCO) (busco_qc.tsv and binsummary.tsv comparable ✅)
- [x] Test with CheckM (checkm_qc.tsv and binsummary.tsv comparable ✅)
- [x] Test with CheckM2 (checkm_qc.tsv exists and binsummary.tsv has table merged with quast results ✅)
- [ ] Test with checkm2 + GUNC (check GUNC executes basically)

Sorry this has taken so long, but the tests are taking around 25-30m each time :roll_eyes: which doesn't help

modules/local/combine_tsv.nf
```diff
@@ -7,7 +7,7 @@ process COMBINE_TSV {
         'biocontainers/bioawk:1.0--hed695b0_5' }"
 
     input:
-    path(bin_summaries)
+    path(bin_summaries, stageAs: "bin_summaries/*.tsv")
```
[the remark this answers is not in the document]
dialvarezs: Yes, it updates the inputs to "1.tsv", "2.tsv" and so on. I did this to avoid filename clashing, because CheckM2 generates the summary as `${output_dir}/quality_report.tsv`.

jfy133:
Crap, BUSCO is borked everywhere isn't it :cry:

dialvarezs:
Yeah, it's already reported here: https://gitlab.com/ezlab/busco/-/issues/776 and it seems that the only solution is to update. Today I have looking at the code to update and migrate to the nf-core BUSCO module.

Do you mind if I integrate those changes in this PR, or do you prefer a new one?

For the meantime, can you give a review to the module update? https://github.com/nf-core/modules/pull/7199
I' getting an error with the meta that I'm not sure how to solve...

jfy133:
Self note: testing GTDBtk with following commands:

- `nextflow run dialvarezs/mag -r dev-checkm2 -profile test,mpcdf_viper --outdir ./results_dialvarezs_checkm2_gdtbk --binqc_tool checkm2 --skip_gtdbtk false --gtdb_db /ptmp/jfellowsy/databases/gtdb/release220/ --binqc_tool checkm`
- `nextflow run nf-core/mag -r dev -profile test,mpcdf_viper --outdir ./results_checkm_gdtbk --binqc_tool checkm2 --skip_gtdbtk false --gtdb_db /ptmp/jfellowsy/databases/gtdb/release220/ --binqc_tool checkm`

jfy133:
> Yeah, it's already reported here: https://gitlab.com/ezlab/busco/-/issues/776 and it seems that the only solution is to update. Today I have been looking at the code to update and migrate to the nf-core BUSCO module.
> 
> Do you mind if I integrate those changes in this PR, or do you prefer a new one?

Let's do a new one, it's breaking small MEGAHIT fix PR, so if it's separate we can pull into that one and this one at the same time.

> For the meantime, can you give a review to the module update? [nf-core/modules#7199](https://github.com/nf-core/modules/pull/7199) I' getting an error with the meta that I'm not sure how to solve...

Will have a look now!

jfy133:
Old BUSCO files should be back now!

And I'm waiting for my last GTDBTK related check but all the other runs looking good!

dialvarezs:
> Old BUSCO files should be back now!

That's good news. The BUSCO update PR ended up getting a bit large with the removal of all those local modules.

jfy133:
@dialvarezs My manual tests with GTDBTk work!

Just need the tests configs as in my comment above, but otherwise this is ready :D

dialvarezs:
@jfy133 great!
And I agree with you about the tests, I will add that right now.

jfy133 (approved):
If tests pass with the new configs...

image

jfy133:
Thank you very much @dialvarezs ! This is huge work!

dialvarezs:
There was a minor bug when CheckM was not run for certain bins (specifically eukaryotic ones). I updated the condition to check if the CheckM bins are a subset of the depth bins, rather than requiring them to be equal.

PR #716, opened 2024-11-23
Shortread subworkflow

muabnezor:
## PR checklist

This PR moves the short read preprocessing into a subworkflow


muabnezor:
Think this is ready for review @jfy133. Simple PR moving short read preprocessing into subworkflow.

subworkflows/local/shortread_preprocessing.nf
```diff
@@ -0,0 +1,188 @@
...
+include { FASTP                                                } from '../../modules/nf-core/fastp/main'
+include { ADAPTERREMOVAL as ADAPTERREMOVAL_PE                   } from '../../modules/nf-core/adapterremoval/main'
+include { ADAPTERREMOVAL as ADAPTERREMOVAL_SE                   } from '../../modules/nf-core/adapterremoval/main'
+include { BOWTIE2_REMOVAL_BUILD as BOWTIE2_HOST_REMOVAL_BUILD   } from '../../modules/local/bowtie2_removal_build'
+include { BOWTIE2_REMOVAL_ALIGN as BOWTIE2_HOST_REMOVAL_ALIGN   } from '../../modules/local/bowtie2_removal_align'
+include { BOWTIE2_REMOVAL_BUILD as BOWTIE2_PHIX_REMOVAL_BUILD   } from '../../modules/local/bowtie2_removal_build'
+include { BOWTIE2_REMOVAL_ALIGN as BOWTIE2_PHIX_REMOVAL_ALIGN   } from '../../modules/local/bowtie2_removal_align'
+include { CAT_FASTQ                                            } from '../../modules/nf-core/cat/fastq/main'
+include { SEQTK_MERGEPE                                        } from '../../modules/nf-core/seqtk/mergepe/main'
+include { BBMAP_BBNORM                                         } from '../..//modules/nf-core/bbmap/bbnorm/main'
```
jfy133, line 16: Please align :)

If you use VSCode, you should install the Nextflow extension which now includes a formatter you can activate via the command palette and  should do that for you :)

jfy133 (approved):
Looks like a straight forward conversion to me! thank you @muabnezor you're making the code base much cleaner!

subworkflows/local/shortread_preprocessing.nf
```diff
@@ -0,0 +1,188 @@
...
+    ch_short_reads_assembly = Channel.empty()
+
+    FASTQC_RAW(
+        ch_raw_short_reads
+    )
+    ch_versions = ch_versions.mix(FASTQC_RAW.out.versions.first())
+    ch_multiqc_files = ch_multiqc_files.mix(FASTQC_RAW.out.zip)
+
+    ch_bowtie2_removal_host_multiqc = Channel.empty()
+    if (!params.assembly_input) {
```
muabnezor, line 37: ```if (!params.assembly_input)``` in the future, it feels like we can move away from this. Could the assembly_input logic be determined by the users input?

subworkflows/local/shortread_preprocessing.nf
muabnezor: Will do. thanks for suggestion.

subworkflows/local/shortread_preprocessing.nf
jfy133: I was wondering about this actually myself:

If you've pre-assembled reads, they should already be cleaned. Which means you could skip this entire subworkflow, so maybe this if else statement could be pushed to the main `mag.nf` around the new subworkflow? @prototaxites what do you think, is that a reasonable assuemption to make (i.e., skip preprocessing of reads)

But what do you mean by 'users input'? This is basically that, if you give the additional extra sheet to `--assembly_input` (that contains the paths to contig FASTAS)  then we can skip assembly. It's not a boolean flag, if that is what you were meaning.

muabnezor: > But what do you mean by 'users input'? This is basically that, if you give the additional extra sheet to --assembly_input (that contains the paths to contig FASTAS) then we can skip assembly. It's not a boolean flag, if that is what you were meaning.

Ah, then this makes more sense. I thought it was a boolean.

muabnezor: > If you've pre-assembled reads, they should already be cleaned. Which means you could skip this entire subworkflow, so maybe this if else statement could be pushed to the main mag.nf around the new subworkflow?

yes, I actually did this first, but then moved it back into the subworkflow. Not sure what makes the most sense.

prototaxites: > If you've pre-assembled reads, they should already be cleaned. Which means you could skip this entire subworkflow, so maybe this if else statement could be pushed to the main mag.nf around the new subworkflow? @prototaxites what do you think, is that a reasonable assuemption to make (i.e., skip preprocessing of reads)

Yes, I think this probably makes the most sense. While I'm sure there's someone out there with the use-case that they want to map a new set of reads against an existing assembly? I think we should assume that the reads accompanying pre-computed assemblies are already clean - someone with that use case probably knows what they are doing!

Personally I'm in favour of putting the `if!(assembly_input)` line in the main WF rather than the subWF; the subWF code should only operate when we actually want to use it. Plus I think it helps make the logic of the main WF clearer by stating when and when not things are run.

muabnezor: Thanks for input. I think it makes sense to have it in the main workflow also. If you think this looks fine @jfy133 and @prototaxites I'll go ahead and merge.

workflows/mag.nf
```diff
@@ -199,145 +188,19 @@ workflow MAG {
...
-            }
-
-        CAT_FASTQ(ch_short_reads_forcat.cat.map { meta, reads -> [meta, reads.flatten()] })
-
-        // Ensure we don't have nests of nests so that structure is in form expected for assembly
-        ch_short_reads_catskipped = ch_short_reads_forcat.skip_cat.map { meta, reads ->
-            def new_reads = meta.single_end ? reads[0] : reads.flatten()
-            [meta, new_reads]
-        }
+        SHORTREAD_PREPROCESSING(
```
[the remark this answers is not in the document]
muabnezor: It is there, line 191, and 205?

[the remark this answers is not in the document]
jfy133: Yup you're right, I'm just blind :sweat:

jfy133 (approved):

jfy133:
You can merge @muabnezor !

PR #718, opened 2024-11-28
Longread only functionality

muabnezor:
This PR adds long-read only functionality to mag. 

## PR checklist


nextflow.config
```diff
@@ -53,6 +53,7 @@ params {
     min_length_unbinned_contigs          = 1000000
     max_unbinned_contigs                 = 100
     skip_prokka                          = false
+    longread_percentidentity             = 90
```
d4straub, line 56: thats missing in the `nextflow_schema.json`, right?

d4straub (commented):
I only had a short look, sorry, but I just wanted to express my gratitude that you tackle long read assembly!

conf/modules.config
```diff
@@ -551,6 +620,7 @@ process {
     }
 
     withName: METABAT2_JGISUMMARIZEBAMCONTIGDEPTHS {
+        ext.args = { meta.assembler in ['FLYE', 'METAMDBG'] ? "--percentIdentity ${params.longred_percentidentity}" : '' }
```
prototaxites, line 623: typo: longred_percentidentity

muabnezor:
> I only had a short look, sorry, but I just wanted to express my gratitude that you tackle long read assembly!

My pleasure hehe. Still WIP. I have to tidy up the code and run some validation on real data, but we're getting there!

conf/modules.config
muabnezor: Whoops, thank you!

muabnezor:
So far in this PR I have
- changed the validation schema to allow for long read only input
- Added a host removal track for long reads using minimap2 as aligner
- Added Flye and metaMDBG as longread assemblers 
- refactoring of code so that all the assembly code is moved to a subworkflow

Downstream from the assembly, except for the binning preparation, the long read and short read assemblies are treated the same.

muabnezor:
Me and some colleagues are running some tests on ont data, and I think this works as expected. We will continue testing and play around with the parameters, but if anyone wants to have a look through the code and suggest improvements, that would be awesome!

subworkflows/local/longread_binning_preparation.nf
```diff
@@ -0,0 +1,51 @@
...
+            .combine(ch_reads_minimap2, by: 0)
+            .map { id, meta_idx, idx, meta_reads, reads -> [ meta_idx, idx, meta_reads, reads ] }
+    }
+
+    ch_minimap2_input_reads = ch_minimap2_input
+        .map { meta_idx, index, meta, reads -> [ meta_idx, reads ] }
+    ch_minimap2_input_idx = ch_minimap2_input
+        .map { meta_idx, index, meta, reads -> [ meta_idx, index ] }
+
+    MINIMAP2_ASSEMBLY_ALIGN ( ch_minimap2_input_reads, ch_minimap2_input_idx, true, 'bai', false, false )
```
jfy133, line 39: What are the true/false/falses? something should be parameterasable by the user?

subworkflows/local/shortread_assembly.nf
```diff
@@ -0,0 +1,40 @@
...
+include { MEGAHIT                                               } from '../../modules/nf-core/megahit/main'
+include { SPADES as METASPADES                                  } from '../../modules/nf-core/spades/main'
+
+workflow SHORTREAD_ASSEMBLY {
+    take:
+    ch_short_reads_grouped           // [ [meta] , fastq1, fastq2] (mandatory)
+    ch_short_reads_spades
+
+    main:
+
```
jfy133, line 12:
```suggestion
```

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -244,12 +247,6 @@ def validateInputParameters(hybrid) {
     if (params.host_fasta && params.host_genome) {
         error('[nf-core/mag] ERROR: Both host fasta reference and iGenomes genome are specified to remove host contamination! Invalid combination, please specify either --host_fasta or --host_genome.')
     }
-    if (hybrid && (params.host_fasta || params.host_genome)) {
-        log.warn('[nf-core/mag]: Host read removal is only applied to short reads. Long reads might be filtered indirectly by Filtlong, which is set to use read qualities estimated based on k-mer matches to the short, already filtered reads.')
-        if (params.longreads_length_weight > 1) {
```
jfy133, line 249: Can you verify this warning is no onger necessary?

jfy133: (the one before is fine obviously)

workflows/mag.nf
```diff
@@ -112,6 +107,8 @@ workflow MAG {
 
     if (!params.keep_phix) {
         ch_phix_db_file = Channel.value(file("${params.phix_reference}"))
+    } else {
```
jfy133, line 110: Is this the fix for the [phix bug](https://github.com/nf-core/mag/issues/740)? If so would appreciate to split this out into a separate PR:

jfy133 (commented):
OK I'm not seeing anything obvious! But I would still like to run a test (can't do that today).

Once commetns (mostly questions) addressed, I will start the manual tests :)

muabnezor:
thank you @jfy133. I'm away this week, but I'll try to find time to go through your comments asap.

subworkflows/local/longread_binning_preparation.nf
muabnezor: first true: if the output should be in bam-format. I think this can be fixed
first false: If the output is set to paf, then a cigar string is generated. This only happens if output format is not bam, so does not make sense for it to be changed by user
second false: to make the cigar string backward compatible with older tools, but no need for that here.

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
muabnezor: The second warning, embedded in the first if statement might still be valuable, but only if filtlong is run. I updated the code here.

muabnezor:
Thanks for the review @jfy133. Sorry I have been away this week, but now I think I addressed all of your comments.

conf/modules.config
```diff
@@ -580,6 +644,7 @@ process {
     }
 
     withName: METABAT2_JGISUMMARIZEBAMCONTIGDEPTHS {
+        ext.args = { meta.assembler in ['FLYE', 'METAMDBG'] ? "--percentIdentity ${params.longread_percentidentity}" : '' }
```
prototaxites, line 647: Should this be more generally configurable for long and short reads?

docs/usage.md
```diff
@@ -43,14 +43,25 @@ sample2,0,0,data/sample2_R1.fastq.gz,data/sample2_R2.fastq.gz,data/sample2.fastq
 sample3,1,0,data/sample3_R1.fastq.gz,data/sample3_R2.fastq.gz,
 ```
 
+If only long read data is available, the columns `short_reads_1` and `short_reads_2` can be simply left empty:
```
prototaxites, line 46: This is processed with nf-schema, right? In which case, I think the short read columns can be left out in their entirety - don't need to be left empty

modules/nf-core/metamdbg/asm/main.nf
```diff
@@ -0,0 +1,62 @@
+process METAMDBG_ASM {
```
prototaxites, line 1: I pushed updates to the metaMDBG module that correct languageserver errors and updates the version to 1.1

nextflow.config
prototaxites, line 56: Why 90% - seems low? 

It's been removed from the VAMB documentation, but in an old readme file it states the following:

> In general, if reads are allowed to map to subjects with a nucleotide identity of X %, then it is not possible to distinguish genomes with a higher nucleotide identity than X % using co-abundance, and these genomes will be binned together. This means you want to tweak your alignment tool to only output alignments with the desired minimum query/subject identity - for example 97%.

So I wonder that this should be set to 97% and be made configurable as a single parameter for both long and short reads? Appreciate this might be different for ONT, but AFAIK even ONT quality is good nowadays, up to 99% - definitely not a 10% error rate!

https://github.com/RasmussenLab/vamb/blob/adeb157d3d16b77e240819b86e810f733d600b28/doc/tutorial.md

nextflow.config
```diff
@@ -65,6 +66,12 @@ params {
     skip_megahit                         = false
     skip_quast                           = false
     skip_prodigal                        = false
+    skip_metamdbg                        = false
+    skip_flye                            = false
+    flye_mode                            = 'nano-raw'
```
prototaxites, line 71: Would it be better to figure this (and metamdbg) out "on-the-fly" - add a field to the samplesheet that states ONT or HiFi, and these are selected accordingly? You can add a check that they're all ONT/all HiFi per group.

subworkflows/local/assembly.nf
```diff
@@ -0,0 +1,148 @@
+// SUBWORKFLOWS
+include { SHORTREAD_ASSEMBLY } from './shortread_assembly'
+include { LONGREAD_ASSEMBLY  } from './longread_assembly'
+include { HYBRID_ASSEMBLY    } from './hybrid_assembly'
+
+// MODULES
+include { POOL_SINGLE_READS as POOL_SHORT_SINGLE_READS          } from '../../modules/local/pool_single_reads'
+include { POOL_PAIRED_READS                                     } from '../../modules/local/pool_paired_reads'
+include { POOL_SINGLE_READS as POOL_LONG_READS                  } from '../../modules/local/pool_single_reads'
```
prototaxites, line 9: Could this local module be replaced with the nf-core CAT module (and called separately for R1 and R2 for paired reads)?

subworkflows/local/longread_hostremoval.nf
```diff
@@ -0,0 +1,58 @@
...
+        .map {
+            meta, reads ->
+                [ meta, reads, [] ]
+        }
+
+    // Generate unmapped reads FASTQ for downstream taxprofiling
+    SAMTOOLS_HOSTREMOVED_VIEW ( ch_minimap2_mapped , [[],[]], [] )
+    ch_versions      = ch_versions.mix( SAMTOOLS_HOSTREMOVED_VIEW.out.versions.first() )
+
+    SAMTOOLS_HOSTREMOVED_FASTQ ( SAMTOOLS_HOSTREMOVED_VIEW.out.bam, false )
```
prototaxites, line 39: This might be a place where a local module might be good - stream samtools view into samtools fastq. Otherwise you're essentially using extraneous storage in the work dir to hold a bam file you're never going to look at again, with the same data stored as fastq.

subworkflows/local/shortread_binning_preparation.nf
```diff
@@ -0,0 +1,55 @@
...
+        // combine assemblies (not co-assembled) with reads from own sample
+        ch_reads_bowtie2 = reads.map{ meta, reads -> [ meta.id, meta, reads ] }
+        ch_bowtie2_input = BOWTIE2_ASSEMBLY_BUILD.out.assembly_index
+            .map { meta, assembly, index -> [ meta.id, meta, assembly, index ] }
+            .combine(ch_reads_bowtie2, by: 0)
+            .map { id, assembly_meta, assembly, index, reads_meta, reads -> [ assembly_meta, assembly, index, reads_meta, reads ] }
+
+    }
+
+    BOWTIE2_ASSEMBLY_ALIGN ( ch_bowtie2_input )
```
prototaxites, line 43: Just a thought, but this change clashes with @jfy133's bowtie PR - might be worth thinking about which one to pull into the other.

prototaxites (commented):
Hi @muabnezor - looks really good! A few thoughts I had going through the PR - some probably easy fixes, some a bit more involved. 

Once it's ready to go, let me know and I can potentially run a test with some of our PacBio HiFi data, if that's of any interest to you.

nextflow.config
prototaxites: (this might require changes to the modules if we wanted to do this per-group)

muabnezor:
Thanks for the review @prototaxites! I will try to get to all your comments this week.

nextflow.config
muabnezor: Maybe, but this parameter can be any of (--pacbio-raw | --pacbio-corr | --pacbio-hifi | --nano-raw | --nano-corr | --nano-hq ), depending also on the quality of the data. I thought it is better for now to let the user specify the longread source instead.

subworkflows/local/assembly.nf
muabnezor: Maybe I have not looked into the pool module at all. maybe this is better to fix in another PR?

subworkflows/local/longread_hostremoval.nf
muabnezor: Good idea, I did something similar first, but later used the nf-core modules instead, as I thought this is preferred when possible :)

subworkflows/local/shortread_binning_preparation.nf
muabnezor: I noticed that too

nextflow.config
muabnezor: yes, I was unsure what the default  for this should be, I ran tests on some older ont data, and for those reads 97% was to high. but as you say ONT quality nowadays is probably ok to run with the default of 97%

docs/usage.md
muabnezor: you are right!

nextflow.config
prototaxites: That does complicate things... I just don't like that if you wanted to run both assemblers you would have to remember to specify two flags in the pipeline call. Maybe there could be a column in the input spreadsheet, but if you specify `flye_mode` (default null) then that overrides whatever is read from the column input?

muabnezor: I think you are right. Whats more minimap2 is also optimized to run with specific longread input. If we want to allow for the flexibility of several different longread sources in the same run, I guess we need to add a column in the samplesheet (what should be the allowed values for this column?), or maybe it would be sufficient to only allow for one longread technology, which can be specified by one parameter, --longread-technology ?

conf/modules.config
muabnezor: Fixed this! moved the logic here to binning.nf workflow instead, and added a parameter --shortread_percent_identity also

nextflow.config
prototaxites: Thinking about it, I'm not sure about allowing mixed input types as I think you're right that the complexity comes with the mapping for binning (if you map ONT to a HiFi assembly and vice-versa, and then try to compare co-abundance across assemblies - I think you might run into problems as the coverages may not be comparable...). So either we allow mixed input types but don't allow cross-mapping (single-sample only) - which will require writing some some complicated logic! - or force the user to specify a single parameter declaring the longread type, which sets a bunch of defaults that can be overridden by parameters that are by default null?

Or we could also just be conservative and say that the pipeline only supports uncorrected ONT and PacBio HiFi (no CLR)?

@jfy133 any thoughts?

prototaxites: Actually, do we need rail guards to ensure that we’re not mixing short read only and long read only data within a single run?

muabnezor: Sorry for late answer, I wasn't able to work on this last week. Actually I'm not sure what would be most sensible here. I agree that a guardrail might be a good idea so that short- and longread can't be mixed within a group. As for the different longread technologies, I feel that most users might only use one technology at a time for a run? so a single longead technology might be ok? What do you think @prototaxites and @jfy133, maybe we should discuss this further?

prototaxites: Happy to discuss this further - my gut feeling is that if you want to mix short and long read data, we either have to enforce hybrid mode, or enforce `binning_map_mode = own` - I don't know this for certain, but suspect that the statistical properties of coverage estimates differ among long and short reads (can imagine this is particularly in the case of mapping to repeats) that would affect binning outputs negatively.

I'm fine with the idea of enforcing a single long-read technology per-run - but this should be selected with a single user-specifyable parameter.

subworkflows/local/shortread_binning_preparation.nf
jfy133: Sorry for the slow progress on that, I'm in the final testing phases of the bowtie2 PR, so it might make sense to wait for that to go in then update this PR (as iwth the plan I described on slack).

nextflow.config
jfy133: > As for the different longread technologies, I feel that most users might only use one technology at a time for a run? so a single longead technology might be ok?

That's hard to say really. In a single research projet it's unlikely you would mix e.g. pacbio/nanopore, but if you're reanalysing public data you may do so...

That said, you would likely want to tweak parameters for each type so it might be worth  having separate runs for these.

That said adding an extra column to the input sheet for the sequencing method would be completely fine with me, I was thinking anyway that adding long-read only support is a very large functionality update so we could go with a mag v4 with the samplsheet change corresponding to the  breaking change?

You could follow the approach we have taxprofiler where we use ENA controlled vocab for the platform type.

> Happy to discuss this further - my gut feeling is that if you want to mix short and long read data, we either have to enforce hybrid mode, or enforce binning_map_mode = own - I don't know this for certain, but suspect that the statistical properties of coverage estimates differ among long and short reads (can imagine this is particularly in the case of mapping to repeats) that would affect binning outputs negatively.

Yeah I agree with the fear here, but I don't fully like enforcing hybrid mode because I can see cases where you want to compare hybrid vs long-read only - so I think I would prefer the latter option

jfy133: We could maybe have a short in person chat during mag dev hours on Thursday if you wanted?

prototaxites: Would be nice to actually join a mag dev hours :)

jfy133: Although this week I will be on truncated attendance so only 15:00-16:00 due to an overlapping meeting

muabnezor: I will be on dev hours 15:00 on Thursday then, but might need to leave before 16:00. See you then!

jfy133:
Conclusion from dev hours today:

- Add new optional columns: short_read_platform, long_read_platform -> use [controlled vocab] as options (https://github.com/enasequence/webin-cli/blob/d1cbb0b202245aa0e79878bcd6cc1f7bd2f8bb39/src/main/resources/uk/ac/ebi/ena/webin/cli/reads/platform.properties) ; if not supplied fallback is ILLUMINA and OXFORD_NANOPORE respectively
- Add a check that if within a given SR or LR group of samples, if there is more than one type platform - error if the mapping mode is _not_ `own`

jfy133:
(post corrections @muabnezor @prototaxites if I got that wrong)

prototaxites:
It was really nice to meet you both today :) 

> Add a check that if within a given SR or LR group of samples, if there is more than one type platform - error if the mapping mode is not own

I would phrase, add two checks:
- Within a group, error if not all platforms are the same
- Across groups, error if there is more than one sequencing platform and `binning_map_mode` is not `own`

muabnezor:
@prototaxites and @jfy133 Sorry I could not work on this the other week, but now I've added "platform checks". If there are more than one sequencing platform within a group, and the binning-map-mode is not own, mag will exit with error.

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -80,7 +80,7 @@ workflow PIPELINE_INITIALISATION {
     ch_raw_short_reads = ch_samplesheet.map { meta, sr1, sr2, _lr ->
         meta.run = meta.run == [] ? "0" : meta.run
         meta.single_end = params.single_end
-
+        meta.sr_platform = meta.sr_platform == [] ? "ILLUMINA" : meta.sr_platform
```
prototaxites, line 83: You can add the "default" key to the samplesheet schema - see https://nextflow-io.github.io/nf-schema/latest/nextflow_schema/nextflow_schema_specification/#default and https://nextflow-io.github.io/nf-schema/latest/nextflow_schema/sample_sheet_schema_examples/#nf-schema-test-case.

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -89,13 +89,36 @@ workflow PIPELINE_INITIALISATION {
         }
     }
 
+    // Check that all short read sequencing platforms are the same within a group
+    ch_raw_short_reads
+        .map { meta, sr -> [ meta.group, meta.sr_platform ] }
```
prototaxites, line 94: What happens here if `meta.group` is not specified? AFAIK it's an optional column and if you're not using `--coassemble_group` then downstream, `meta.group` is set to `meta.id`.

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -89,13 +89,36 @@ workflow PIPELINE_INITIALISATION {
...
+        .groupTuple(by: 0)
+        .map { group, platform ->
+            def platforms = platform.unique()
+            if (platforms.size() > 1 && params.binning_map_mode != 'own') {
+                exit(1, "Multiple short read sequencing platforms found for group ${group}: ${platforms.join(', ')}. Run with '--binning_map_mode own' or use same platform for all samples in a group.")
```
prototaxites, line 99: Is the recommended practice to use `log.exit()`?

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -89,13 +89,36 @@ workflow PIPELINE_INITIALISATION {
...
+            }
+        }
+
     ch_raw_long_reads = ch_samplesheet.map { meta, _sr1, _sr2, lr ->
         if (lr) {
             meta.run = meta.run == [] ? "0" : meta.run
+            meta.lr_platform = meta.lr_platform == [] ? "OXFORD_NANOPORE" : meta.lr_platform
```
prototaxites, line 106: Again, this can be enforced in the samplesheet schema :)

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -89,13 +89,36 @@ workflow PIPELINE_INITIALISATION {
...
             return [meta, lr]
         }
     }
 
+    // Check that all long read sequencing platforms are the same within a group
+    ch_raw_long_reads
+        .map { meta, lr -> [ meta.group, meta.lr_platform ] }
```
prototaxites, line 113: Same question as above about unspecified groupIDs

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
muabnezor: I'll fix that!

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -89,13 +89,36 @@ workflow PIPELINE_INITIALISATION {
         }
     }
 
+    // Check that all short read sequencing platforms are the same within a group
+    ch_raw_short_reads
+        .map { meta, sr -> [ meta.group, meta.sr_platform ] }
+        .groupTuple(by: 0)
+        .map { group, platform ->
+            def platforms = platform.unique()
+            if (platforms.size() > 1 && params.binning_map_mode != 'own') {
```
prototaxites, line 98: I think this will only trigger if there are multiple platforms within a group. When this is the case, we just want the pipeline to exit.

We need the `binning_map_mode` own check to trigger when different groups have different sequencing platforms.

Compare these two examples:

```
sample,group,platform
1,g1,ILLUMINA
2,g1,NOT_ILLUMINA,
3,g2,ILLUMINA
4,g2,ILLUMINA
```

Should exit (if --coassemble_group is on) as there is a group with mixed platforms.


```
sample,group,platform
1,g1,ILLUMINA
2,g1,ILLUMINA,
3,g2,NOT_ILLUMINA
4,g2,NOT_ILLUMINA
```

Should pass the within group check, but exit if `binning_map_mode` is not own.

prototaxites: In the case that `--coassemble_group` is off, `meta.id` should be treated as "group" and we should enforce `binning_map_mode` own if there is more than one sequencing platform per read type.

muabnezor: Shouldn't 
```
sample,group,platform
1,g1,ILLUMINA
2,g1,ILLUMINA,
3,g2,NOT_ILLUMINA
4,g2,NOT_ILLUMINA
```
be ok with binning_map_mode "own" or "group", but fail on "all" ? 

I might confuse the binning_map_mode options here, but I thought that "own" refers to one sample, not the group?

prototaxites: In the case of cosssembly, “own” refers to a single group, I think?

muabnezor: From binning_preparation.nf workflow
```
    } else {
        // i.e. --binning_map_mode 'own'
        // combine assemblies (not co-assembled) with reads from own sample
        ch_reads_bowtie2 = reads.map{ meta, reads -> [ meta.id, meta, reads ] }
        ch_bowtie2_input = BOWTIE2_ASSEMBLY_BUILD.out.assembly_index
            .map { meta, assembly, index -> [ meta.id, meta, assembly, index ] }
            .combine(ch_reads_bowtie2, by: 0)
            .map { id, assembly_meta, assembly, index, reads_meta, reads -> [ assembly_meta, assembly, index, reads_meta, reads ] }

    }
```

Here own refers to the own sample if I'm not mistaken? For coassembly, I find no use of "own", the ```--coassemble_group``` parameter is a boolean

prototaxites: It's a little bit convoluted, but this is my understanding of what is going on. When `coassemble_group` is off, `binning_map_mode` can be one of "own" - map reads only to the originating assembly; "group" - combinatorial mapping of all reads to all assemblies within a group; or "all', combinatorial mapping of all reads to all assemblies.

When `coassemble_group` is set, the `meta.id` field is overwritten with `group-${meta.group}`. That means that from a perspective of the code you linked above, `binning_map_mode` own is identical to `binning_map_mode` group. To map reads from each group across groups, you need to set "all". See https://github.com/nf-core/mag/blob/b2377f1629a0345276b16281642ed432d95b2c71/workflows/mag.nf#L316 for when this happens.

So we need to catch the following situations:

No coassembly:
- If all sequencing platforms identical for read type:
    - Any binning map mode
- If 2+ sequencing platforms for read type:
    - Binning map mode own

Coassembly:
- Within a group, all sequencing platforms must be identical or pipeline exits
- Between groups:
    - If all sequencing platforms identical for read type across groups (e.g. all groups have SR ILLUMINA):
        - Any binning map mode
    - If any groups differ in read type - e.g. a group with all ILLUMINA and one with all NOT_ILLUMINA
        - Binning map mode "own" or "group" enforced (these are identical)
        
On top of this, we probably also want to catch when there will be both short-read-only and long-read-only assemblies and enforce "own" in these cases.

There might be other cases I missed...

muabnezor: Yes, you are right, and I just got a headache hehe. I'll fix the checks.

muabnezor: I gave it another shot @prototaxites, let me know if you think it looks all right.

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -76,15 +76,53 @@ workflow PIPELINE_INITIALISATION {
...
+    // if coassemble_group or binning_map_mode is set to not 'own', check if all samples in a group have the same platform
+    ch_samplesheet
+        .map { meta, _sr1, _sr2, lr -> [ meta.group, meta.sr_platform, meta.lr_platform ] }
+        .groupTuple(by: 0)
+        .map { group, sr_platform, lr_platform ->
+            def sr_platforms = sr_platform.unique()
+            def lr_platforms = lr_platform.unique()
+            if (sr_platforms.size() > 1 || lr_platforms.size() > 1) {
+                if (params.coassemble_group) {
+                    exit(1,"Multiple short read sequencing platforms found for group. if --coassemble_group is used, use same platform for all samples in a group.")
```
prototaxites, line 88: Need separate error for long read case

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -76,15 +76,53 @@ workflow PIPELINE_INITIALISATION {
...
+                }
+                if (params.binning_map_mode != 'own') {
+                    exit(1,"Multiple long read sequencing platforms found for group. if --binning_map_mode is not 'own', use same platform for all samples in a group.")
```
prototaxites, line 91: Needs separate error for short read case

prototaxites:
```suggestion
                    exit(1,"Multiple long read sequencing platforms found for group ${group}. if --binning_map_mode is not 'own', use same platform for all samples in a group.")
```

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
prototaxites, line 88:
```suggestion
                    exit(1,"Multiple short read sequencing platforms found for group ${group}. if --coassemble_group is used, use same platform for all samples in a group.")
```

prototaxites (commented):
Couple of small points/suggestions. I /think/ this looks OK, but it might be worth running a suite of tests to make sure it errors correctly. Here's what I think the tests should show, though maybe I've missed something:

coassemble_group on:
- Mixed platforms within groups: fail
- Mixed platforms among groups + binning map mode == all: fail
- Mixed platforms among groups + binning map mode == group: pass
- Mixed platforms among groups + binning map mode == own: pass

coassemble_group off:
- Mixed platforms within groups: fail
- Mixed platforms + binning map mode == all: fail
- Mixed platforms + binning map mode == group: fail
- Mixed platforms + binning map mode == own: pass

jfy133:
Thanks for getting to this! I'm swamped with yet more Kita strikes, and preparing for the hackathon next week.

I will be able to catch up with mag after next week! But note you will need to resolve the conflicts so tests pass here too - but I'm liking the tests!

muabnezor:
Fixed the merge conflicts @jfy133, test fails for "latest-everything" with
```
Downloading plugin nf-schema@2.3.0
A workflow named 'UTILS_NFSCHEMA_PLUGIN' is already defined or included in script: /home/runner/work/mag/mag/subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```
I can't figure out why

prototaxites:
> Fixed the merge conflicts @jfy133, test fails for "latest-everything" with
> 
> ```
> Downloading plugin nf-schema@2.3.0
> A workflow named 'UTILS_NFSCHEMA_PLUGIN' is already defined or included in script: /home/runner/work/mag/mag/subworkflows/local/utils_nfcore_mag_pipeline/main.nf
> ```
> 
> I can't figure out why

See https://github.com/nextflow-io/nextflow/pull/5894

Should hopefully be fixed soon!

muabnezor:
hello @prototaxites and @jfy133,

I have now updated this PR

1. Run merging for long reads should work
2. Based on the long read platform provided in the samplesheet, the options for minimap2, flye, and metaMDBG are set thereafter. 

I have not enforced the use of sequencing platform in the samplesheet. Since downstream options are set depending on what platform the reads come from, I think the user should be encouraged to enter these anyway (otherwise they default to illumina for SR or nanopore for LR). If you still think these columns in the samplesheet should be required, then I'll fix that, but I will need to change the samplesheets for the tests as well.

prototaxites:
@muabnezor Are you happy with these changes? If so, I can try to find time this week to run a test on our PacBio data.

jfy133:
> et. Since downstream options are set depending on what platform the reads come from, I think the user should be encouraged to enter these anyway (otherwise they default to illumina for SR or nanopore for LR).

I think that information _is_ important for selecting downstream options, then I think it's reasonable to _require_ users to fill it in (and not have a default). So I would go with that, it should be just a single line in the `schema_input.json` no?

prototaxites:
> I think that information is important for selecting downstream options, then I think it's reasonable to require users to fill it in (and not have a default). So I would go with that, it should be just a single line in the schema_input.json no?

I'm not necessarily opposed to having a default "Illumina" for short reads, but I think there's enough diversity in long read sequencing that it would be preferable to force specification. But I think we should treat them the same, so we should just require the column 👍

muabnezor:
> I think that information _is_ important for selecting downstream options, then I think it's reasonable to _require_ users to fill it in (and not have a default). So I would go with that, it should be just a single line in the `schema_input.json` no?

Yes it is a simple fix, only I have to update the test-datasets mag branch as well

muabnezor:
> @muabnezor Are you happy with these changes? If so, I can try to find time this week to run a test on our PacBio data.

@prototaxites yes, I think you can go ahead and test this on pacbio data, that would be great!

muabnezor:
Giving sequencing platform in samplesheet is now required

jfy133:
Thanks @muabnezor ! Let's see how the pacbio data runs!

erikrikarddaniel (approved):
LGTM, but there's no `CHANGELOG.md` entry.

dialvarezs (approved):
lGTM!
Only need to fix the changelog (move changes to the dev section), and change nf-schema to 2.2.1 in to get tests passing.

docs/usage.md
```diff
@@ -12,45 +12,68 @@ The input data can be passed to nf-core/mag in two possible ways, either using t
...
+- `PACBIO_SMRT`: Pacific Biosciences Single Molecule Real-Time (SMRT) sequencing reads.
+
+These platform fields are important for downstream alignment and assembly tools.
+
 At a minimum CSV file should contain the following columns:
 
-`sample,group,short_reads_1,short_reads_2,long_reads`
+`sample,group,short_reads_1,short_reads_2,long_reads,short_reads_platform,long_reads_platform`
 
 The path to `long_reads` and `short_reads_2` is optional. Valid examples could look like the following:
```
prototaxites, line 31: and `short_reads_1`?

prototaxites:
Can we synchronise the naming schemes for the pre-processing parameters between long reads and short reads? Currently there is e.g. `skip_clipping` and `skip_adapter_trimming`

CHANGELOG.md
```diff
@@ -96,13 +106,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-| chopper |                  | 0.9.0       |
-| GUNC    | 1.0.5            | 1.0.6       |
-| nanoq   |                  | 0.10.0      |
+| Tool     | Previous version | New version |
+| -------- | ---------------- | ----------- |
+| CheckM   | 1.2.1            | 1.2.3       |
+| CheckM2  |                  | 1.0.2       |
+| chopper  |                  | 0.9.0       |
+| GUNC     | 1.0.5            | 1.0.6       |
+| nanoq    |                  | 0.10.0      |
```
jfy133, line 115: **suggestion:** 
```suggestion
| Tool    | Previous version | New version |
| ------- | ---------------- | ----------- |
| CheckM  | 1.2.1            | 1.2.3       |
| CheckM2 |                  | 1.0.2       |
| chopper |                  | 0.9.0       |
| GUNC    | 1.0.5            | 1.0.6       |
| nanoq   |                  | 0.10.0      |
```

jfy133:
I've merged lastest `dev` changes in, and updated CHANGELOG

I agree with @prototaxites in regards to the consistency

prototaxites:
Hi @muabnezor, having a bit of trouble with run merging in my local test.

Nextflow correctly runs 4 NANOPLOT_RAW jobs, but is only running 1 CHOPPER job - it looks like something is either not being consumed correctly, or (after discussing in the dev meeting with @jfy133) is possibly to do with variable overwriting, i.e. 

```
            } else if (params.longread_filtering_tool == 'chopper') {
                CHOPPER (
                    ch_long_reads,
                    ch_lambda_db.ifEmpty([])
                )
                ch_long_reads = CHOPPER.out.fastq
                ch_versions = ch_versions.mix(CHOPPER.out.versions.first())
            }
```

CHOPPER may be overwriting its own inputs.

Can you make sure that each step of long read preprocessing saves its results to an intermediate variable, to see if this fixes the problem? e.g. `ch_long_read_adapter_trimmed`, `ch_long_read_lambda_filtered`, `ch_long_read_host_removed`, etc..

---

samplesheet:
```
sample,run,group,long_reads,long_reads_platform
odRhaGlob1,1,0,/lustre/scratch123/tol/teams/tola/users/jd42/mag_testing/data/m84047_230720_114847_s3.ccs.bc2082.bam.fastq.gz,PACBIO_SMRT
odRhaGlob1,2,0,/lustre/scratch123/tol/teams/tola/users/jd42/mag_testing/data/m84098_240412_131355_s2.ccs.bc2020.rmdup.bam.fastq.gz,PACBIO_SMRT
odRhaGlob1,3,0,/lustre/scratch123/tol/teams/tola/users/jd42/mag_testing/data/m84098_240413_125930_s1.ccs.bc2024.rmdup.bam.fastq.gz,PACBIO_SMRT
odRhaGlob1,4,0,/lustre/scratch123/tol/teams/tola/users/jd42/mag_testing/data/m84309_250320_114609_s1.ccs.bc2024.rmdup.bam.fastq.gz,PACBIO_SMRT
```

params.yml:
```
skip_adapter_trimming: true
longread_filtering_tool: "chopper"
keep_lambda: true
host_fasta: /lustre/scratch123/tol/teams/tola/users/jd42/mag_testing/ref_g/GCA_964261875.1_odRhaGlob2.1_genomic.fna.gz
skip_gtdbtk: true
```

muabnezor:
Hi @prototaxites, I'll try to fix this today, thanks for pointing it out!

subworkflows/local/shortread_preprocessing.nf
```diff
@@ -79,13 +79,17 @@ workflow SHORTREAD_PREPROCESSING {
         ch_short_reads_prepped = ch_raw_short_reads
     }
 
-    if (params.host_fasta) {
+    if (params.host_fasta || params.host_genome) {
         if (params.host_fasta_bowtie2index) {
             ch_host_bowtie2index = file(params.host_fasta_bowtie2index, checkIfExists: true)
         }
         else {
+            ch_host_fasta_for_build = ch_host_fasta.combine(ch_short_reads_prepped)
```
muabnezor, line 87: @prototaxites and @jfy133, so the bowtie build for the host genome have been built even if no short reads is passed if a host genome is given, as @prototaxites pointed out. I came up with this as a solution, ch_host_fasta_for_build will be empty, if ch_short_reads is empty. I feel this is an ugly solution, but can't think of anything better now. do you have suggestions?

muabnezor:
> Nextflow correctly runs 4 NANOPLOT_RAW jobs, but is only running 1 CHOPPER job - it looks like something is either not being consumed correctly, or (after discussing in the dev meeting with @jfy133) is possibly to do with variable overwriting, i.e.

This is fixed. I changed so that ch_lambda_db is a value channel.

jfy133:
> > Can we synchronise the naming schemes for the pre-processing parameters between long reads and short reads? Currently there is e.g. `skip_clipping` and `skip_adapter_trimming`
> 
> Good suggestions, and of course you are right. Do you have any suggestions for the naming scheme? I'm thinking there will be cases where the user wish to, for example, adapter trim the short reads, but not the long reads. Sould we then have two parameters, e.g. --skip_longread_adapter_trimming, and --skip_shortread_adapter_trimming ?

Yes, definitely two separate ones - but just need to make sure the names are consistent. We can be more aggressive here as we will go for a mag v4 release which also gives us the freedom to 'break stuff'

prototaxites:
There seem to be four broad categories of pre-processing:

- Adapter trimming
- Quality filtering
- PhiX/Lambda removal
- Host removal

So perhaps we need a long and short read parameter for each of these broad stages?

muabnezor:
So for short reads we have `--skip_clipping`, skipping all adapter and filtering tools, for longreads we have `--skip_adapter_trimming` and `--skip_longread_filtering`, I'm not sure how to make this more consistent between over the two modes. Would it make sense to rename them to `skip_shortread_clipping` for short reads and 'skip_longread_adapter_trimming' and `skip_longread_filtering` for long reads? 

Host removal now is done automatically if the user provide a host_fasta, same with phix and lambda removal.

jfy133:
> So for short reads we have `--skip_clipping`, skipping all adapter and filtering tools, for longreads we have `--skip_adapter_trimming` and `--skip_longread_filtering`, I'm not sure how to make this more consistent between over the two modes. Would it make sense to rename them to `skip_shortread_clipping` for short reads and 'skip_longread_adapter_trimming' and `skip_longread_filtering` for long reads?

I would split the former into two - the more finegrained power the better

> Host removal now is done automatically if the user provide a host_fasta, same with phix and lambda removal.

Yeah I personally never liked that behaviour... I always want to have active 'opt in' parameters (and rather error/warn out if you supply a reference genome but not teh `--run_<xyz>`, but that's just me. What do you think @prototaxites

subworkflows/local/shortread_preprocessing.nf
jfy133: I don't really follow... why are reads needed for the reference indexing??

prototaxites: If short reads aren't provided, they should exist as an empty channel - combining an empty channel with a non-empty channel returns an empty channel. So it's a clean way of conditionally emptying a channel since you can't check channel contents in an `if()` block.

Personally I think it's quite a clean solution!

prototaxites:
> I would split the former into two - the more finegrained power the better

I would concur, but a quick glance at the functions of [fastp, trimmomatic, adapterremoval] suggests they are all-in-one tools that do both adapter trimming and quality filtering. Depending on how the command line arguments work for those tools, we could potentiallly separate those out. So something like:

- `skip_short_read_qc` - skips all of [fastp, trimmomatic, adapterremoval]
    - `skip_short_read_adapter_trimming` - disables adapter-trimming args of the above tools IF POSSIBLE
    - `skip_short_read_quality_filtering` - disables quality filtering/trimming args of the above tools IF POSSIBLE
- `skip_long_read_qc` - skips all of [porechop, filtlong, nanoq, chopper], with sub-filters:
    - `skip_long_read_adapter_trimming` - skips porechop, specifically
    - `skip_long_read_quality_filtering` - skips [filtlong, nanoq, chopper] specifically
- `keep_phix` - don't remove phix from short reads (maybe rename to `skip_phix_removal`)
- `keep_lambda` - don't remove lambda from long reads (maybe rename to `skip_lambda_removal`)
- `skip_short_read_host_removal` - (default: false) - have a specific flag to disable host removal even if host fasta specified
- `skip_long_read_host_removal` - (default: false) - have a specific flag to disable host removal even if host fasta specified

Some of these changes might be beyond the scope of this PR!

jfy133:
Sounds good, ok as this seems like a bigger change can we make an issue for it (I will do it in a second), and we can proceed with this PR with the current scope. I guess just fixing a few things still as listed above (e.g. db), or?

prototaxites:
Sounds like a plan! Let's try and harmonise these parameters before making the 4.0 release though - otherwise they will be a big interface change that would immediately necessitate a 5.0 if they came out after?

subworkflows/local/shortread_preprocessing.nf
jfy133: I still don't think I'm following.

But I mean if the FASTA genome will not be used even, if accidently provided, it's wasted time building the database. So you only index if the FASTA is provided and bowtie2index is false (i.e. not supplied)... why do you need to clear anything :sweat: 

In fact: if no short reads are supplied, then this whole subworkflow should not be executed in the first place?

prototaxites: Subworkflows don't work like modules, AFAIK - if they have two inputs, and some processes depend on one input and others on another - they don't wait for all the inputs to exist before starting. So providing an empty channel for short reads will skip the processes that require them, but the processes that require the host fasta (index building) can still proceed if they get an input, which they will if it is provided.

prototaxites: Unless I’m not understanding what you mean?

muabnezor:
This sounds like a good plan! 
I believe I have fixed the problems you pointed out @prototaxites, and I also realized that hybrid assembly with spades only worked for nanopore data, I changed it so that pacbio data can also be used for hybrid runs. 

Have you had the time to test the branch on your pacbio data @prototaxites ?

And sorry for being a bit unavailable the past weeks, I currently have a lot on my plate. I'll try to show up at the next dev-meeting!

prototaxites:
> Have you had the time to test the branch on your pacbio data @prototaxites ?

Not yet, need to re-start the run now that you've fixed the problem 👍

subworkflows/local/shortread_preprocessing.nf
muabnezor: Yes, exactly. This was the cleanest solution I could think of, since we can't check if a channel is empty.

prototaxites:
`POOL_LONG_READS` in `assembly.nf` (which pools long reads to a single file for Spades short-read assembly) runs even though no short reads are provided.

muabnezor:
> `POOL_LONG_READS` in `assembly.nf` (which pools long reads to a single file for Spades short-read assembly) runs even though no short reads are provided.

Solved this in a similar way as i did with the short-short read bowtie building if no short reads are given. i.e. combining the long-read channel with short-read channel on id. if the short-read channel is empty, then the long-read channel for pooling will also be empty.

conf/modules.config
```diff
@@ -306,6 +306,64 @@ process {
...
+            pattern: "*.bam",
+            enabled: params.save_hostremoved_reads
+        ]
+    }
+
+    withName: MINIMAP2_ASSEMBLY_ALIGN {
+        ext.args = {
+            meta.lr_platform == "OXFORD_NANOPORE" ? "-x map-ont" :
+            meta.lr_platform == "OXFORD_NANOPORE_HQ" ? "-x lr:hq" :
+            meta.lr_platform == "PACBIO_SMRT" ? "-x map-pb" : ""
```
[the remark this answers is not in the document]
muabnezor: At the moment the only pacbio longread platform that the user can give is `PACBIO_SMRT`. Maybe we should have `PACBIO_CLR` and `PACBIO_HIFI` instead?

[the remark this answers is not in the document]
muabnezor: or `PACBIO_SMRT_CLR` and `PACBIO_SMRT_CLR`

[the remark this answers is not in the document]
muabnezor: @prototaxites now the user have the options of `PACBIO_CLR` and `PACBIO_HIFI` and the options for minimap2 and flye should be set accordingly.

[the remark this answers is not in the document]
muabnezor: and well spotted!

subworkflows/local/longread_binning_preparation.nf
jfy133: OK cool :+1:

subworkflows/local/longread_hostremoval.nf
```diff
@@ -0,0 +1,59 @@
...
+
+    main:
+    ch_versions       = Channel.empty()
+    ch_multiqc_files  = Channel.empty()
+
+    ch_host_reference = reference.map { [ [:], it ] }
+    ch_host_fasta_for_build = ch_host_reference.combine(reads)
+                .map { host_meta, host_fasta, _meta, _reads ->
+                    [ host_meta, host_fasta ]
+                }.first() // makes sure to only use the host fasta if the long read channel is not empty
```
jfy133, line 26: It keeps tripping me up that you need the `reads` here for indexing... but if it works it works.

subworkflows/local/longread_hostremoval.nf
```diff
@@ -0,0 +1,59 @@
...
+    ch_host_fasta_for_build.view()
```
jfy133, line 27:
```suggestion
```

subworkflows/local/shortread_assembly.nf
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    if (!params.skip_megahit) {
+        MEGAHIT(ch_short_reads_grouped)
+        ch_megahit_assemblies = MEGAHIT.out.contigs.map { meta, assembly ->
+            def meta_new = meta + [assembler: 'MEGAHIT']
+            [meta_new, assembly]
+        }
+        ch_assembled_contigs = ch_assembled_contigs.mix(ch_megahit_assemblies)
+        ch_versions = ch_versions.mix(MEGAHIT.out.versions.first())
```
jfy133, line 32: Apparently there is now a MEGAHIT module for MultiQC, https://docs.seqera.io/multiqc/modules/megahit

So you should be able to take the `log` channel and mix that into multiQC here

subworkflows/local/shortread_binning_preparation.nf
jfy133: Yeah, let's merge this in then I re-do my PR

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -75,15 +75,53 @@ workflow PIPELINE_INITIALISATION {
...
+    // if coassemble_group or binning_map_mode is set to not 'own', check if all samples in a group have the same platform
+    ch_samplesheet
+        .map { meta, _sr1, _sr2, _lr -> [ meta.group, meta.sr_platform, meta.lr_platform ] }
+        .groupTuple(by: 0)
+        .map { group, sr_platform, lr_platform ->
+            def sr_platforms = sr_platform.unique()
+            def lr_platforms = lr_platform.unique()
+            if (sr_platforms.size() > 1 || lr_platforms.size() > 1) {
+                if (params.coassemble_group) {
+                    exit(1,"Multiple short- or long read sequencing platforms found for group ${group}. if --coassemble_group is used, use same platform for all samples in a group.")
```
jfy133, line 87:
```suggestion
                    error("[nf-core/mag] ERROR: Multiple short- or long read sequencing platforms found for group ${group}. if --coassemble_group is used, use same platform for all samples in a group.")
```

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -75,15 +75,53 @@ workflow PIPELINE_INITIALISATION {
...
+                }
+                if (params.binning_map_mode != 'own') {
+                    exit(1,"Multiple short- or long read sequencing platforms found for group ${group}. if --binning_map_mode is not 'own', use same platform for all samples in a group.")
```
jfy133, line 90:
```suggestion
                    error("[nf-core/mag] ERROR: Multiple short- or long read sequencing platforms found for group ${group}. if --binning_map_mode is not 'own', use same platform for all samples in a group.")
```

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -75,15 +75,53 @@ workflow PIPELINE_INITIALISATION {
...
+
+    // if binning_map_mode is set to 'all', check if all samples have the same platform
+    if (params.binning_map_mode == "all") {
+        ch_samplesheet
+            .map { meta, _sr1, _sr2, _lr -> meta.sr_platform }
+            .unique()
+            .collect()
+            .map {
+                if (it.size() > 1) {
+                    exit(1,"Multiple short read sequencing platforms found in samplesheet. Use same platform for all samples when running with binning_map_mode 'all'.")
```
jfy133, line 103:
```suggestion
                    error("[nf-core/mag] ERROR: Multiple short read sequencing platforms found in samplesheet. Use same platform for all samples when running with binning_map_mode 'all'.")
```

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -75,15 +75,53 @@ workflow PIPELINE_INITIALISATION {
...
+                }
+            }
+        ch_samplesheet
+            .map { meta, _sr1, _sr2, _lr -> meta.lr_platform }
+            .unique()
+            .collect()
+            .map {
+                if (it.size() > 1) {
+                    exit(1,"Multiple long read sequencing platforms found in samplesheet. Use same platform for all samples when running with binning_map_mode 'all'.")
```
jfy133, line 112:
```suggestion
                    error("[nf-core/mag] ERROR: Multiple long read sequencing platforms found in samplesheet. Use same platform for all samples when running with binning_map_mode 'all'.")
```

subworkflows/local/shortread_preprocessing.nf
jfy133: OK, for now if it works, it works - I can fix it myself in teh future if it still bothers me

jfy133 (commented):
I think only blocker is essentially the `multiMap` issue in the long read assembly thing subworkflow!

The rest are minor optimisations, the end of the tunnel is here @muabnezor !

I will later today or tomorrow do a couple of test runs with the test data and report back. But otherwise code wise it's just that last thing taht I saw!

assets/schema_input.json
```diff
@@ -40,19 +40,44 @@
                 "pattern": "^\\S+\\.f(ast)?q\\.gz$",
                 "errorMessage": "short_reads_2 needs to be a file path with no spaces.\n\nFile needs to exist!"
             },
+            "short_reads_platform": {
+                "type": "string",
+                "enum": [
+                    "ILLUMINA",
```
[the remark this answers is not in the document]
prototaxites: I think a future PR could branch on the meta.reads_platform value and feed reads in only if they match a tool’s expectation. But for now I think non-Illumina is rare enough that we should expect people who have it to know?

subworkflows/local/binning.nf
```diff
@@ -24,19 +25,30 @@ workflow BINNING {
...
+    // Process each through appropriate module
+    METABAT2_JGISUMMARIZEBAMCONTIGDEPTHS_LONGREAD ( ch_summarizedepth_input.longread )
+    METABAT2_JGISUMMARIZEBAMCONTIGDEPTHS_SHORTREAD ( ch_summarizedepth_input.shortread )
 
-    ch_metabat_depths = METABAT2_JGISUMMARIZEBAMCONTIGDEPTHS.out.depth.map { meta, depths ->
-        def meta_new = meta + [binner: 'MetaBAT2']
-        [meta_new, depths]
-    }
+    // Merge the outputs
+    ch_metabat_depths = METABAT2_JGISUMMARIZEBAMCONTIGDEPTHS_LONGREAD.out.depth
```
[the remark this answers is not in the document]
prototaxites: These objects do exist if they never receive an input - it’s only when they’re wrapped in some if block that you really need to be careful in my experience

muabnezor:
@jfy133 thanks for the review. I think I have addressed your suggested changes. I hope the tests go well!

subworkflows/local/longread_hostremoval.nf
muabnezor: hehe, as I mentioned I think this solution is a bit ugly, but could not come up with anything better. To be clear `reads` are not used in the indexing, it is only a way to condition when the index build should run. If the reads channel is empty, then the host fasta will also become empty, so, so no index will be built.

prototaxites:
FYI all - we have had a pretty catastrophic outage on our cluster and I'm not currently able to finish the test I was running - and it's unlikely I'll get a chance to finish it any time soon.

But I'm happy with the PR as-is - the long-read bits were working well after we fixed those wee bugs!

jfy133:
@muabnezor From my test - and now looking at `test_long` output here too, the final `bin_summary` process doesn't get executed.

Is that to be expected? This is one of the critical output files...

nextflow.config
```diff
@@ -54,6 +54,8 @@ params {
     min_length_unbinned_contigs          = 1000000
     max_unbinned_contigs                 = 100
     skip_prokka                          = false
+    longread_percentidentity             = null
+    shortread_percentidentity            = null
```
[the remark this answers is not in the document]
jfy133: @muabnezor is there any reason why you didn't set this?

jfy133 (commented):
Code-wise I'm happy now!

I just want to explore the results a bit more (see comment above about missing `bin_summary.tsv` process

muabnezor:
> @muabnezor From my test - and now looking at `test_long` output here too, the final `bin_summary` process doesn't get executed.
> 
> Is that to be expected? This is one of the critical output files...

no, this is not expected, are any of the channels from your test empty going into the process?

nextflow.config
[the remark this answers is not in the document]
muabnezor: No particular reason, but leaving them as null just uses the tools default, 97 anyway

jfy133:
> > @muabnezor From my test - and now looking at `test_long` output here too, the final `bin_summary` process doesn't get executed.
> > Is that to be expected? This is one of the critical output files...
> 
> no, this is not expected, are any of the channels from your test empty going into the process?

I'm running another long running test on my laptop at the moment so I can't check (unfortuantely)... any chance you could look? Otherwise I will do tomorrow

muabnezor:
> I'm running another long running test on my laptop at the moment so I can't check (unfortuantely)... any chance you could look? Otherwise I will do tomorrow

Sorry, can't get to that this week, I will have a look at it on monday

muabnezor:
@jfy133 
I'm looking into the bin_summary stuff, it ran fine a couple of commits ago, and I missed that it stopped running. good catch. I don't know if it is something I've done, or if the DEPTH process has had any changes recently, since I haven't touched anything after all the binning is done in this PR I think. I will try to get to what's wrong, but I'm currently having a lot of work to be done before summer vacations, so I don't know how much time I can spend on this.

jfy133:
I did some exploration last night: https://nfcore.slack.com/archives/CE9MS66BS/p1749154196789889?thread_ts=1748503770.520549&cid=CE9MS66BS

Essentially for some reason there is an extra meta being added somewhere meaning the reeads and bins can't be `combined` in DEPTHS()

@dialvarezs as you are waiting for this to be able to continue with your stuff, any chance you could have a look?

jfy133:
And to clarify @muabnezor it's not the DEPTHS subworkflow itself that is the issue, just the contents of the input channels :)

muabnezor:
@jfy133, there, I think I fixed it in the last commit in how depth channel is returned from binning workflow. I think (and hope!) this will fix it. Thanks for helping with this!

jfy133:
@muabnezor testing!

muabnezor:
@jfy133 just merged! Thanks for all the help with this PR!

Issue #750, opened 2025-01-22
Samtools warning on BAM file of reads mapped to assembly: The index file is older than the data file


amizeranschi:
I was running v. 3.0.2 of the pipeline, although I'm confident the issue would happen with the latest version as well, because the module hasn't been modified for 2 years now:

I assembled reads with SPAdes and saved the BAM files with `--save_assembly_mapped_reads`. In some cases, it seems that the BAM index file appears to be created/modified earlier than the BAM file itself. This causes LOTS of Samtools warnings such as `Warning: The index file is older than the data file`. I'm including a screenshot of the BAM and its index below. The timestamp of the index is 1 second earlier than the BAM file.

Looking at BOWTIE2_ASSEMBLY_ALIGN module, `samtools index` is being issued separately, after the BAM file has been created, so it's weird to see the index being older than the BAM file itself.

A simple fix could be to `touch` the index after it's been created.

![Image](https://github.com/user-attachments/assets/f4f38063-304e-4127-bae0-60b787b310f7)

jfy133:
@amizeranschi this is actually related to how Nextflow _publishes_ the files into the directory you specify with `--output`. 

Basically everything gets executed in the `work` directory, and then once completed, will be 'transferred' using the mode specified by `--publish_dir_mode` to the `--outdir`. 

By default this is copy, which of course as the `.bai` is smaller will finish copying to the `--outdir` much faster (and due to the async nature of Nextflow, may be copied first). E.g.: https://github.com/ablab/IsoQuant/issues/244#issuecomment-2369141485

If this is an issue, then you can try changing the `--publish_dir_mode` (full list of optiopns herE: https://www.nextflow.io/docs/latest/reference/process.html#publishdir under 'mode')

Given this does not cause an actual error within Samtools (just warnings), I will close this. 

I will also (sadly :( ) close your other PR as this will also not solve your problem as the `touch` helps the `work` directory, not what is copied to the `--outdir`. But please don't let that put you off 🙏 

Note I also checked my own runs, and didn't see this issue myself.

jfy133:
Ping me on slack if you want more details/don't full understand :)

PR #751, opened 2025-01-22
Update bowtie2_assembly_align.nf

amizeranschi:
Make sure the timestamps of the BAM file and its index are in proper order. Should fix https://github.com/nf-core/mag/issues/750


jfy133:
Unfortunately this doesn't solve the issue reported.

See description herE: https://github.com/nf-core/mag/issues/750#issuecomment-2607363807

PR #777, opened 2025-03-24
Json schema improvements

agusinac:
This PR updates the JSON schema to be more complete

### DIFF
--- a/assets/schema_input.json
+++ b/assets/schema_input.json
@@ -31,21 +31,21 @@
                 "format": "file-path",
                 "exists": true,
                 "pattern": "^\\S+\\.f(ast)?q\\.gz$",
-                "errorMessage": "short_reads_1 needs to be a file path with no spaces!"
+                "errorMessage": "short_reads_1 needs to be a file path with no spaces.\n\nFile needs to exist!"
             },
             "short_reads_2": {
                 "type": "string",
                 "format": "file-path",
                 "exists": true,
                 "pattern": "^\\S+\\.f(ast)?q\\.gz$",
-                "errorMessage": "short_reads_2 needs to be a file path with no spaces!"
+                "errorMessage": "short_reads_2 needs to be a file path with no spaces.\n\nFile needs to exist!"
             },
             "long_reads": {
                 "type": "string",
                 "format": "file-path",
                 "exists": true,
                 "pattern": "^\\S+\\.f(ast)?q\\.gz$",
-                "errorMessage": "long_reads needs to be a file path with no spaces!"
+                "errorMessage": "long_reads needs to be a file path with no spaces.\n\nFile needs to exist!"
             }
         },
         "required": ["sample", "group", "short_reads_1"]
--- a/assets/schema_input.json
+++ b/assets/schema_input.json
@@ -10,36 +10,42 @@
             "sample": {
                 "type": "string",
                 "pattern": "^\\S+$",
-                "meta": ["id"]
+                "meta": ["id"],
+                "errorMessage": "Sample needs to be string or integer with no spaces!"
             },
             "run": {
                 "type": ["string", "integer"],
                 "pattern": "^\\S+$",
                 "meta": ["run"],
-                "unique": ["sample"]
+                "unique": ["sample"],
+                "errorMessage": "Run needs to be string or integer with no spaces!"
             },
             "group": {
                 "type": ["string", "integer"],
                 "pattern": "^\\S+$",
-                "meta": ["group"]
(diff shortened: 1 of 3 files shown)


assets/schema_input.json
```diff
@@ -10,36 +10,42 @@
...
+                "meta": ["group"],
+                "errorMessage": "Group needs to be string or integer with no spaces!"
             },
             "short_reads_1": {
                 "type": "string",
                 "format": "file-path",
                 "exists": true,
-                "pattern": "^\\S+\\.f(ast)?q\\.gz$"
+                "pattern": "^\\S+\\.f(ast)?q\\.gz$",
+                "errorMessage": "short_reads_1 needs to be a file path with no spaces!"
```
nvnieuwk, line 34: can you also mention that the file needs to exist? (And also do this for each other file error message?)

nextflow_schema.json
```diff
@@ -214,16 +224,19 @@
...
+                    "hidden": true,
+                    "help_text": "Location of pipeline test dataset files"
                 },
                 "trace_report_suffix": {
                     "type": "string",
                     "fa_icon": "far calendar",
                     "description": "Suffix to add to the trace report filename. Default is the date and time in the format yyyy-MM-dd_HH-mm-ss.",
-                    "hidden": true
+                    "hidden": true,
+                    "pattern": "^\\d{4}-(0[1-9]|1[0-2])-(0[1-9]|[12]\\d|3[01])_(0\\d|1\\d|2[0-3])-[0-5]\\d-[0-5]\\d$"
```
nvnieuwk, line 239:
```suggestion
```
This suffix can be any string :)

nvnieuwk (changes requested):
Great work!!!! Here are some comments :)

assets/schema_input.json
agusinac: Ah yes I will do that!

agusinac:
I have encountered an unexpected behaviour when executing the following command:
`nextflow run main.nf -profile test, docker --outdir results -stub --bowtie2_mode "--very-sensitive"`.

The above results in the following warning:
`* --verySensitive: true`

This results in the json receiving a boolean instead of a string. The nextflow_schema.json, line 743 has been changed to account for the boolean. After discussing with @nvnieuwk we thought this should be brought to your attention @jfy133 
```
"bowtie2_mode": {
    "type": ["string", "boolean"],
    "description": "Bowtie2 alignment mode",
    "help_text": "Bowtie2 alignment mode options, for example: `--very-fast` , `--very-sensitive-local -N 1` , ... Must be used like this: --bowtie2_mode \"--very-sensitive\"",
    "pattern": "^\\'.*\\'$"
}
```

awgymer:
I believe the only correct way to supply a param value which starts with `--` is to use explicit equals:
```
 --bowtie2_mode="--very-sensitive"
 ```

**Comment by agusinac** [comment]
> Yeah I was personally never particularly happy with these flags, but it is what it is. Thanks for giving the solution @awgymer ! @agusinac could you check with Arthur's solution?

I checked with Arthur's solution and adapted the `nextflow_schema.json`. The json file now returns a string, the help_text and pattern keywords are updated.

**Comment by jfy133** [comment]
Thanks @agusinac !

File: nextflow_schema.json [hunk 3]
```diff
@@ -309,11 +316,13 @@
                 "adapterremoval_adapter1": {
                     "type": "string",
                     "default": "AGATCGGAAGAGCACACGTCTGAACTCCAGTCACNNNNNNATCTCGTATGCCGTCTTCTGCTTG",
+                    "pattern": "^[ATGCRYKMSWBDHVN]*$",
```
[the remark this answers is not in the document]
agusinac: It should accept any of the mentioned characters to be present either zero or more times. The adapter2 didn't contain any `N` and passed the test. I am happy to hear if this should be improved in the future!

nextflow_schema.json
```diff
@@ -196,28 +200,31 @@
...
                 },
                 "pipelines_testdata_base_path": {
                     "type": "string",
+                    "format": "directory-path",
                     "fa_icon": "far fa-check-circle",
                     "description": "Base URL or local path to location of pipeline test dataset files",
                     "default": "https://raw.githubusercontent.com/nf-core/test-datasets/",
-                    "hidden": true
+                    "hidden": true,
+                    "help_text": "Location of pipeline test dataset files"
```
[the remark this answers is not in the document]
agusinac: Yes, I went through the whole document and added keywords if they were missing.

PR #811, opened 2025-05-25
Remove usage of aria2 module

dialvarezs:
Remove a overly specific and a bit ugly patch of aria2 module to download and untar the CheckM db.
Updates the module.

And also update other "util" modules:
- cat/fastq
- gunzip
- untar

## PR checklist


jfy133:
[question](https://pullpo.io/cc?l=question) 
Can you explain the benefit of the aria2 patch removal bit of this PR?

(Unless I misunderstand it) You say 'unnecessary' patch but now you just move a patch to UNTAR instead.
Wouldn't it make more sense to just keep that patch in the one module to reduce the number of small jobs submitted to a cluster in this case?

dialvarezs:
IMHO, modules should do one thing, as they do in their pure form on nf-core, unless there is a clear advantage to coupling things together (for example, using a mapper like minimap2 and piping the output to samtools to get BAM or CRAM output instantly and reduce storage usage).

In mag we don't have many patched modules either, only this one and `UNTAR` (which I patched to be able to produce the BUSCO db directory without introducing another process). The aria2 patch in particular was very specific, it even had the directory names hardcoded, so it wouldn't be possible to use the module for anything else.

> Wouldn't it make more sense to just keep that patch in the one module to reduce the number of small jobs submitted to a cluster in this case?

That's a fair point, but in this case it's only one extra job, and only if you use CheckM. I guess there are other parts of MAG where this could be significantly improved. For example, `GUNZIP_BINS` is run per bin and produces a lot of small jobs, we could reduce to one per sample or something like that.

dialvarezs:
But, on second thought, maybe we could just use `UNTAR(file(url))` and say goodbye to the aria2 module. It's a <300 MB file, so I guess there's no drawback in letting Nextflow handle the download.

jfy133:
OK fair point @dialvarezs !

I think maybe we should go with your `UNTAR(file(url))` method.

I'm trying to remember why I made the standalone `aria2c` module... I _think_ it was due to downloads from the server was very flaky... but generally letting NExtflow download for you remains the recommeneded method via the Nextflow devs...

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Changed`
 
+- [#811](https://github.comf/nf-core/mag/811) - Update util modules, and remove aria2 module (by @dialvarezs)
```
jfy133, line 12:
```suggestion
- [#811](https://github.comf/nf-core/mag/811) - Update util modules, and remove aria2 module to replace with native Nextflow downloading of CheckM database (by @dialvarezs)
```

PR #815, opened 2025-05-27
Remove gtdb_db "exists" in schema to prevent nf-schema validation if gtdbtk is skipped

dialvarezs:
Removes `exists: true` in `gtdb_db` to prevent path validation in nf-schema.
Right now GTDB servers are down and this is causing all tests to fail, this update is to get them working again.

## PR checklist

### DIFF
--- a/conf/test.config
+++ b/conf/test.config
@@ -32,6 +32,7 @@ params {
     busco_db                    = params.pipelines_testdata_base_path + 'mag/databases/busco/bacteria_odb10.2024-01-08.tar.gz'
     busco_db_lineage            = 'bacteria_odb10'
     busco_clean                 = true
+    gtdb_db                     = null
     skip_gtdbtk                 = true
     gtdbtk_min_completeness     = 0.01
     skip_concoct                = true
--- a/conf/test.config
+++ b/conf/test.config
@@ -32,7 +32,6 @@ params {
     busco_db                    = params.pipelines_testdata_base_path + 'mag/databases/busco/bacteria_odb10.2024-01-08.tar.gz'
     busco_db_lineage            = 'bacteria_odb10'
     busco_clean                 = true
-    gtdb_db                     = null
     skip_gtdbtk                 = true
     gtdbtk_min_completeness     = 0.01
     skip_concoct                = true
--- a/conf/test_adapterremoval.config
+++ b/conf/test_adapterremoval.config
@@ -32,7 +32,6 @@ params {
     min_length_unbinned_contigs = 1
     max_unbinned_contigs        = 2
     binqc_tool                  = 'checkm'
-    gtdb_db                     = null
     skip_gtdbtk                 = true
     bin_min_size                = 750000
     bin_max_size                = 850000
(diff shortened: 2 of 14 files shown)


prototaxites:
I uploaded the mock GTDB database to the mag test-data branch here: https://github.com/nf-core/test-datasets/tree/mag/databases/gtdbtk

Maybe we can replace `gtdb_db` here with the URL to that, so we can run actual GTDB tests?

dialvarezs:
@prototaxites It doesn't seem to work, there is a problem with pplacer.
I will leave the nulls for now, and we can look deeper into the gtdb tests in another instance. Do you agree?

jfy133 (commented):
I think we need to make this actually make `null` default here.

The issue is that _any_ run by _any_ user will fail (even if they don't want to run GTDBTK!), because the default is set as the URL.

I personally never really liked this implicit running if you supply a URL.

I would rather have most heavy/optional processes to be executed with a `--run_gtdbtb --gtdb_db`.

That said:  time being to keep backwards compatibility maybe we can move the default URL into the code internally rather than in `nextflow.config`, so `if ( !skip_gtdb && !gtdb_null) { ch_gdtbk_tar = file(<URL>, checkIfExists) } elif (...)`

That way we don't have to set `gtdb_db` at all in the tests and nothing gets validated for it. We just need `skip_gtdb = true`.

Does that make sense?

(until we get the GTDB test data working - but I was planning on doing the whole config restructing during the nf-test test fest in 2 weeks)

jfy133:
> I'm not quite sure I follow your logic - or perhaps I'm just averse to hard-coding a URL in the pipeline code. The URL should only get accessed/staged if it gets passed into a process, AFAIK? So if `skip_gtdbtk` is set I would expect the URL remains unused. But perhaps I'm missing something.
> 
> In any case an alternative for tests would be to stage the GTDB (mock, once it's working) db inside the GHA YAML file and pass that directly to the pipeline as a command-line flag (if that's allowed). Avoids hard-coding a URL anywhere in the actual pipeline code while letting the runners download the DB and run the tests?

Yeah, I realise I wasn't probably clear.

To summarise the issue:

- Basically on one hand, we should support the pipeline auto-downloading the GTDB database.
- But we can't have the pipeline fail just because the database for downloading is done -> _particularly_ if the user doesn't want to run it.

Currently:

- We do 'hardcode' a URL in the nextflow.config, and this gets validated via nf-schema (if server is down, then it exists)
- This happens _regardless_ if `--skip_gtdbtk` is true

My suggestion was to fix the latter part:

- Moving the already hardcoded URL into the code to a place so it _doesn't_ get validated by nf-schema, but can still autodownload if something else is not supported.
- This allows us to manually validate and ONLY if `--skip_gtdbtk` is false

To be fair, the database downloading is so difficult with GTDB we could just drop auto-downloading of the database (but then it's out of sync with CAT).

The PR in it's current is slightly misleading as it only refers to fixing tests, but the issue of validating applies for ANY run (i've been having it today when testing the megatests)

Is that a better summary?

jfy133:
> I'm not a fan of putting URLs inside the pipeline code either. For me, config should be separated from logic (as it is now).
> 
> @prototaxites idea seems good to me, because we already do `file(path, checkIfExists: true)` in the pipeline code.

Oh, good point! Perfect solution @dialvarezs !

dialvarezs:
Ok, so this PR has become a oneliner now!

Issue #713, opened 2024-11-07
Be able to run multiple bin QC metrics


harper357:
### Description of feature

I posed this in the slack, but I think it would be great if it was possible to run multiple bin QC metrics (BUSCO, GUNC, CheckM2, etc) and get all of that data in the MultiQC report.

I noticed #707 which adds some new options, but I don't know if it is also allowing for multiple or not.

jfy133:
I think conceptually this makes sense however it's not so trivial as the output of each one goes into downstream process and possibly a custom script (not written by myself), so it might not be so trivial. It might take a bit of time to get to it

harper357:
Personally, even if it is just reported in the MultiQC report and not used for anything downstream I think it would be a nice thing to have.

Once the mentioned PR is merged, I can try to work up a new PR to try it out. 

Or maybe I should try getting work hours to write up a whole benchmarking pipeline so its easier to compare new software. That seems rather ambitious though.

jfy133:
So had a chance to quickly through the code:

- One area where we have the one or other for BUSCO/CheckM is filtering contigs for completeness/contamination for going into GTDBTkl but this could then be replaced by parameter to say 'which' of the two outputs should be used for that filtering 
- CheckM(1) is required for GUNC (busco doesn't work), but that doesn't block a user from also running BUSCO
- I was mistaken regarding the downstream custom script, that should be fine (BIN_SUMMARY), the script seems to allow inserting all

So actually this might be relatively easy to do in fact (conceptually) ino terms of running everthing  :) 

However only BUSCO is in MultiQC, so adding CheckM/GUNC etc. into MultiQC is a whole other task 😬

harper357:
Oh, hmm. Well maybe that is a better use of my time first then. 

Thanks for looking into this though!

jfy133:
> Oh, hmm. Well maybe that is a better use of my time first then. 


Which bit?

harper357:
Sorry, I meant writing some MultiQC plugins for CheckM, CheckM2 and GUNC.

jfy133:
Yes, that would be good! Divide and conquer (at least once I'm not travelling or have sick kids as I now have 😵‍💫🤦🏼‍♀️)

harper357:
Not sure if I should make a new issue or track just things here, but I have made some MultiQC module PRs:
GTDB (https://github.com/MultiQC/MultiQC/pull/2970)
[...]

PR #842, opened 2025-08-05
Support for running multiple bin QC tools

harper357:
This PR adds support for running multiple bin QC tools. Fixing issue #713 .

It does this by adding an extra parameter, `binqc_tool_extras` where users can specify extra tools to run and then just saving the outputs, but not using them for any downstream process.

This will probably have to be rebased after [PR#841](https://github.com/nf-core/mag/pull/841).

Not sure if this needs any new tests.

I have this marked as a draft because I am not completely happy with having to add a new parameter, and it doesn't add everything into the bin summary, but I am not sure if those are deal breakers. 

## PR checklist

- [X] Output Documentation in `docs/output.md` is updated.

### DIFF
--- a/subworkflows/local/utils_nfcore_mag_pipeline/main.nf
+++ b/subworkflows/local/utils_nfcore_mag_pipeline/main.nf
@@ -7,15 +7,15 @@
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 */
 
-include { UTILS_NFSCHEMA_PLUGIN     } from '../../nf-core/utils_nfschema_plugin'
-include { paramsSummaryMap          } from 'plugin/nf-schema'
-include { samplesheetToList         } from 'plugin/nf-schema'
-include { paramsHelp                } from 'plugin/nf-schema'
-include { completionEmail           } from '../../nf-core/utils_nfcore_pipeline'
-include { completionSummary         } from '../../nf-core/utils_nfcore_pipeline'
-include { imNotification            } from '../../nf-core/utils_nfcore_pipeline'
-include { UTILS_NFCORE_PIPELINE     } from '../../nf-core/utils_nfcore_pipeline'
-include { UTILS_NEXTFLOW_PIPELINE   } from '../../nf-core/utils_nextflow_pipeline'
+include { UTILS_NFSCHEMA_PLUGIN   } from '../../nf-core/utils_nfschema_plugin'
+include { paramsSummaryMap        } from 'plugin/nf-schema'
+include { samplesheetToList       } from 'plugin/nf-schema'
+include { paramsHelp              } from 'plugin/nf-schema'
+include { completionEmail         } from '../../nf-core/utils_nfcore_pipeline'
+include { completionSummary       } from '../../nf-core/utils_nfcore_pipeline'
+include { imNotification          } from '../../nf-core/utils_nfcore_pipeline'
+include { UTILS_NFCORE_PIPELINE   } from '../../nf-core/utils_nfcore_pipeline'
+include { UTILS_NEXTFLOW_PIPELINE } from '../../nf-core/utils_nextflow_pipeline'
 
 /*
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -62,7 +62,7 @@ workflow PIPELINE_INITIALISATION {
 \033[0;35m  nf-core/mag ${workflow.manifest.version}\033[0m
 -\033[2m----------------------------------------------------\033[0m-
 """
-    after_text = """${workflow.manifest.doi ? "\n* The pipeline\n" : ""}${workflow.manifest.doi.tokenize(",").collect { "    https://doi.org/${it.trim().replace('https://doi.org/','')}"}.join("\n")}${workflow.manifest.doi ? "\n" : ""}
+    after_text = """${workflow.manifest.doi ? "\n* The pipeline\n" : ""}${workflow.manifest.doi.tokenize(",").collect { "    https://doi.org/${it.trim().replace('https://doi.org/', '')}" }.join("\n")}${workflow.manifest.doi ? "\n" : ""}
 * The nf-core framework
     https://doi.org/10.1038/s41587-020-0439-x
 
@@ -71,7 +71,7 @@ workflow PIPELINE_INITIALISATION {
 """
     command = "nextflow run ${workflow.manifest.name} -profile <docker/singularity/.../institute> --input samplesheet.csv --outdir <OUTDIR>"
 
-    UTILS_NFSCHEMA_PLUGIN (
+    UTILS_NFSCHEMA_PLUGIN(
         workflow,
         validate_params,
         null,
@@ -80,7 +80,7 @@ workflow PIPELINE_INITIALISATION {
         show_hidden,
         before_text,
         after_text,
-        command
+        command,
     )
 
     //
@@ -360,8 +360,8 @@ def validateInputParameters(hybrid) {
 
     // Check if BUSCO parameters combinations are valid
     if (params.skip_binqc) {
-        if (params.run_busco || params.busco_db) {
-            error("[nf-core/mag] ERROR: Both --skip_binqc and --run_busco or --busco_db are specified! Invalid combination, please specify either --skip_binqc or --run_busco with --busco_db.")
+        if (params.run_busco) {
+            error("[nf-core/mag] ERROR: Both --skip_binqc and --run_busco specified! Invalid combination, please specify either --skip_binqc or --run_busco with --busco_db.")
         }
     }
 


prototaxites:
Hi, thanks for the PR - the code looks really clean! I have a couple of initial thoughts about the implementation which might be good to discuss before diving into proper review of the code itself.

> It does this by adding an extra parameter, binqc_tool_extras where users can specify extra tools to run and then just saving the outputs, but not using them for any downstream process.

I agree with you that this seems a bit clunky! I'm not 100% sure what the best option is here, but I can think of two options off the top of my head:

1. Have `--enable_busco`, `--enable_checkm`, `--enable_checkm2` parameters that turn on these tools, and `--binqc_tool` is explicitly defined as the tool used downstream. This might require reworking some logic to catch when they're all disabled but `--skip_binqc` is still on.

2. Have a single parameter, `--binqc_tools`, that takes a comma-separated list. The first listed tool is the one used downstream - we can be clear about it in the docs. It's cleaner from a parameterisation perspective, but maybe not so discoverable.

subworkflows/local/bin_qc.nf
```diff
@@ -94,16 +94,26 @@ workflow BIN_QC {
...
-        ch_qc_summaries = BUSCO_BUSCO.out.batch_summary
-            .map { _meta, summary -> [[id: 'busco'], summary] }
-            .groupTuple()
+        if (params.binqc_tool == "busco") {
+            ch_qc_summaries = BUSCO_BUSCO.out.batch_summary
+                .map { _meta, summary -> [[id: 'busco'], summary] }
+                .groupTuple()
+        }
+        else {
+            BUSCO_BUSCO.out.batch_summary
```
prototaxites: IMO I would prefer that these are published via the `publishDir` directive (and when it arrives in nf-core, the new publishing system) than using collectFile

dialvarezs:
IMHO option 2 feels more "natural", and we can clarify that the first tool is the one used to determine which bins get in the taxonomic classification. And them integrate the results from all of them in the bin summary.

prototaxites:
> IMHO option 2 feels more "natural", and we can clarify that the first tool is the one used to determine which bins get in the taxonomic classification. And them integrate the results from all of them in the bin summary.

I agree - but just want to flag up that we've explicitly avoided comma-separated lists for parameters in the past, so if we're going to do it we should be mindful about the interface

dialvarezs:
> > IMHO option 2 feels more "natural", and we can clarify that the first tool is the one used to determine which bins get in the taxonomic classification. And them integrate the results from all of them in the bin summary.
> 
> I agree - but just want to flag up that we've explicitly avoided comma-separated lists for parameters in the past, so if we're going to do it we should be mindful about the interface

That's fair, we should decide if we start including list parameters and probably refactor some parameter to be consistent if we decide too (binning comes to my mind, specially if we plan to include more binners in the future).

Personally, I don't see the downsides of using list parameters, they make the implementation of new tools simpler, and the user interface is good too. Other nf-core pipelines have them, for example sarek has a `tools` param (and it's huge, but we'll only have 3 options).

subworkflows/local/bin_qc.nf
harper357: Doesn't `publishDir` need to be inside a process? I had originally tried using calls to `csvtk/concat`, but had some trouble getting it actually publish the combined summary.

harper357:
@prototaxites Thanks for the quick comments!

I am happy to change things if there is a consensus

Option 1 is definitely the most explicit, but does make it more more difficult to add more tools in the future compared to what I did. I don't think it makes the logic too much more difficult, but maybe it will when/if Nextflow requires params are explicitly passed to workflows.

Option 2 I worry about using an ordered list. To me, this will make things confusing for new users since very few options are order dependent. It would require that people really read the documentation to understand how it works, which in my experience doesn't happen as often as it should.

As much as it feels clunky, my implementation does look like a good middle ground between these two options.

But again, I will change it if others think it should be done differently.

subworkflows/local/bin_qc.nf
prototaxites: It does (well, or in nf-core, defined in `modules.conf`). The way the subWF currently works passes everything to a `csvtk/concat` process called `CONCAT_BINQC_TSV` - it should be possible to either clone this code three times (and its config in `modules.conf`), and emit the three summaries separately (repetitive), or mix all the outputs from each binQC tool into a single channel with a `meta.binqc_tool`, group, and then feed it into the process for up to three iterations of the process. Then we can filter the right binQC summary using the meta object.

harper357: > (and its config in `modules.conf`)

Oh! This is why it didn't work for me originally. Ok, I changed it back and got it working.

harper357:
Hmm, I tried to rebase, but I think I might have messed up the commit history as it shows some extra changes in the file view

prototaxites:
> Option 1 is definitely the most explicit, but does make it more more difficult to add more tools in the future compared to what I did. I don't think it makes the logic too much more difficult, but maybe it will when/if Nextflow requires params are explicitly passed to workflows.

I think the problem with option 1 is that there's a potential incompatible selection of parameters - e.g. the only bin QC tool enabled is BUSCO, but `--binqc_tool` can still be set to "checkm2". I'd rather avoid having a situation where it's possible to specify breaking parameter combinations, if possible. 

> Option 2 I worry about using an ordered list. To me, this will make things confusing for new users since very few options are order dependent. It would require that people really read the documentation to understand how it works, which in my experience doesn't happen as often as it should.

While I agree that order-dependence would be a new interaction pattern with the pipeline, I don't think it in itself is particularly problematic, if it's documented well. I imagine many users would probably just want to run a single tool, in which case the option is fine and operates how it currently does, with no change to the interface - only users needing the 'advanced' feature would need to read the documentation. And hopefully if they get it wrong, the results would not be too dramatically different in any case.

My bigger problem with option 2 is that (aside from changing the schema from a string to a comma-separated list, as discussed above) it would also represent a switch from "on-by-default" tools with `--skip_X` parameters, to "off-by-default` tools which get enabled by specific parameters. 

(I actually prefer the latter, but don't think we should mix them)

harper357:
@prototaxites Just to make sure I understand you, are you suggesting that neither option 1 or 2 should move forward? Implying either my current implementation (for consistency lets call it option 0) or some new option needs to be thought of? 

Because I was thinking about your message over this last weekend and I think option 0 is still a nice middle ground. `--binqc_tool_additional` would be analogous to `run_gunc` while not messing with `--skip_binqc` or `--binqc_tool`. 

I am not sure if this is what you meant, but there are several cases of "off-by-default" tools/subworkflows in this pipeline, ex: `--ancient_dna` & `--run_virus_identification`, but I do agree that trying to stay consistent is important.

prototaxites:
> @prototaxites Just to make sure I understand you, are you suggesting that neither option 1 or 2 should move forward? 

To clarify, I still largely prefer option 2 - just pointing out that it would be a change compared to the way many parameters are specified. That's a good point about there already being some `--enable` or `--run` parameters, though.

@jfy133 Wonder if you have any opinions on this?

jfy133:
Finally catching up - and thank you for the patience.

Actually I _personally_ generally prefer to be more up-front declarative. I never really liked comma separated lists as user-wise they are more prone to user typos and validation of these means writing more custom validation code. Booleans are easier to handle with options etc. (and this is what we already do).

_That said_, I agree that we should not mix two approaches though and be considerate of not breaking 'backwards compatibility' too much.

Given we have `--bin_qc_tool 'busco'` as a string selection, I think it would be more user friendly to go with option 2 just to change this to allow a comma separated list (for people with existing params.yaml files, they won't necessarily have to change anything).

But diverging from option 2: not worry about ordering: we can the have the second flag then to pick the exact tool to use for downstream tools. When the number of binners in `--bin_qc_tool` is one, the default could be to take just that for the downstream parameter, otherwise when it is more than 1 we have validation code to error out unless the user specifies which one should go to e.g. GTDBTk etc.

What does everyone think?

jfy133:
Pseudo code:

```
params.bin_qc_tool = 'busco'

def nr_binning_tools = params.bin_qc_tool.split(',').length()
if (nf_binning_tools > 1 && !params.bin_downstream_tool) { error([nf-core/mag] ERROR: You specified to run multiple Bin QC tools but not which one to use downstream with --bin_downstream_tool. Please check input!) }
def binner_for_downstream =  nf_binning_tools > 1 && !params.bin_downstream_tool ? params.binc_qc_tool : bin_downstream_tool
```

prototaxites:
Could we mix both approaches? Have a comma-separated list of binqc tools and a `downstream` parameter, but we pick the first in the list (with a visible warning!) if the downstream parameter is not explicitly set? 

That way, the pipeline will run if a user forgets to specify both parameters - but notify them about the default action:

```
params.bin_qc_tool = 'busco'

def nr_binning_tools = params.bin_qc_tool.split(',').length()
if (nf_binning_tools > 1 && !params.bin_downstream_tool) { 
    log.warn("[nf-core/mag] WARN: You specified to run multiple Bin QC tools but not which one to use downstream with --bin_downstream_tool. Defaulting to the first in the list, which is ${params.bin_qc_tool.split(',')[0]}") 
}
def binner_for_downstream =  nf_binning_tools > 1 && !params.bin_downstream_tool ? params.binc_qc_tool.split(",")[0] : params.bin_qc_tool
```

In either case it also requires a check to make sure that `bin_downstream_tool` is actually in `bin_qc_tool`.

harper357:
> But diverging from option 2: 

> Could we mix both approaches? 

Doesn't option 0 already solve these with fewer changes? I am just a little confused because you are both talking about adding a new parameter to effectively replace `binqc_tool` and change the usage of `binqc_tool`, while option 0 keeps `binqc_tool` as is and just add an optional parameter. This maintains backwards compatibility, and doesn't require that people specify a tool in multiple places.

> more prone to user typos and validation of these means writing more custom validation code...

This is already taken care of by the parameter schema, which is a "simple" regex.

prototaxites:
My problem with option 0 is that it splits the same information (which bin QC tools to run) over multiple parameters. With this way, we specify the tools with one parameter and choose a downstream processor with another.

harper357:
I get that, I just think having a contextual parameter (`binqc_tool` being order dependent if you don't use `bin_downstream_tool` and not order dependent if you do) is more clunky having one parameter for "run and use this tool's output" and another for "run these tools and don't use the outputs".

If i am in the minority in thinking that, then ill stop trying to convince you

jfy133:
I'm still against using ordering to guide anything. It is an implicit behaviour which people will not intuitively understand based on typical Unix behaviour (which at least for me I can't think of any tool that uses ordering for things...).

My approach is that if you want to have your cake and eat it , you need to be responsible for it. I.e. if you truly want to run multiple QC tools, you need to be responsible in picking which one to use for downstream steps - this is an important choice that could have quite big ramifications ) thus what I mean by I like being more declarative and upfront)

That said I agree with Jim that I don't like splitting the same 'functionality' across two parameters.

We should always go broad to specific as you are narrowing down analysis to be more precise in downstream steps 

I also think that if you want to resume but switch QC tool you want to to use downstream, you would only have to switch the one parameter to pick another of your choices rather than shuffle options between the two:


```
--binqc_tool checkm,busco,checkm2
--downstream checkm

## to

--binqc_tool checkm,busco,checkm2
--downstream busco
```

Va


```
--binqc_tool checkm
--extra busco,checkm2

## to

--binqc_tool busco
--extra checkm,checkm2
```

(Assuming I am understanding option 0 directly)

prototaxites:
Hi all, discussed this briefly with James on Slack over the weekend - I think I have a solution that is better and avoids the main problems we've been circling around above.

We should just abandon the idea of the "downstream bin QC tool" entirely, and instead, all bins that meet the minimum quality threshold for any of the enabled bin QC tools should enter GTDB-Tk for classification. Essentially, we should assume that if a user runs multiple binQC tools, they are interested in comparing the sets of bins that result, which means we need to make sure that the analyses are standardised.

This has some advantages:

- We can switch the pipeline parameterisation to have `--enable_{busco|checkm{2}}`...  parameters as boolean flags, which is in keeping with the rest of the pipeline. No order dependence, no text strings.
- GTDB-Tk is a resource-intensive process and can take a long time - if a user changes their mind after running the pipeline and wants to switch tools, they would have to re-run it from scratch using the previously discussed schema - but this way, they will have results for all possible passable bins
- Doesn't requre us to have logic gates/write warnings to validate and parse parameters - works with any number of bin QC tools (including none - we can pass all bins in this case)

Though it does have a couple of cons:

- If we switch to boolean flags, this will break existing parameter configurations.
- We assume users "know what they are doing" regarding comparing these outputs

In terms of how to process the data, the GTDB subworkflow can just take the three binQC TSV channels (possibly empty), process them, and then `groupTuple` and `collect`:

```
take:
ch_busco_tsv
ch_checkm_tsv
ch_checkm2_tsv

main:

ch_busco_scores = ch_busco_tsv
    | splitCsv ... 
    | map { ... -> [bin_id, [completeness, contamination] }

&c. for checkm, checkm2

ch_scores = ch_busco_scores.mix(ch_checkm_scores, ch_checkm2_scores)
ch_scores.groupTuple(by: 0 ).map { bin, score ->  score.collect{ it[1] >= comp_thresh & it[2] =< contam_thresh}.any() }
```

Thoughts?

dialvarezs:
> We should just abandon the idea of the "downstream bin QC tool" entirely, and instead, all bins that meet the minimum quality threshold for any of the enabled bin QC tool

100% agree about using all the enabled tools to determine which bins pass the threshold for gtdbtk, that's the most sane option in that sense.

> We can switch the pipeline parameterisation to have --enable_{busco|checkm{2}}

Just to confirm on the parameters: if we leave one enabled by default, and to follow the principle of "no boolean flags turned on by default", if I a user wants to use CheckM2 instead of BUSCO (assuming that's the default), will have to turn on two parameters: one for disabling BUSCO and another for enabling CheckM2?

So, the command would look like this:
```
nextflow run nf-core/mag --disable_busco --enable_checkm2 ...
```

prototaxites:
> Just to confirm on the parameters: if we leave one enabled by default, and to follow the principle of "no boolean flags turned on by default", if I a user wants to use CheckM2 instead of BUSCO (assuming that's the default), will have to turn on two parameters: one for disabling BUSCO and another for enabling CheckM2?

Personally I am not bothered by a mix of different defaults for boolean flags - I'd rather have a consistent interface for a class of parameters (`--enable_busco/--enable_checkm/--enable_checkm2`) with some turned on/off by default, than have to know off the top of my head which are turned on by default and need to be unset with a `--disable` flag.

harper357:
>We should just abandon the idea of the "downstream bin QC tool" entirely, and instead, all bins that meet the minimum quality threshold for any of the enabled bin QC tool

Ok, if most other people are good with it, than I have no major objections. 

I probably wont be able to do much on this PR until at least next week, but there are some minor questions that come to mind,

1) all of this talk has been ignoring GUNC, should it be included in this or just left alone? 
    a. Should `run_gunc` be renamed to be to be consistent with the others (ex: `enable_gunc`)?
2) i always forget, but is it "Nextflow strict syntax" that says parameters should only be called in main.nf? With this in mind, should we generate an array of which tools to run and just pass it to the bin QC subworkflow?

jfy133:
> > We should just abandon the idea of the "downstream bin QC tool" entirely, and instead, all bins that meet the minimum quality threshold for any of the enabled bin QC tool
> 
> Ok, if most other people are good with it, than I have no major objections.
> 
> I probably wont be able to do much on this PR until at least next week, but there are some minor questions that come to mind,
> 
>     1. all of this talk has been ignoring GUNC, should it be included in this or just left alone?
>        a. Should `run_gunc` be renamed to be to be consistent with the others (ex: `enable_gunc`)?
> 
>     2. i always forget, but is it "Nextflow strict syntax" that says parameters should only be called in main.nf? With this in mind, should we generate an array of which tools to run and just pass it to the bin QC subworkflow?

1. IIRC GUNC has a _slightly_ different purpose which doesn't fit into the classical definition of contamination (and doesn't support completeness, I think?) so may not fit here... but happy to be corrected
2. We _can_ ignore that for the time being as the current pipeline is doing so 😬 but I'm personally happy to start a gradual transition towards this if everyone agrees? (A wholesale switch will be a big pain in the butt)

harper357:
Sorry for not having any new updates on this lately. Life got crazy, but I am hoping to get back to this next week.

jfy133:
No worries @harper357 ! Looking forward to it!

jfy133:
So I just spent 20 minutes merging the latest release into your branch @harper357 and then...

```
codespace /workspaces/mag (multiple_bin_qc) $ git push origin
remote: Permission to harper357/mag.git denied to jfy133.
fatal: unable to access 'https://github.com/harper357/mag/': The requested URL returned error: 403
codespace /workspaces/mag (multiple_bin_qc) $ git push upstream
remote: Permission to nf-core/mag.git denied to jfy133.
fatal: unable to access 'https://github.com/nf-core/mag.git/': The requested URL returned error: 403
codespace /workspaces/mag (multiple_bin_qc) $ 
``` 

:melting_face: :melting_face: :melting_face: :melting_face: :melting_face: 

For reference, I merged in the nf-core/mag `dev` branch, accepting all incoming changes, then re-constructed your changes with

 
[multiple_bin_qc_merge.patch](https://github.com/user-attachments/files/22674345/multiple_bin_qc_merge.patch)


As this PR would be really useful for https://github.com/nf-core/mag/pull/861 (and thus currently blocking), would you like us to take over to make the required changes (but you retain credit?)

If so, what I would suggest is be that I make a new branch from dev on this repo, re-direct the target of this PR to that one, we then just merge in 'as is' (so you keep the commits), and then I can finish up based on the discussion above. 

Would that be OK? (I also suggest this as don't want to add more pressure if crazy time as you said before!)

harper357:
Hi @jfy133 ,

I think I have finally blocked out some time today and tomorrow to get back to working on this, so if don't mind waiting until Monday to take it over, that would be great. 

I think it is mostly straight forward changes with maybe the exception of the bin_summary since there could be some column name conflicts. If that is the case, I am not 100% on the best way to fix that. Maybe just add a prefix or suffix?

jfy133:
> Hi @jfy133 ,
> 
> I think I have finally blocked out some time today and tomorrow to get back to working on this, so if don't mind waiting until Monday to take it over, that would be great. 

Oh great, perfect timing! 

That's completely fine! Less work for me the better ;)

Thank you!

> I think it is mostly straight forward changes with maybe the exception of the bin_summary since there could be some column name conflicts. If that is the case, I am not 100% on the best way to fix that. Maybe just add a prefix or suffix?

Yeah just add a suffix now that is fine :)

jfy133:
> > Hi @jfy133 ,
> > 
> > I think I have finally blocked out some time today and tomorrow to get back to working on this, so if don't mind waiting until Monday to take it over, that would be great. 
> 
> Oh great, perfect timing! 
> 
> That's completely fine! Less work for me the better ;) thank you!
> 
> > I think it is mostly straight forward changes with maybe the exception of the bin_summary since there could be some column name conflicts. If that is the case, I am not 100% on the best way to fix that. Maybe just add a prefix or suffix?
> 
> Yeah just add a suffix now that is fine :)
>

harper357:
i think i need one more day (monday) to work on this if that is ok.

harper357:
@jfy133 if you have time to work on this in the next few days, I might have to hand it off to you. I won't have time to work on it again for a few days. 

In short, I added TODOs to all the spots/files I think need updating and then started from the beginning of the pipeline updating things. I was trying to keep all the enabled tools summaries in a single channel so it would be easier to add more down the line.

This means that GTDBTK would need to just a single "summaries" channel, split it out and then process each file. Slightly different than what @prototaxites suggested, but hopefully you two like the idea.

Files needing updating:  
- [x] nextflow.config 
- [x] nextflow_schema.json
- [x] modules.config
- [ ] mag.nf (partial)
- [x] bin_qc/main.nf
- [ ] gtdbtk/main.nf
- [ ] bin_summary/main.nf
- [ ] combine_tables.py

jfy133:
Ok, so for the last bit of code changes, I think all we need to do is modify these lines:

https://github.com/nf-core/mag/blob/main/subworkflows%2Flocal%2Fgtdbtk%2Fmain.nf#L26-L37

To more dynamically perform the operations based in a meta field (that says which tool it's from, hopefully that's already there), rather than 'hard coding' it via a parameter

workflows/mag.nf
```diff
@@ -386,7 +386,11 @@ workflow MAG {
         if (!params.skip_binqc) {
             BIN_QC(ch_input_for_postbinning)
             ch_versions = ch_versions.mix(BIN_QC.out.versions)
-            ch_bin_qc_summary = BIN_QC.out.qc_summary
+            // TODO update, probably needed to update. Just remember this is now a list
```
jfy133:
```suggestion
```

workflows/mag.nf
```diff
@@ -447,15 +451,16 @@ workflow MAG {
         else {
             ch_gtdbtk_summary = Channel.empty()
         }
-
+        // TODO update
```
jfy133:
```suggestion
```

workflows/mag.nf
```diff
@@ -564,7 +569,7 @@ workflow MAG {
     if (!params.skip_binning && !params.skip_prokka) {
         ch_multiqc_files = ch_multiqc_files.mix(PROKKA.out.txt.collect { it[1] }.ifEmpty([]))
     }
-
+    // TODO update to get all QC files
```
jfy133:
```suggestion
```

jfy133:
OK @mag-team, to summarise my changes

- Updated `combine_tables.py` to join to the table all/any of the QC tools not just one, and add suffixes to all columns of these with the tool name
- Greg already added publication of combined QC results of each of the tools
- Tweaked  the `test_altneratives` and `test_full`, and `single_end` to make sure all QC tools are tested at least once as we had in the previous structure with `--bin_qc_tool`
- Greg updated the `usage.md` already to refernce the new per-tool concatenated QC results, but I also updated the description of the `bin_summary.tsv`
- Update the BIN_SUMMARY process to match `combine_tables.py` 
- Deprecated `--bin_qc_tool` and replaced with per tool `--run_<tool_name>` everywhere, separating the original singular `ch_qc_summaries` with per tool summary channels
- In `bin_qc` in addition to the above, generate 'splitCSV'd version of the concatenated QC results, for GTDBtK filtering
- In `gtdbtk` added the bin_qc name as a meta field, and updated the code to combine with the bins, do the filtering, but also ensure we only pass a bin a single time (as in, if it passes the filters of two tool's metrics, we only report one)
- Update the initialisation validation checks to match the new parameters

conf/modules.config
```diff
@@ -511,9 +511,33 @@ process {
...
+        publishDir = [
+            path: { "${params.outdir}/GenomeBinning/QC" },
+            mode: params.publish_dir_mode,
+            saveAs: { filename -> filename.equals('versions.yml') ? null : filename },
+        ]
+    }
 
-    withName: CONCAT_BINQC_TSV {
-        ext.prefix = { "${params.binqc_tool}_summary" }
+    withName: 'CONCAT_BUSCO_TSV|CONCAT_CHECKM_TSV|CONCAT_CHECKM2_TSV' {
```
dialvarezs, line 515: There's a bit redundancy here I guess
Maybe we can keep the publishDir in the shared one, and the prefix in the individuals? (Asumming that Nextflow likes that).

subworkflows/local/bin_qc/main.nf
```diff
@@ -200,15 +199,35 @@ workflow BIN_QC {
...
-    CONCAT_BINQC_TSV(ch_qc_summaries, 'tsv', 'tsv')
-    ch_versions = ch_versions.mix(CONCAT_BINQC_TSV.out.versions)
-
-    ch_qc_summary = CONCAT_BINQC_TSV.out.csv.map { _meta, summary -> summary }
+    ch_busco_final_summaries = Channel.empty()
+    ch_checkm_final_summaries = Channel.empty()
+    ch_checkm2_final_summaries = Channel.empty()
+
+    if (params.run_busco) {
+        CONCAT_BUSCO_TSV(ch_busco_summaries, 'tsv', 'tsv')
```
dialvarezs: I would prefer to put the concat steps inside the if block where the tool runs, not in new if blocks.

subworkflows/local/gtdbtk/main.nf
```diff
@@ -67,9 +78,9 @@ workflow GTDBTK {
         .count()
         .map { it == 0 ? log.warn("No contigs passed GTDB-TK min. completeness filters. GTDB-TK summary will execute but results will be empty!") : "" }
 
-
+    // We set unique to remove duplicate bins when the same bin has passed the filter of two different QC tools
     GTDBTK_CLASSIFYWF(
-        ch_filtered_bins.passed.groupTuple(),
+        ch_filtered_bins.passed.unique().groupTuple(),
```
dialvarezs: Discarded bins will have the same problem, and the worse is that there could be bins shared between passes and discarded.
Maybe we need a new way to get the discarded bins, filtering out the passed from the complete list comes to my mind.

prototaxites: Can we instead group all the data into a single channel entry before filtering:

```
[ bin_name.fa, [[90, 10], [82, 5], [70, 6]] ]
```

Where the list is all the pairs of completeness and contam scores.

You can then iterate over it like:

```
ch_bins_filtered = ch_bins.filter { meta, scores ->
    scores.collect { it[0] >= comp_thresh && it[1] <= cont_thresh }.any()
}
```

This way you only get a single list for both.

conf/modules.config
jfy133: I think these files should be for all bins across samples, so the independent ones are not necessary, I'll retain just the combined ones.

subworkflows/local/bin_qc/main.nf
jfy133: Done!

subworkflows/local/gtdbtk/main.nf
dialvarezs: That would be a better design indeed!

jfy133: Good points both - I definitely didn't consider this (got too carried away about getting python to work :laughing: )

I'm going to come back to this - I've hit a slightly different error in `test_single_end` when concoct is activated, then I come back to this

jfy133:
Thank you to @dialvarezs he pointed out we need to set `busco_clean                 = true` in test_alternatives.config, to get that last test to run

subworkflows/local/gtdbtk/main.nf
jfy133: We talked about this, and agree with the approach that Jim says above, except that instead of filter we still just the branch (just changing the condition), so that we still retain teh 'discarded' stuff that can go into GDTBKT_SUMMARY.

harper357:
I didn't have a chance to run it, but it looking at it, i think it makes sense and looks good.

My only question is did the documentation get updated to explain how running multiple QC tools works? Just something like "Bins will be considered passing filter if they pass on completeness AND contamination for ANY tool. For example a bin will pass if it is passes completeness using BUSCO and passes contamination using CheckM2." I am just imagining some someone running it and then wondering why a bin passed if no single tool passed it.

subworkflows/local/gtdbtk/main.nf
```diff
@@ -61,23 +74,22 @@ workflow GTDBTK {
...
         ch_filtered_bins.passed.groupTuple(),
         ch_db_for_gtdbtk,
         params.gtdbtk_pplacer_useram ? false : true,
     )
     ch_versions = ch_versions.mix(GTDBTK_CLASSIFYWF.out.versions)
 
+    // Print warning why GTDB-TK summary empty if passed channel gets no files
+    ch_filtered_bins.passed
+        .count()
+        .map { it == 0 ? log.warn("No contigs passed GTDB-TK min. completeness filters. GTDB-TK summary will execute but results will be empty!") : "" }
```
prototaxites:
```suggestion
    ch_filtered_bins.passed
        .count()
        .combine(ch_filtered_bins.discarded.count())
        .subscribe { passed, failed ->
            if((passed + failed) > 0 && passed == 0) {
                log.warn("No contigs passed GTDB-TK min. completeness filters. GTDB-TK summary will execute but results will be empty!")
            }
        }
```

Fix the warning here too in a similar fashion to the other PR

jfy133: Love it!

jfy133:
@nf-core-bot fix linting

subworkflows/local/gtdbtk/main.nf
d4straub: Potentially a small change:
"No contigs passed GTDB-TK min. completeness filters. GTDB-TK summary will execute but results will be empty!"
maybe rather
"No bins or contigs passed GTDB-TK min. completeness filters. GTDB-TK summary will execute but results will be empty!"
because primarily bins are forwarded here, right? Just occasionally a large but unbinned contig might be among those sequences that pass.

subworkflows/local/bin_qc/main.nf
```diff
@@ -42,7 +47,7 @@ workflow BIN_QC {
     if (params.checkm_db) {
         ch_checkm_db = file(params.checkm_db, checkIfExists: true)
     }
-    else if (params.binqc_tool == 'checkm') {
+    else if (params.run_checkm) {
```
prototaxites, line 50: No need to change, but do we want to be foresighted and make these inputs to the subworkflow instead of calling params?

subworkflows/local/gtdbtk/main.nf
```diff
@@ -18,36 +18,49 @@ workflow GTDBTK {
...
+        .join(ch_bin_metrics)
+        .map { bin_name, bin, meta, _bin_qc_tool, completeness, contamination -> [bin_name, meta, bin, completeness, contamination] }
+        .groupTuple(by: 0)
+        .branch { _bin_name, meta, bin, completeness, contamination ->
+            passed: (completeness.any { it != -1 } && completeness.any { it >= params.gtdbtk_min_completeness } && contamination.any { it != -1 } && contamination.any { it <= params.gtdbtk_max_contamination })
+            return [meta[0], bin[0]]
+            discarded: true
+            return [meta[0], bin[0]]
         }
+    // Note we have to call `meta[0], bin[0]` because of the groupTuple above
```
prototaxites, line 63: Minor thought: You could use `.join(ch_bin_metrics.groupTuple(by: 0))` instead and then that would avoid the problem?

subworkflows/local/gtdbtk/main.nf
```diff
@@ -61,23 +74,27 @@ workflow GTDBTK {
...
     }
 
-
-    // Print warning why GTDB-TK summary empty if passed channel gets no files
-    ch_filtered_bins.passed
-        .count()
-        .map { it == 0 ? log.warn("No contigs passed GTDB-TK min. completeness filters. GTDB-TK summary will execute but results will be empty!") : "" }
-
-
+    // We set unique to remove duplicate bins when the same bin has passed the filter of two different QC tools
```
prototaxites: Is this a defunct comment now

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -354,18 +354,18 @@ def validateInputParameters(hybrid) {
...
+    if (params.skip_binqc && (params.run_busco || params.run_checkm || params.run_checkm2)) {
+        error("[nf-core/mag] ERROR: Both --skip_binqc and --run_<bin_qc_tool_name> are specified! Invalid combination, please specify either --skip_binqc or --run_<bin_qc_tool_name>.")
     }
 
     // Check if BUSCO parameters combinations are valid
     if (params.skip_binqc) {
-        if (params.busco_db) {
-            error("[nf-core/mag] ERROR: Both --skip_binqc and --busco_db are specified! Invalid combination, please specify either --skip_binqc or --binqc_tool 'busco' with --busco_db.")
+        if (params.run_busco || params.busco_db) {
+            error("[nf-core/mag] ERROR: Both --skip_binqc and --run_busco or --busco_db are specified! Invalid combination, please specify either --skip_binqc or --run_busco with --busco_db.")
```
prototaxites: If `busco_db` is supplied but `run_busco` is false, my expectation would be that the pipeline should be able to run. Makes turning things off a bit easier if BUSCO is acting up.

prototaxites (approved):
Hi all, from a Nextflow standpoint this looks good to me. A few very small comments, none blocking. 

Note that I haven't validated the outputs so feel free to take my review/approval with a pinch of salt.

subworkflows/local/bin_qc/main.nf
jfy133: Good idea - however how the strict syntax is becoming more stricter, my plan was to do this in one big PR of just that everywhere (or severaly PRs, but just the same thing each time)

subworkflows/local/gtdbtk/main.nf
jfy133: I think I tried that before but it didn't do what I expected at the time... but we can refine that again next time we (inevitably 🙃) come back to this subworkflow

subworkflows/local/bin_qc/main.nf
prototaxites: That's fine too - just always looking for ways to save a bit of future effort 😉

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
jfy133: Good point, I think this was left over from the previous 'run if db supplied' system - updated

jfy133: Done!

jfy133:
Thanks @prototaxites !

jfy133:
And thank you @harper357 for kicking this off!

PR #848, opened 2025-08-15
Dev test alternatives

dialvarezs:
## PR checklist


dialvarezs:
@nf-core-bot fix linting

conf/test_alternatives.config
```diff
@@ -0,0 +1,41 @@
...
+params {
+    config_profile_name        = 'Test alternatives profile'
+    config_profile_description = 'Minimal test dataset to check pipeline function'
+
+    // Input data
+    input                         = params.pipelines_testdata_base_path + 'mag/samplesheets/samplesheet.v4.csv'
+	clip_tool                     = 'trimmomatic'
+	binqc_tool                    = 'checkm2'
+	bin_domain_classification     = true
+    skip_spades                   = true
```
jfy133: [suggestion](https://pullpo.io/cc?l=suggestion) 
Odd indentation here?

tests/.nftignore
jfy133, line 1: [question](https://pullpo.io/cc?l=question) 
Why are all the log files being excluded here? We an still check for strings for of theses as most finish on 'ends on XYZ'

Fro DIAMOND I assume we would expect the same number of reads, for example?

dialvarezs: TBH I was a bit lazy here, I only excluded everything not stable on checksums.
I will look at the excluded files and add some specific checks.

dialvarezs:
I added some extra checks for the logs and also for the DIAMOND output from CheckM2.

I also reviewed all the stable files, and their content looks fine.
However, I was thinking we could include nft-fasta and/or nft-fastq to verify if the format is correct and that they contain sequences in a more automated way. What do you think of this, @jfy133?

jfy133:
> I added some extra checks for the logs and also for the DIAMOND output from CheckM2.
> 
> I also reviewed all the stable files, and their content looks fine. However, I was thinking we could include nft-fasta and/or nft-fastq to verify if the format is correct and that they contain sequences in a more automated way. What do you think of this, @jfy133?

Yes that would be great! I think that's a very good idea even if we _may_ find a lot of variability (particularly with sequencer headers), but worth it in the long run!

dialvarezs:
I added FASTA checks, it's ready for a new review now

conf/test_alternatives.config
```diff
@@ -0,0 +1,41 @@
...
+    bin_domain_classification  = true
+    skip_spades                = true
+    skip_quast                 = true
+    skip_prodigal              = true
+    skip_prokka                = true
+    skip_gtdbtk                = true
+    skip_maxbin2               = true
+    skip_concoct               = true
+    skip_metaeuk               = true
+    megahit_fix_cpu_1          = true
```
prototaxites, line 40: Question, is there a reason to set this if we aren't checking for an exact file match on the output of megahit?

tests/test_alternatives.nf.test
```diff
@@ -0,0 +1,117 @@
...
+                        },
+                    ).match('logs')
+                },
+                { assert snapshot(
+                        checkm2_diamond_results.collect { results -> results.text.readLines().size() },
+                    ).match('checkm2_diamond_results')
+                },
+                { assert snapshot(
+                        megahit_contigs.collect { contigs -> path("${contigs}").fasta.size() > 0 },
+                    ).match('contigs')
```
prototaxites: Is it worth actually snapshotting these (and other cases where you check a set of booleans)? You just get a list of e.g. "true" in the .snap file. You could just assert that they're all true instead?

e.g. 

```suggestion
                { assert megahit_contigs.collect { contigs -> path("${contigs}").fasta.size() > 0 }.every() }
```

conf/test_alternatives.config
dialvarezs: Yes, the snapshot includes the checksums of the assembled contigs, and also of the bins. That's why I added the option to have reproducible results.

prototaxites: Ah, I missed that - I just saw the bit I commented on below checking the files are not empty!

tests/test_alternatives.nf.test
dialvarezs: That's a good point indeed. I will make it like that for every truthy check

Issue #874, opened 2025-10-06
Add additional binners


d4straub:
### Description of feature

I am going to add some more binning tools because there seem to be very well performing binners that are not yet in this pipeline.

Currently there are 3 binners in nf-core/mag: MetaBAT2, MaxBin2, CONCOCT. In case several binners are used, bins are refined with DAS Tool.

In the CAMI 2 challenge, binners (among other tools) were benachmarked and the best performing ones were implemented into a fork of nf-core/mag (https://github.com/hzi-bifo/mag). The binners added are: [MetaBinner](https://github.com/ziyewang/MetaBinner) and [MetaWRAP](https://github.com/bxlab/metaWRAP). Additionally, UltraBinner 1 performs well.

Additionally, the benchmarking website of CAMI 2 allows continuous submission of results and evaluation. Since the publication two new binners that perform very well have been listed, see [here, checked 6th Oct 25](https://cami-challenge.org/datasets/Marine/marmgCAMI2_short_read_pooled_gold_standard_assembly/genome_binning/). The binners are COMEBin & COMEBERT-2 / COMBERT-S.

The following is planned:
- binners to look at: MetaBinner, MetaWRAP, UltraBinner, COMEBin, COMEBERT
- (a) check if those (or more) binners have nf-core modules
- (b) check if binners have conda packages, ideally also docker/singularity containers
- check if binners can be added to DAS Tool refinement
- implement binners one by one to nf-core/mag (potentially via nf-core/modules)

Comments/discussion here or in nf-core/slack welcome.

jfy133:
- Another binner that has been bandied around is SemiBin (which I think is quite new)
- There is also https://github.com/nf-core/mag/issues/714 which is 'meant' to replace metaWRAP as bin refiner (and I think this would be nice as an alternative to DAS_Tool, a) because altenratives are good, and b. metaWRAP still performs best for many people, just UX is poor!)

amizeranschi:
Hi everyone,

I've recently come across another binner called kMetaShot, which looks promising at first glance. There's no nf-core module yet, but a bioconda package is available:
https://anaconda.org/bioconda/kmetashot
https://academic.oup.com/bib/article/26/1/bbae680/7941744

Could this be a good candidate for nf-core/mag?

d4straub:
Now that I added COMEBin, MetaBinner and SemiBin2, this task is done for now from my side.

PR #875, opened 2025-10-07
Add binner COMEBin

d4straub:
Attempt to add COMEBin. Follows https://github.com/nf-core/mag/issues/874.

Issues:
- doesnt run with any CI test dataset (but the module does succeed with `-profile test_full` and output looks fine to me)
- runtime seems long: 1-7h per samples, while METABAT2_METABAT2 requires <1min, BINNING:MAXBIN2 a few minutes

As a comment:
`ext.args = "-b 20"` helps with test data, because default value fails with "-profile test" (100 passes that step) as in `https://github.com/ziyewang/COMEBin/issues/4` but instead the binner fails later after: "[CheckM - analyze] Identifying marker genes in bins.", so it doesnt make it work all the way through.

## PR checklist

- [x] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/mag/tree/main/.github/CONTRIBUTING.md)


modules/nf-core/comebin/runcomebin/main.nf
```diff
@@ -0,0 +1,75 @@
...
+    script:
+    def args               = task.ext.args ?: ''
+    prefix                 = task.ext.prefix ?: "${meta.id}"
+    def clean_assembly     = assembly.toString() - ~/\.gz$/
+    def decompress_contigs = assembly.toString() == clean_assembly ? "" : "gunzip -q -f $assembly"
+    def cleanup            = decompress_contigs ? "rm ${clean_assembly}" : ""
+    """
+    ${decompress_contigs}
+
+    cp ${clean_assembly} . # ISSUE: temporary files will be generated in the directory of the assembly file, following links, copying prevents that
```
d4straub: This is because the module attempts to write into the work dir of the source data (assembly), so I modified the nf-core module but the change is inefficient. Any better idea @prototaxites @jfy133 ?

prototaxites: Set the stage in mode to copy?

modules/nf-core/comebin/runcomebin/main.nf
```diff
@@ -0,0 +1,75 @@
...
+        -a ${clean_assembly} \\
+        -p bam/ \\
+        -o . \\
+        $args
+
+    mv comebin_res ${prefix}
+
+    find ${prefix}/comebin_res_bins/*.fa -exec gzip {} \\;
+
+    for filename in ${prefix}/comebin_res_bins/*.fa.gz; do mv "\${filename}" "${prefix}/comebin_res_bins/${prefix}.\$(basename \${filename})"; done; # avoid file name collisions
```
d4straub: An issue occurred without that line in the downstream module saying:
```
Process `NFCORE_MAG:MAG:BINNING:SEQKIT_STATS` input file name collision -- There are multiple input files for each of the following file names: 0.fa.gz
```
and indeed the bins were in files with name `<int>.fa.gz` in folder `${prefix}/comebin_res_bins`. This line renames each file so that its `${prefix}.<int>.fa.gz` which prevents any file name collisions. 
Do you have here any more elegant solutions?

modules/nf-core/comebin/runcomebin/main.nf
d4straub: You mean in line 11 `tuple val(meta), path(assembly, mode: 'copy'), path(bam, stageAs: "bam/*")` ? I dont find that in the nextflow docu but I can try it. 
There is no way to prevent copying the complete assembly (its just such a waste of space! but I dont have an idea right now)?

prototaxites: https://www.nextflow.io/docs/latest/reference/process.html#stageinmode

Ah, it's not an argument to path but a separate process directive

modules/nf-core/comebin/runcomebin/main.nf
prototaxites: You could write a really ugly find command that does it in one go?

Otherwise that looks fine!

modules/nf-core/comebin/runcomebin/main.nf
d4straub: Thanks that looks interesting, never considered that. However, that means _all_ input files will be copied, not only the assembly but also the BAM files. Do we really want that?

prototaxites: Wonder if it's possible to write a closure that only applies it to certain inputs?

d4straub: > Wonder if it's possible to write a closure that only applies it to certain inputs?

Forgive me, but if a closure is needed then this is not making the code better, imho. The cp is plain and simple and works fine as well. I would have preferred to avoid copying altogether though.

d4straub:
@nf-core-bot fix linting

modules/nf-core/comebin/runcomebin/comebin-runcomebin.diff
```diff
@@ -0,0 +1,36 @@
...
+     label 'process_high'
++    label 'process_gpu'
+ 
+     conda "${moduleDir}/environment.yml"
+     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+@@ -30,6 +31,8 @@
+     """
+     ${decompress_contigs}
+ 
++    cp ${clean_assembly} . # ISSUE: temporary files will be generated in the directory of the assembly file, following links, copying prevents that
```
prototaxites: Actually, how about changing this to:

```
def setup contigs = assembly.toString() == clean_assembly ? "cat ${assembly} > local_assembly.fasta" : "zcat ${assembly} > local_assembly.fasta"
script:
"""
${setup_contigs}

<run comebin on local_assembly.fasta>
...

rm local_assembly.fasta
"""
```

So we always just copy to a clean file and then delete it afterwards?

prototaxites:
> added a GPU label but didnt test it.

This won't work as the container environment needs to have GPU-specific versions of PyTorch etc.. Would need a GPU-specific container for this to work.

conf/test_assembly_input.config
```diff
@@ -42,6 +42,7 @@ params {
     skip_prokka                     = true
     skip_gtdbtk                     = true
     skip_concoct                    = false
+    skip_comebin                    = true
```
prototaxites, line 45: Unfortunately COMEBin seems to require data of sufficient complexity to complete without failure... Hence why the module has a stub test only

conf/modules.config
```diff
@@ -806,6 +806,22 @@ process {
         ext.prefix = { "${meta.assembler}-CONCOCT-${meta.id}" }
     }
 
+    withName: COMEBIN_RUNCOMEBIN {
```
prototaxites, line 809: We should probably set errorStrategy = "ignore" on this one?

modules/nf-core/comebin/runcomebin/comebin-runcomebin.diff
d4straub: Ah that sound good, will investigate.

conf/modules.config
d4straub: Will make the pipeline more robust in execution, yes, I like it.

d4straub:
@nf-core-bot fix linting

modules/nf-core/comebin/runcomebin/comebin-runcomebin.diff
d4straub: Seems to work as intended, run with that change is still running though.

prototaxites: Excellent! This PR looks great now - but it would be worth backporting these changes to the Comebin nf-core module before merging, I think?

d4straub: A task for tomorrow :D

d4straub:
Done: 
- final confirmation that -profile test_full works fine as is -> fine
- port back changes of the COMEBin module to nf-core/modules -> https://github.com/nf-core/modules/pull/9202
- re-install module here -> done

modules/nf-core/comebin/runcomebin/comebin-runcomebin.diff
d4straub: Done: Changes backported to Comebin nf-core module & re-installed here.

d4straub:
Thanks! Will open an issue!

PR #894, opened 2025-10-25
Fix POOL_SHORT_READS receiving reads in wrong roder resulting in faulty pooling

### LINKED ISSUE #890: METASPADES exit status 21: paired read files contain unequal number of reads
### Description of the bug

I'm using mag 5.0.0  for the first time after successfully running it with earlier versions and I mainly wanted to make use of the `--fastp_trim_polyg` option to see if it could improve earlier co-assemblies of my Illumina NovaSeq paired-end data. Everything went well until the assembling step as I kept on getting:
```
[de/077bc1] NOTE: Process `NFCORE_MAG:MAG:ASSEMBLY:SHORTREAD_ASSEMBLY:METASPADES (group-2)` terminated with an error exit status (21) -- Execution is retried (1)
```
I looked in the `.command.log` and `.spades.log` and found:
```
13:41:50.984   129G / 172G  ERROR   General                 (hammer_tools.cpp          : 194)   Pair of read files "/.../nf-core_mag/mag_5.0.0_Run1_22_10_2025/_scratch_/de/077bc1c4526da920b3f45c29a9f3dd/group-2_1.merged.fastq.gz" and "/.../nf-core_mag/mag_5.0.0_Run1_22_10_2025/_scratch_/de/077bc1c4526da920b3f45c29a9f3dd/group-2_2.merged.fastq.gz" contain unequal amount of reads
``` 
When inspecting the POOL_SHORT_READS step I noticed a big mismatch in file size of the merged fastq files and inspected those further with  seqkit stats to confirm the issue
``` 
seqkit stats group-2_1.merged.fastq.gz group-2_2.merged.fastq.gz 
processed files:  2 / 2 [======================================] ETA: 0s. done
file                       format  type     num_seqs         sum_len  min_len  avg_len  max_len
group-2_1.merged.fastq.gz  FASTQ   DNA   119,559,197  17,878,497,150       15    149.5      151
group-2_2.merged.fastq.gz  FASTQ   DNA   101,589,939  15,161,387,675       15    149.2      151
``` 
This issue is true for all my merged fastq files so I'm unable to run SPADES at the moment. I do not think this is supposed to happen and would like to know if you know a fix for this issue as I would assume paired end date should drop the read if the pair is missing  or was removed in trimming step?  Thank you for your help!

### Command used and terminal output

```console
# Nextflow command
nextflow run nf-core/mag \
    -resume \
    -r 5.0.0 \
    -work-dir "_scratch_" \
    -profile singularity \
    -process.executor "slurm" \
    -process.queueSize 10 \
    -process.maxForks 10 \
    -c "custom.config" \
    --input "samplesheet.csv" \
    --outdir "./results" \
    --spades_fix_cpus 32 \
    --fastp_trim_polyg \
    --host_fasta "./references/genomes/homo_sapiens/refdata-gex-GRCh38-2024-A/fasta/genome.fa" \
    --host_removal_verysensitive \
    --coassemble_group \
    --spades_downstreaminput "contigs" \
    --skip_megahit \
    --prokka_compliance_centre EBI \
    --prokka_with_compliance \
    --skip_concoct \
    --min_contig_size 1500 \
    --save_assembly_mapped_reads \
    --busco_db_lineage "auto_prok" \
    --save_busco_db \
    --busco_clean \
    --gtdb_db ./gtdbtk_r226_data.tar.gz \
```

### Relevant files

_No response_

### System information

_No response_

jfy133:
Closes #890 

- Essentially the input tuple for pooling was in the wrong format meaning only R1s were being pooled and out of order (i.e, what was meant to be a samples R2, was the second samples R1)
- This was missed as the `--coassembly_group` parameter was missed out in the new config structures

TODO:

- Run tests for all other configs to make sure nothing else changed
- Regenerate snapshot for test_alternative now coassembly activated

## PR checklist

### DIFF
--- a/subworkflows/local/assembly/main.nf
+++ b/subworkflows/local/assembly/main.nf
@@ -48,7 +48,7 @@ workflow ASSEMBLY {
 
         // We have to merge reads together to match tuple structure of POOL_SHORT_READS/
         // This MUST be in a interleaved structure (s1_r1, s1_r2, s2_r1, s2_r2, ...)
-        // So we merge the two list of R1 and R2s, and sort them to ensure correct order above
+        // So we zip the lists of R1s and R2s, consistent order is ensured by the existing structure
         ch_short_reads_grouped_for_pooling = ch_short_reads_grouped.map { meta, reads1, reads2 ->
             [meta, [reads1, reads2].transpose().flatten()]
         }


jfy133:
@nf-core-bot fix linting

subworkflows/local/assembly/main.nf
```diff
@@ -44,6 +44,12 @@ workflow ASSEMBLY {
                     [meta, reads.collect { it[0] }, reads.collect { it[1] }]
                 }
             }
+
+        // We have to merge reads together to match tuple structure of POOL_SHORT_READS/
+        // This MUST be in a interleaved structure (s1_r1, s1_r2, s2_r1, s2_r2, ...)
+        // So we merge the two list of R1 and R2s, and sort them to ensure correct order above
+        ch_short_reads_grouped_for_pooling = ch_short_reads_grouped.map { meta, reads1, reads2 -> [meta, [reads1 + reads2].flatten().sort()] }
```
prototaxites: Can we assume that the reads files here are standardly-named such that a `sort()` won't break the order?

subworkflows/local/assembly/main.nf
d4straub, line 79: I think thats actually incomplete: when no long reads are used, --skip_spades alone should suffice to not concatenate reads for spades/spadeshybrid?
I assume POOL_SHORT_READS is not required for MEGAHIT.

subworkflows/local/assembly/main.nf
d4straub: Yes, I think we can assume that because all those files are renamed for ${prefix} at that point, imho. Or is it possible to skip the complete QC so that original files names come through? Not entirely sure...

prototaxites: I do think it's possible to basically completely skip QC...

jfy133: How likely do you think it would be that people don't have a `_R1` / `_R2`, `_1` / `_2`, `_F`, `_R` in their FASTQ files?

subworkflows/local/assembly/main.nf
jfy133: Right, but we can't do that in this condition - the presence of long reads depends on the contents of the channels? Unless I don't understand?


Otherwise correct, MEGAHIT can accept each fastq file spearately

jfy133: Feel free to make a direct suggestion to illustrate what you mean :)

subworkflows/local/assembly/main.nf
prototaxites: I would be wary of assuming anything about file names unless we have strictly controlled it. One way to do that would be also to force a schema like the above in the samplesheet validation, so we stop early before errors. 

Otherwise we have to be careful with channel order, etc.?

jfy133: Conceptually I agree, but this would require us to force people to potentially rename all of their files (which will put people off) - and I'm not sure how easy it would be define such a regex in that the position of `_1` `_2` in the file name can very, and also you could theoretically have multiple e.g. `_1` in a file name if it's a part of a sample name.

Furthermore, the reason why I'm having to do this faff anyway is because pretty much all nf-core modules that handle pairs make this assumption anyway (i.e., pairs will have exactly the same file name except the different `_1` `_2` patterns above...so if we want to go and truly fix this we would need to go and re-do all the modules 😅

This isn't a satisfying answer I know but my feeling is it's a common enough pattern that the edge cases someone doesn't follow that will be very few and not worth time/value effort to make more robust vs the level of brokenness coassembly in mag is in...

subworkflows/local/assembly/main.nf
d4straub: Its actually out of scope of that PR I guess. The co-assembly should be fixed asap.

subworkflows/local/assembly/main.nf
d4straub: I agree (with both). Practically that implementation now will only very seldom fail I think. If all QC is skipped, one could potentially rename files with $prefix in a separate process.
But for me that isnt needed, potentially we could add a comment (Warning) in the docs about the sorting issue with SPAdes & skipping all QC & file names, maybe to the co-assembly step (https://nf-co.re/mag/5.1.0/docs/usage/#the-group-column?), or to param `skip_shortread_qc`?

prototaxites: I'm trying to re-familiarise myself with this bit of code, but am struggling to understand at a quick glance exactly what is going wrong here. But I would prefer to either be really explicit about the inputs (and annoy a handful of users), or figure out a channel-order based solution that doesn't require a sort, just to avoid errors sneaking into peoples' analyses that they wouldn't have caught otherwise.

> Conceptually I agree, but this would require us to force people to potentially rename all of their files (which will put people off) - and I'm not sure how easy it would be define such a regex in that the position of _1 _2 in the file name can very, and also you could theoretically have multiple e.g. _1 in a file name if it's a part of a sample name.

Something like the following in the samplesheet schema would work. 

short_reads_1: `"pattern": "^\\S+(_R1|_[1F])\\.f(ast)?q\\.gz$"`
short_reads_1: `"pattern": "^\\S+(_R2|_[2R])\\.f(ast)?q\\.gz$"`

d4straub: Typical Illumina output from the sequencing facilities & companies I know is `<sample>_R1_<lane>.fastq.gz`. Single-end read files might not have any of those pattern to identify direction (R1/1F/whatever).
I think that makes it already more complicated to catch? I am not an regex expert though.

jfy133: @d4straub 's pattern the one of the most common patterns I've seen too... and other people append the adapter index sequence to the end after the lane ID too... so I really don't think this will be simply be solvable.

> But for me that isnt needed, potentially we could add a comment (Warning) in the docs about the sorting issue with SPAdes & skipping all QC & file names, maybe to the co-assembly step (https://nf-co.re/mag/5.1.0/docs/usage/#the-group-column?),

I'm erring for this, but I want this to be a democracy.

@dialvarezs any thoughts?

prototaxites: Deleted my previous comment, as I wasn't sure it actually worked, but I think it does:

```
ch_a = Channel.of(["meta", ["a", "c", "b"], ["d", "b", "f"]])
ch_a.map { meta, f1, f2 ->
    def transposed_pairs = [f1, f2].transpose()
    println transposed_pairs
    
    def sorted_pairs = transposed_pairs.sort { it[0] }  
    println sorted_pairs
      
    def interleaved = sorted_pairs.flatten()

    return [meta, interleaved] 
}.view()
transposed: [[a, d], [c, b], [b, f]]
sorted: [[a, d], [b, f], [c, b]]

output: [meta, [a, d, b, f, c, b]]
```

So we can just sort on fasta1's name, avoiding issues with naming entirely.

prototaxites:
```suggestion
        ch_short_reads_grouped_for_pooling = ch_short_reads_grouped.map { meta, reads1, reads2 -> [meta, [reads1, reads2].transpose().sort { it[0].getName() }.flatten()] }
```

d4straub: I tried that code above and it seems fine to me. It also works with e.g.
`ch_a = Channel.of(["meta", ["a_s1_R1_a", "c_s3_R1_c", "b_s2_R1_b"], ["d_s1_R2_d", "b_s3_R2_b", "f_s2_R2_f"]])`
that is sorted to
`[meta, [a_s1_R1_a, d_s1_R2_d, b_s2_R1_b, f_s2_R2_f, c_s3_R1_c, b_s3_R2_b]]`

jfy133: OK nice thanks for the cross-validation @d4straub ! When I am more functional I will try to implement it!

dialvarezs:
This was almost ready TBH. I just implemented the suggestions from @prototaxites and @d4straub in the discussion and verified that everything works as expected. The only change I made was moving the sorting to the grouping step, so the files start sorted from there. Then, as `.transpose()` keeps the order, just flattening the result is enough to get the interleaved reads for pooling.

conf/test_alternatives.config
```diff
@@ -27,6 +27,7 @@ params {
     // Input data
     input                      = params.pipelines_testdata_base_path + 'mag/samplesheets/samplesheet.v4.csv'
     clip_tool                  = 'trimmomatic'
+    coassemble_group           = true
```
d4straub, line 30: Is that addition and the consequently changing files in the snap file intended?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Fixed`
 
+- [#893](https://github.com/nf-core/mag/pull/893) - Fix R1 being mistakenly provided to POOL_SHORT_READS instead of R2 for paired end data (reported by @maartenciers, fix by @jfy133 with contributions from @prototaxites, @d4straub and @dialvarezs)
```
d4straub:
```suggestion
- [#894](https://github.com/nf-core/mag/pull/894) - Fix read order in metaSPAdes to allow co-assembly of paired-end data of multiple samples (reported by @maartenciers, fix by @jfy133 with contributions from @prototaxites, @d4straub and @dialvarezs)
```

subworkflows/local/assembly/main.nf
```diff
@@ -32,18 +32,27 @@ workflow ASSEMBLY {
...
                 }
                 else {
-                    [meta, reads.collect { it[0] }, reads.collect { it[1] }]
+                    [meta] + reads.sort { files -> files[0].getName() }.transpose()
                 }
             }
+
+        // We have to merge reads together to match tuple structure of POOL_SHORT_READS/
+        // This MUST be in a interleaved structure (s1_r1, s1_r2, s2_r1, s2_r2, ...)
+        // So we merge the two list of R1 and R2s, and sort them to ensure correct order above
```
d4straub: but that sorting isnt done here now?

conf/test_alternatives.config
jfy133: Yes, this this option was meant to be set to `true` when we first made this profile, but was left accidently out

subworkflows/local/assembly/main.nf
```diff
@@ -32,18 +32,27 @@ workflow ASSEMBLY {
...
+                    single_end: assemble_as_single,
+                    sr_platform: metas.sr_platform[0]
+                ]
                 if (assemble_as_single) {
-                    [meta, reads.collect { it }, []]
+                    [meta, reads.collect { file -> file }, []]
                 }
                 else {
-                    [meta, reads.collect { it[0] }, reads.collect { it[1] }]
+                    [meta] + reads.sort { files -> files[0].getName() }.transpose()
```
jfy133, line 45: I'm not fully understanding this line, does the `.transpose()` result in a `meta, reads1, reads2` tuple here?

dialvarezs: Yes, after the `groupTuple(0)` we have this:
```
[meta, [[l1r1, l1r2], [l2r1, l2r2], ...]
````
And that line is basically
```
[meta] + [[l1r1, l2r1, ...], [l2r1, l2r2, ...]]
```
So the final structure is
```
[meta, [l1r1, l2r1, ...], [l2r1, l2r2, ...]]
```

It's the reverse operation we're doing below for creating the interleaved reads (in which we need to go back to step 1).

subworkflows/local/assembly/main.nf
dialvarezs: Sure, I will update that line.
In fact, the sorting step isn’t required for this to work, because since we preserve the tuple order, the r1 and r2 pairs will be in the same order. I added it anyway for the sake of result stability (to get snapshots working, I have been adding sorting steps basically everywhere).

subworkflows/local/assembly/main.nf
jfy133: OK sounds good!

jfy133:
I'm ok with this, but can't approve my own PR @d4straub could you give a :heavy_check_mark: if you're happy, and ofc @prototaxites !

jfy133:
Let's go! @dialvarezs thanks for wrapping this up :pray: 

I suggest we wait one more week to see if any more last minute bug fixes come up, otherwise let's get this out in a release!

dialvarezs:
> I suggest we wait one more week to see if any more last minute bug fixes come up, otherwise let's get this out in a release!

So we can continue with our flood of bi-weekly releases 😂

jfy133:
> > I suggest we wait one more week to see if any more last minute bug fixes come up, otherwise let's get this out in a release!
> 
> So we can continue with our flood of bi-weekly releases 😂


Exactly

PR #905, opened 2025-10-28
Snapshot for `-profile test_assembly_input`

dialvarezs:
## PR checklist

### DIFF
--- a/tests/test_alternatives.nf.test
+++ b/tests/test_alternatives.nf.test
@@ -10,7 +10,7 @@ nextflow_pipeline {
 
         when {
             params {
-                outdir = "$outputDir"
+                outdir = "${outputDir}"
             }
         }
 
--- a/tests/test_assembly_input.nf.test
+++ b/tests/test_assembly_input.nf.test
@@ -10,7 +10,7 @@ nextflow_pipeline {
 
         when {
             params {
-                outdir = "$outputDir"
+                outdir = "${outputDir}"
             }
         }
 
@@ -105,6 +105,7 @@ nextflow_pipeline {
 
                 // CSV checks
                 { assert path(bin_summary).csv(sep: "\t").rowCount > 0 },
+                { assert path(bin_summary).csv(sep: "\t").rowCount == dastool_bins.size() },
                 { assert path(busco_summary).csv(sep: "\t").rowCount > 0 },
                 { assert snapshot(
                     path(bin_summary).csv(sep: "\t").rowCount,


dialvarezs:
@nf-core-bot fix linting

dialvarezs:
Ok, after a long time struggling this the stability of this  snapshot, it's finally ready for merging.

The instability because of:
- An issue of the publishing of BUSCO files, it's fixed now
- Inconsistent sorting on CSV files. I added a sort step in the Python scripts to keep `bin_summary.tsv` stable.

Regarding BUSCO, I found out that is not able to produce consistently sorted CSV  batch files, even when running with a single core... So I decided to left out of the snapshots both `*.batch_summary.txt` and `busco_summary.tsv`.

dialvarezs:
I also updated the `get_mag_depths.py` script to produce depth summaries with consistent column ordering, to prevent swapped depth values on the merged depth summary.

conf/modules.config
```diff
@@ -463,12 +463,13 @@ process {
     }
 
     withName: BUSCO_BUSCO {
+        tag        = { "${meta.assembler}-${meta.binner}-${meta.domain}-${meta.refinement}-${meta.id}" }
```
d4straub: this tag might be overly long? At least to me, on the console outut that wouldnt be helpful. In the log file might be good?

conf/modules.config
```diff
@@ -463,12 +463,13 @@ process {
...
         ext.args   = [
             params.busco_db ? '--offline' : ''
         ].join(' ').trim()
         publishDir = [
             [
-                path: { "${params.outdir}/GenomeBinning/QC/BUSCO/${meta.id}" },
+                path: { "${params.outdir}/GenomeBinning/QC/BUSCO/${meta.assembler}-${meta.binner}-${meta.domain}-${meta.refinement}-${meta.id}" },
```
d4straub: Will that need an update of the output documentation?

conf/test_assembly_input.config
```diff
@@ -22,8 +22,16 @@ process {
 
 process {
     withName: METAEUK_EASYPREDICT {
+        cpus = 1
```
d4straub: multithreading makes the process non-reproducible? If thats true, then it might be good to mention that in the docs at least? Probably in section https://nf-co.re/mag/5.1.0/docs/usage/#reproducibility ?
Same question for BUSCO below.

conf/test_assembly_input.config
```diff
@@ -22,8 +22,16 @@ process {
...
         ext.args = '-s 1 --max-seqs 10 --max-accept 1'
     }
+    withName: BUSCO_BUSCO {
+        cpus = 1
+        ext.args = '--offline --opt-out-run-stats'
+    }
+    withName: CONCAT_BINQC_TSV {
+        cpus = 1
```
d4straub: Why? Could that be always on 1 cpu? Thats not a cpu-hungry process, right?

tests/test_assembly_input.nf.test
```diff
@@ -15,6 +15,63 @@ nextflow_pipeline {
...
+            )
+            def stable_path_annotation = getAllFilesFromDir(
+                params.outdir, include: ['Annotation/**'], ignoreFile: 'tests/.nftignore'
+            )
+
+            // Output directory: GenomeBinning
+            def stable_name_binning = getAllFilesFromDir(
+                params.outdir,
+                include: ['GenomeBinning/**'],
+                ignore: ['GenomeBinning/QC/BUSCO/**/{logs,prodigal_output,busco_sequences}/**'],
```
d4straub, line 38: doesnt that overlap with .nftignore? Wouldnt it be better to use that file instead?

conf/test_assembly_input.config
dialvarezs: This is probably not needed.
In my desperation to get stable results I started to put `cpus = 1` on everything that was failing the snapshot.
I will remove and see what happens.

tests/test_assembly_input.nf.test
dialvarezs: Not really, because right now we are using `.nftignore` to exclude files unstable on content but stable on filename. This ignore exclude those BUSCO files from the  filename checking too. If we use the nftignore here we will exclude all the files listed on `.ntfignore` from filename checking, and we don't want that.

d4straub: Ah I see, sorry for my misunderstanding.

conf/test_assembly_input.config
dialvarezs: The  results seem to be the same regardless of the number of CPUs, but the ordering not. So, it's neccessary to keep the ability to snapshot those files only. I added a comment.

dialvarezs:
I added some basic CSV checks for `busco_summary` and `bin_summary` files. I also updated the nft-utils plugin.
And fixed the error reporting of tiara_classify that was breaking the `test_alternatives` test.

Most of the small fixes I did on this PR originally were ported to other PR for v5.2.0, so this PR has become more simple now.

tests/test_alternatives.nf.test
```diff
@@ -72,13 +72,12 @@ nextflow_pipeline {
...
 
+            assert workflow.success
             assertAll(
-                { assert workflow.success },
                 { assert snapshot(
-                        workflow.trace.succeeded().size(),
-                        removeNextflowVersion("${outputDir}/pipeline_info/nf_core_mag_software_mqc_versions.yml"),
-                    ).match()
-                },
+                    workflow.trace.succeeded().size(),
```
prototaxites, line 78: There was something about this being removed from the template due to incompatibilities with the latest Nextflow versions... is that resolved?

tests/test_assembly_input.nf.test
```diff
@@ -27,12 +86,48 @@ nextflow_pipeline {
...
+                { assert concoct_bins.collect { file -> path("${file}").fasta.size() > 0 }.every() },
+                { assert maxbin2_bins.collect { file -> path("${file}").fasta.size() > 0 }.every() },
+                { assert maxbin2_unbins.collect { file -> path("${file}").fasta.size() >= 0 }.every() },
+                { assert metabat2_bins.collect { file -> path("${file}").fasta.size() > 0 }.every() },
+                { assert metabat2_unbins.collect { file -> path("${file}").fasta.size() >= 0 }.every() },
+                { assert dastool_bins.collect { file -> path("${file}").fasta.size() > 0 }.every() },
+                { assert dastool_unbins.collect { file -> path("${file}").fasta.size() >= 0 }.every() },
+
+                // CSV checks
+                { assert path(bin_summary).csv(sep: "\t").rowCount > 0 },
```
prototaxites, line 107: Can you assert that this is equal to concoct_bins + maxbin2_bins + ... etc?

tests/test_assembly_input.nf.test
```diff
@@ -27,12 +86,48 @@ nextflow_pipeline {
...
+                    }.every()
+                },
+                { assert concoct_logs.collect {
+                        log -> log.readLines().last().contains('CONCOCT Finished')
+                    }.every()
+                },
+                { assert busco_logs.collect {
+                        log -> log.readLines().any { it.contains('BUSCO analysis done.') }
+                    }.every()
+                },
```
prototaxites, line 127: Does BUSCO ever fail on any of the bins in the tests?

prototaxites (approved):
A few questions/comments, but this is looking fine as-is!

tests/test_alternatives.nf.test
dialvarezs: Yes, it works with both 25.04 and 25.10. It seems that the issue was with a pre-release version (25.07)

tests/test_assembly_input.nf.test
dialvarezs: In this case we're only using the refined bins so it would be = to dastool bins. I'll add that.

tests/test_assembly_input.nf.test
dialvarezs: IIRC it doesn't fail in any of the tests, but as we keep adding the snapshots, if we find a fail we can update this rule.

PR #907, opened 2025-10-31
include COMEBin in dastool/bins output folder

AlexHoratio:
I just ran this pipeline, and there weren't any COMEBin bins in the `DASTool/bins` folder! All the others were there, and the .tsv summary files all listed the COMEBin bins, so DASTool definitely processed them. 

I'm pretty sure it's just because they didn't get included with this pattern, so weren't published. They were all present in the relevant work directory, just not the output folder.

## PR checklist


conf/modules.config
```diff
@@ -873,7 +873,7 @@ process {
             [
                 path: { "${params.outdir}/GenomeBinning/DASTool/bins" },
                 mode: params.publish_dir_mode,
-                pattern: '*-{MetaBAT2,MaxBin2,CONCOCT}Refined-*.fa',
+                pattern: '*-{MetaBAT2,MaxBin2,CONCOCT,COMEBin}Refined-*.fa',
```
prototaxites: Do we need a glob at all here? Can we just have a `*`, as we'll be adding new binners in future?

prototaxites (approved):

prototaxites:
Thank you @AlexHoratio!

PR #912, opened 2025-11-04
Fix multiple sequencing platform validation

dialvarezs:
Fix for an issue [reported on Slack](https://nfcore.slack.com/archives/CE9MS66BS/p1762241530937409).
The source of the issue is that `.collect()` is returning a string, not a list, so `.size()` will give > 1 always.
I tried adding `flat: false` but didn't work either, so I used `.toList()`.

## PR checklist

### DIFF
--- a/tests/.nftignore
+++ b/tests/.nftignore
@@ -18,6 +18,7 @@ GenomeBinning/depths/bins/*.png
 GenomeBinning/MetaBAT2/unbinned/discarded/*.unbinned.pooled.fa.gz
 GenomeBinning/QC/CheckM2/**/DIAMOND_RESULTS.tsv
 GenomeBinning/QC/CheckM2/*/checkm2.log
+GenomeBinning/QC/busco_summary.tsv
 GenomeBinning/QC/BUSCO/**/*.log
 GenomeBinning/QC/BUSCO/**/.checkpoint
 GenomeBinning/QC/BUSCO/**/*.json
--- a/tests/test_alternatives.nf.test.snap
+++ b/tests/test_alternatives.nf.test.snap
@@ -250,7 +250,6 @@
                 "MEGAHIT-MetaBAT2-test_minigut.2.faa:md5,577adc1aa1ebd51fed46f726ad90d9b3",
                 "quality_report.tsv:md5,9365828b072487af69333ecb35353255",
                 "MEGAHIT-MetaBAT2-prokarya-unrefined-test_minigut_checkm2_report.tsv:md5,9365828b072487af69333ecb35353255",
-                "busco_summary.tsv:md5,f2108b2ad319f7bdbe1d42c185e34e43",
                 "checkm2_summary.tsv:md5,9365828b072487af69333ecb35353255",
                 "tiara_summary.tsv:md5,6f97d6016921205c7913e54850b52dad",
                 "bin_summary.tsv:md5,654c1d572f0709145cf87b5480007555",


jfy133 (approved):
Looks good, thanks for the rapid fix @dialvarezs !

jfy133:
Looks like `quay.io` is having problems...

dialvarezs:
Not now BUSCO, not now! >:(

dialvarezs:
@jfy133 I excluded `bin_summary.tsv` from the snapshot, because as I related in #905 it seems impossible to keep it stable, unless  we add a sorting step aftewards.

If that's ok, I'll merge.

jfy133:
> @jfy133 I excluded `bin_summary.tsv` from the snapshot, because as I related in #905 it seems impossible to keep it stable, unless we add a sorting step aftewards.
> 
> If that's ok, I'll merge.

I'm not following (sorry if it's illness-brain), What exactly is not stable? The md5sum? 

I would not exclude entirely as it's a critical file. At a minimum check number of rows and maybe a few strings (you can have multiple assertions for the same file).

But is the problem that `bin_summary.csv` has rows in different order? If so I think your `.sort()` fix for the `mag_depths` script was a good one.

dialvarezs:
@jfy133 My bad here, I did this in a hurry and for some reason I thought it was `busco_summary` instead of `bin_summary`.
And yes, the problem is the row ordering, but in the bin summary is solvable, it just need a sort in the Python script.
I'm reverting this now and aplying the proper fix.

tests/.nftignore
```diff
@@ -18,6 +18,7 @@ GenomeBinning/depths/bins/*.png
 GenomeBinning/MetaBAT2/unbinned/discarded/*.unbinned.pooled.fa.gz
 GenomeBinning/QC/CheckM2/**/DIAMOND_RESULTS.tsv
 GenomeBinning/QC/CheckM2/*/checkm2.log
+GenomeBinning/QC/busco_summary.tsv
```
jfy133, line 21: I don't find this satisfying, can we check for the number of rows instead? But I'll give a preemptive approval on case that's not possible

dialvarezs: Absolutely, I agree 100%.

But I guess we can address it later in a snapshot related PR? I can add it to my PR for `test_assembly_input` and copy it from all the snapshots using BUSCO.

The information of that file es contained on `bin_summary.tsv`, so we're checking it someway. The problem is that BUSCO in batch mode definitively can't produce stable results.

prototaxites: >can we check for the number of rows instead

A quick heads up from my own metagenome pipeline - I have found that the binning tools can output varying numbers of bins between runs. So you need to be a bit clever and can't just hard-code a number

dialvarezs: > A quick heads up from my own metagenome pipeline - I have found that the binning tools can output varying numbers of bins between runs. So you need to be a bit clever and can't just hard-code a number

Our binning results seem stable (maybe because the dataset small and simple enough). We have snapshots for all the bin FASTA files, and the number is consistent too. So I guess it's not so bad to have the number harcoded until we change the dataset. But if you have a better idea @prototaxites, we can implement it that way.

prototaxites: Not specifically (I'm just comparing counts across files) - I just point it out because in my experience MetaBat2 was the big culprit and did not consistently put out the same files or in the same order...

But if it's working here that's fine!

PR #930, opened 2025-11-14
Add binner SemiBin2

d4straub:
Adds SemiBin2, related to https://github.com/nf-core/mag/issues/874

SemiBin2 handles cases where it finds no bins gracefully: exit code 0. It does error though when there are too less contigs (but I could not find how many are required). Therefore errors dont need to be ignored, I think.

Whenever long reads are used in the assembly (if `meta.lr_platform`) then long read binning mode is activated. However I am not sure that is supposed to be used for SPAdes assemblies, because they are (at least last time I checked) short read assembles with long read stitching. But I follow the FAQ `What should I do if I have hybrid data (short- and long-reads)? From SemiBin's point-of-view, you should generally treat this using the long-reads pipeline (--sequencing-type=long_read).` from [here](https://semibin.readthedocs.io/en/latest/faq/).

#### About CI tests:
- 8/8 succeed: 3 with SemiBin2 activated; see post below
- `test_alternatives` should work as well and could be activated (requires snap update though)
- `longreadonly` & `longreadonly_alternatives` produce assemblies with only 3 contigs, SemiBin2 complains; couldnt find any required minimum contig number though.

#### Still to do:
- confirmation for successful `-profile test_full` pending (test running atm)

## PR checklist


d4straub:
@nf-core-bot fix linting

d4straub:
#### Investigated CI test outcomes:

- [x]  1 '-profile test' -> apparently no new bin files (locally SemiBin does produce bins), summary files changed -> snap mismatch causes fail
- [x] [**skipped**] 2 '-profile test_alternatives' has warning:
```
WARN: Input tuple does not match tuple declaration in process `NFCORE_MAG:MAG:DOMAIN_CLASSIFICATION:TIARA:TIARA_SUMMARY` -- offending value: [[:], [/home/runner/_work/mag/mag/~/tests/c090c86a5a5a2ac52952ba3cc887e719/work/e0/bf0d5166fd4ce157c46c583edeb3c7/MEGAHIT-SemiBin2-bins-test_minigut.binclassification.tsv], [:], [/home/runner/_work/mag/mag/~/tests/c090c86a5a5a2ac52952ba3cc887e719/work/d4/b3e898cef3fa26b001a9a0d38099dc/MEGAHIT-MetaBAT2-bins-test_minigut.binclassification.tsv]]
```
    -> TIARA was not executed
    -> new bin files, summary files changed -> snap mismatch causes fail
    -> locally I can confirm the warning, but the pipeline executes all expected processes
- [x] 3 '-profile assembly_input' new bin files, summary files not changed -> snap mismatch causes fail
- [x] 4 '-profile hybrid' no new bin files! but process was executed and summary files changed -> snap mismatch causes fail
- [x] [**skipped**: assembly with too less contigs] 5 '-profile longreadonly' fails with:
```
Process `NFCORE_MAG:MAG:BINNING:SEMIBIN_SINGLEEASYBIN (minigut)` terminated with an error exit status (1)
2025-11-14 14:46:41 99ff53988499 SemiBin2[32] ERROR There are 3 contigs in input file FLYE-minigut.assembly.fasta, but only 3 contain(s) at least 1500 basepairs.
```
- [x] [**skipped**: assembly with too less contigs] 6 '-profile longreadonly_alternatives' fails as above
- [x] [**skipped**: because minimal test] 7 '-profile test_minimal'
- [x] [**skipped**: because too short contigs] 8 '-profile test_single_end' fails with:
```
Process `NFCORE_MAG:MAG:BINNING:SEMIBIN_SINGLEEASYBIN (test_minigut_sample2)` terminated with an error exit status (1)
2025-11-14 14:44:43 bf456a6a163f SemiBin2[32] ERROR There are 215 contigs in input file MEGAHIT-test_minigut_sample2.fa, but only 1 contain(s) at least 1500 basepairs.
```
    --> skip that because data isnt appropriate.

Conclusion: Most tests might be fine, but not long reads only tests. I'll see what I can do to make them pass.

nextflow_schema.json
```diff
@@ -788,6 +788,36 @@
...
                 },
+                "skip_semibin": {
+                    "type": "boolean",
+                    "description": "Skip SemiBin2 Binning"
+                },
+                "semibin_rng_seed": {
+                    "type": "integer",
+                    "default": 1,
+                    "description": "RNG seed for SemiBin2.",
+                    "help_text": "SemiBin2 is run by default with a fixed seed within this pipeline, thus producing reproducible results. You can set it also to any other positive integer to ensure reproducibility. Set the parameter to 0 to use a random seed."
```
dialvarezs, line 799: I don't like that `0` is the value for a random seed, because 0 is also a integer. Why not `null` instead?

conf/modules.config
```diff
@@ -871,6 +871,31 @@ process {
...
+                mode: params.publish_dir_mode,
+                pattern: '*.log',
+            ],
+        ]
+        ext.prefix = { "${meta.assembler}-SemiBin2-${meta.id}" }
+        ext.args2  = {[
+                params.semibin_rng_seed == 0 ? "" : "--random-seed ${params.semibin_rng_seed}",
+                "--min-len ${params.min_contig_size}",
+                "--compression gz",
+                "--engine cpu",
```
dialvarezs, line 892: Could we add a param for this?  Seems to be a common thing to change of you have GPUs available.

nextflow_schema.json
d4straub: Valid point. I literally just copied that from `metabat_rng_seed`. So to keep it consistent for now I would leave it as is if you dont feel strongly about changing it.

conf/modules.config
d4straub: When I added binner COMEBin I was made aware that specific software in the container is required to run it on GPUs, see [here](https://github.com/nf-core/mag/pull/875#issuecomment-3406984049). I dont know myself, but I was assuming the same might be true for the SemiBin2 container. Until now I dont use GPUs, so I am not sure how to test it and make sure it indeed runs on the GPU as intended.
So essentially, do you think here the container would be equipped to execute it as is on a GPU? If yes, then I agree to add that as a param.

dialvarezs: That's a good point. I was assuming that the container had already the required runtime to run on a GPU. I guess we can skip this for now, an leave it as a possible future improvement (for both COMEBin and SemiBin).

prototaxites: I think for these, we would need GPU and non-GPU versions of containers with the correct pytorch runtimes. I don't know if there's an "nf-core" way to do that yet?

nextflow_schema.json
prototaxites: Oh, that explains why you see static metabat outputs in the pipeline snapshots and I don't, @dialvarezs... 🤦‍♂️The MetaBat config doesn't even do this 0 = unset thing - it's just fixed to 0 by default.

IMO in both cases the seed should not be set unless specified by the user.

conf/modules.config
d4straub: There are many nf-core modules with GPU acceleration. [Here](https://github.com/nf-core/tools/pull/3272) is a PR that adds gpu labels into the template. And [here](https://nf-co.re/scdownstream/dev/docs/usage/#gpu-acceleration) a docu from a nf-core pipeline about gpu usage. Didnt find anything in the nf-core docu though, maybe missed it.
In any case, that seems out of scope for this particular PR?

dialvarezs: @d4straub yes, we can leave it as it is right now.

nextflow_schema.json
d4straub: > IMO in both cases the seed should not be set unless specified by the user.

I thought its generally beneficial to set a seed to obtain the same result when running the pipeline twice. Is that not the way to go? Maybe I misunderstand?

conf/modules.config
d4straub: Alright, I opened https://github.com/nf-core/mag/issues/936 to not forget that.

nextflow_schema.json
SPPearce: I generally recommend that when modules are included in nf-core/modules that they should fix the random seed by default (but able to be overridden), to make the pipeline reproducible.

d4straub: Alright, thanks, then I leave it as is.

d4straub:
Profile `test_full` successfully finished. I think though SemiBin2 and COMEBin need relatively long, maybe we have to make them opt-in in future.

Thanks for all the comments and the reviews!

PR #933, opened 2025-11-17
Set concurrent connections for CheckM2 database download

dialvarezs:
Our slowest test is almost always `test_alternatives`, and this is entirely because of Zenodo’s transfer speeds when downloading the CheckM2 database. In some cases it takes over an hour (example run: https://github.com/nf-core/mag/actions/runs/19432875847/job/55596172319?pr=932). For reference, a "good" test run takes only around 11 minutes.

Since it’s currently not possible to generate our own CheckM2 test database because of the [db checksum requirement](https://github.com/chklovski/CheckM2/issues/140), and using the S3 mirror isn’t an option because of egress costs when users run tests locally, I propose increasing the number of concurrent connections to 4 when downloading the database.

This shouldn’t be considered abusive to Zenodo, as we’re downloading the same file as usual, just with multiple connections to improve reliability and reduce the impact of exceptionally slow transfers. AFAIK, only the API has rate limits, not the downloads themselves.

## PR checklist

### DIFF
--- a/conf/test_alternatives.config
+++ b/conf/test_alternatives.config
@@ -20,6 +20,12 @@ process {
     ]
 }
 
+process {
+    CHECKM2_DATABASEDOWNLOAD {
+        ext.args = '-x 4'
+    }
+}
+
 params {
     config_profile_name        = 'Test alternatives profile'
     config_profile_description = 'Minimal test dataset to check pipeline function'


jfy133:
Ooh clever! @dialvarezs do you have a comparison of run times? It looks like the tests didn't run here (we need to fix that so nf-test runs when modules.conf changes, also in the template

dialvarezs:
It’s partly a matter of luck. Sometimes we do get faster transfer speeds, which is why we occasionally see the 11-minute runtime. In local tests on my connection, the speed does improve, jumping from around 350 KB/s to about 1.2 MB/s, and sometimes even higher if it catches a good connection.

Given that, I think we should go ahead with the change and see how it performs on the ci runner.

jfy133 (approved):

PR #835, opened 2025-06-22
Add pydamage results to binqc table

jfy133:
See code TODOs

To close https://github.com/nf-core/mag/issues/833

## PR checklist

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/mag/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/mag _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [ ] Make sure your code lints (`nf-core pipelines lint`).
- [ ] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

PR #963, opened 2026-01-07
Summarise ancient DNA pyDamage results per bin and add to bin summary table 

### LINKED ISSUE #833: Ancient DNA mode: Add pyDamage results to bin_summary.tsv
### Description of feature

Currently the pyDamage results are on a contig level, but it would be good to get a summary statistic in `bin_summary.tsv` so a user can evaluate which bins contain contigs with aDNA damage, when running with ancient DNA mode.

This will likely require parsing the pydamage results file, parsing ??? to get contigs in each bin, sorting the pydamage results per-contig results into the bins, making the median average across all columns (@alexhbnr 's approach), then adding this to the bin_summary table.

jfy133:
To close #833 

Supersedes  Nextflow only version in #835 (which was having problems with data-flow consistency) with a python script only method.

Essentially:

1. Makes a contig to bin map
2. Separates per-contig pydamage results
3. Groups per-contig pydamage result per bin assignment (save)
4. Summarises pydamage contig results per bin with a median value
5. Attaches the median summaries to the bin_summary.tsv    

Additional minor changes:

- Fixes the official pyDamage results unnecessarily publishing `versions.yml` alongside the actual results
- Minor reformatting in `mag.nf` based on language server 

## PR checklist


subworkflows/local/binning_pydamage/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    // DOWNSTREAM: Final emitted TSV is then used in `mag.nf` to bind to final `bin_summary.tsv` table
+
+    ch_collected_pydamage_results = ch_contig_pydamage_results
+        .map { _meta, pydamage_report -> pydamage_report }
+        .toSortedList { pydamage_report ->
+            // Sort based on filename only (not full path) as work directory path will be different each run
+            file(pydamage_report).getBaseName()
+        }
+
+    ch_bin_contig_names = ch_input_for_postbinning
```
prototaxites:
```suggestion
    ch_bin_contig_names = ch_input_for_postbinning
```

Is there a reason not to use the contig2bin files from bin refinement (we could move their generation to elsewhere if needed!)?

jfy133: That is a very good point... I also had the same thought  it's odd that we didn't have that already... I just didn't think to look after binning 🤦‍♂️. But I had assumed if it wasn't already there maybe it isn't useful 😬.

Otherwise, I think it does make sense to make it immediatetely after binning anyway. I will look into whether it makes sense to move from binrefinement or via a different method.

Do you have any other comments otherwise?

prototaxites: > Do you have any other comments otherwise?

I had a quick skim, it looked fine at an early glance but I wanted to check the above (and offer the chance for potentially some bigger overhauls) prior to doing a proper review...

FWIW the binning refinement subworkflow includes FASTATOCONTIG2BIN 4 times... might be a good opportunity to move this to binning and just include it once, tracking stuff by `meta`? Would simplify a bunch of code!

jfy133: Ok sounds good! 

I don't fully understand what you mean about moving stuff to meta though? As in the contig names for each bin?

jfy133: Ok sounds good! 

I don't fully understand what you mean about moving stuff to meta though? As in the contig names for each bin?

prototaxites: I mean tracking the bin sets’ provenance (binner, etc.) using a tag in meta rather than by splitting them into separate process calls!

jfy133: Ah that bit! Yes understood! I vaguely remember now that where was maybe some tricky requirements with the contig2bin thing resulting in the current pattern, but will check to tomorrow!

jfy133: https://github.com/nf-core/mag/issues/964

docs/output.md
```diff
@@ -867,22 +867,29 @@ The output file in this directory is used as input for the dashboard [BIgMAG](ht
...
   - `[assembler]_[sample/group]/[sample/group]_/pydamage_results.csv`: PyDamage filtered result tabular file in `.csv` format. Format described here: [pydamage.readthedocs.io/en/0.62/output.html](https://pydamage.readthedocs.io/en/0.62/output.html)
+- `GenomeBinning/QC/
+  - `contig_to_bin_map.tsv`: Mapping file describing which contig of each assembly is assigned to which bin
+  - `pydamage_bins_summary.tsv`: A summary of pyDamage results for each bin, derived from a median average of each pyDamage values across all contigs of a particular each bin
+  - `summarise_pydamagebins/[assembler]-[binner]-[sample/group]_pydamagebin_results.tsv`: A 'raw' `pydamage analyse` results format file but reorded to have the per-contig values grouped together for each given bin
 
 </details>
 
+:::note
+If `--binning_map_mode` is set to `group` or `all`, you may see variations in the output of pyDamage between runs or `-resume`, due to evaluation on different sets of BAM files.
```
prototaxites, line 890: Why should that be the case?

workflows/mag.nf
```diff
@@ -545,7 +557,8 @@ workflow MAG {
     //
     // Collate and save software versions
     //
-    def topic_versions = channel.topic("versions")
+    def topic_versions = Channel
```
prototaxites:
```suggestion
    def topic_versions = channel
```

prototaxites (commented):
With the caveat that I don't have the energy to look at the Python code right now... 🥲

This looks good in general, but my main concern is the issue you raise with `--binning_map_mode`. I don't understand why you would get different sets of BAM files on resume - that seems concerning to me as I would assume this should be deterministic?

jfy133:
> With the caveat that I don't have the energy to look at the Python code right now... 🥲
> 
> This looks good in general, but my main concern is the issue you raise with `--binning_map_mode`. I don't understand why you would get different sets of BAM files on resume - that seems concerning to me as I would assume this should be deterministic?

I think this is to do with maybe a brittle implementation of support for co-binning that was added a while ago:

https://github.com/nf-core/mag/blob/ccfb744aab0c00a805a2dfbf7d9c9d8729c30805/subworkflows/local/ancient_dna/main.nf#L20-L21

I suspect the order of the BAM files in that channel are not stable across resumes when in `group` or `all` so differnet BAM filesa re used to generate the pyDamage results (and thus result in different contig names).

 I don't have the time to look deeper/experiment with this now at all though, and I only noticed it because the default in the test profile was set to 'group' (aDNA people normally will just set with `own`.
 
 So it's something I Can come back to in the future if it becomes a big problem..., but otherwise I really need this functioanlity with the common aDNA usecase merged in 3 months ago 😅

bin/summarise_pydamagebins.py
```diff
@@ -0,0 +1,122 @@
...
+    contig_to_bin_map.rename(columns={"contig_id": "reference"}, inplace=True)
+    contig_to_bin_map["assembly_id"] = contig_to_bin_map["assembly_id"].astype(str)
+
+    ## Repair contig names to match
+    ## Some tools remove everything after first space, so pyDamage has truncated headers vs. raw contigs
+    ## We strip this information from the raw contigs to meet at the simplest/lowest common denominator
+    contig_to_bin_map["reference"] = contig_to_bin_map.apply(
+        lambda row: (row["reference"].split(" ")[0]),
+        axis=1,
+    )
```
dialvarezs:
```suggestion
    contig_to_bin_map["reference"] = contig_to_bin_map["reference"].str.extract("([^ ]+)")
```
A bit more elegant

bin/summarise_pydamagebins.py
```diff
@@ -0,0 +1,122 @@
...
+import argparse
+import pandas as pd
+from pathlib import Path
+
+
+def parse_args(args=None):
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-c",
+        "--contig_to_bin_map",
```
dialvarezs:
```suggestion
        "--contig-to-bin-map",
```
[picky]
Personally I don't like underscores in cli options. Python does the conversion automatically, so changing it here will not affect the code (but cli call in .nf could need updating).

subworkflows/local/binning_pydamage/main.nf
```diff
@@ -0,0 +1,32 @@
...
+    ch_versions = channel.empty()
+
+    // 1. Collect all pydamage results into single tuple
+    // 2. Generate list of contigs per bin
+    // 3. Local module that reorders/reassigns contigs (save), and also summarises pydamage results per bin via median
+    // 4. Emit final tsv and versions
+    // DOWNSTREAM: Final emitted TSV is then used in `mag.nf` to bind to final `bin_summary.tsv` table
+
+    ch_collected_pydamage_results = ch_contig_pydamage_results
+        .dump(tag: 'ch_contig_pydamage_results')
```
dialvarezs: Leaving the dumps is intended?

subworkflows/local/binning_pydamage/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - ch_input_for_postbinning:
+      description: Binned contigs in FASTA file format
+      structure:
+        - meta:
+            type: map
+            description: Metadata map
+        - fasta:
+            type: file
+            description: Binned contigs in FASTA file format
+            pattern: "*.{fasta,fa}  "
```
dialvarezs:
```suggestion
            pattern: "*.{fasta,fa}"
```
Also: add ontology

bin/summarise_pydamagebins.py
```diff
@@ -0,0 +1,122 @@
...
+        required=True,
+        metavar="FILE",
+        help="Input file of tab-separated list of <bin_id>\t<contig_id> mappings. No header.",
```
dialvarezs: Help says that it expects no header, but it does expect `["contig_id", "assembly_id"]` in the code

bin/summarise_pydamagebins.py
```diff
@@ -0,0 +1,122 @@
...
+        if args.verbose:
+            print("[summarise_pydamagebins.py] PROCESSING: Loading " + report)
+        assembly_id = os.path.basename(report).replace("_pydamage_results.csv", "")
+        pydamage_df = pd.read_csv(report, sep=",")
+        pydamage_df["assembly_id"] = assembly_id
+        pydamage_df.insert(0, "assembly_id", pydamage_df.pop("assembly_id"))
+        pydamage_reports_dfs.append(pydamage_df)
+
+    pydamage_reports_all = pd.concat(pydamage_reports_dfs)
+    pydamage_reports_all["assembly_id"] = pydamage_reports_all["assembly_id"].astype(
```
dialvarezs, line 84: This should not be necessary, because `assembly_id` is the result of `basename(report).replace(...)`, which should always return a string.

bin/summarise_pydamagebins.py
```diff
@@ -0,0 +1,122 @@
...
+
+    for name, group in pydamage_bin_summary:
+        filename = Path(name).stem + "_pydamage_bin_results.tsv"
+        group.to_csv(filename, sep="\t", index=False)
+
+    pydamage_bin_summary_median = pydamage_bin_summary.median(
+        numeric_only=True
+    ).sort_values(by=["bin_id"])
+
+    ## Testing only
```
dialvarezs: Why testing?

dialvarezs (approved):
Looks perfect to me!
None of my comments is blocking.

bin/summarise_pydamagebins.py
jfy133:
```suggestion
        help="Input file of tab-separated list of <bin_id>\t<contig_id> mappings.",
```

bin/summarise_pydamagebins.py
jfy133: It didn't which is why I added it :sweat_smile: I'll leave it fornow ;)

bin/summarise_pydamagebins.py
jfy133:
```suggestion
```

subworkflows/local/binning_pydamage/main.nf
jfy133:
```suggestion
```

subworkflows/local/binning_pydamage/main.nf
```diff
@@ -0,0 +1,32 @@
...
+        .map { _meta, pydamage_report -> pydamage_report }
+        .toSortedList { pydamage_report ->
+            // Sort based on filename only (not full path) as work directory path will be different each run
+            file(pydamage_report).getBaseName()
+        }
+        .dump(tag: 'ch_collected_pydamage_results')
```
jfy133:
```suggestion
```

PR #965, opened 2026-01-09
Generate contig to bin maps for all bins during binning and simplify bin refinment DAS_Tool contig2bin code

### LINKED ISSUE #964: Generate contig-to-bin maps for all bins during binning and simplify code
### Description of feature

As noted by @prototaxites here: https://github.com/nf-core/mag/pull/963#discussion_r2668561699

We currently only generate contig-to-bin name maps when running binrefinement with DAS_Tool, even though it could be useful in other contexts too (case in point: https://github.com/nf-core/mag/pull/963).

Furthermore, the current use of the DAS_Tool module for making such a file is also very inefficient code wise, we import it multiple times for each binner whereas we could just run it once for all, even though they now all have the same suffix.

Therefore the proposal here is to move the contig-to-bin map generation to binning, invoke the module only once for all bins, and export this as a channel that can then be loaded into bin refinement and elsewhere that needs them if executed.

Then I can update/further simplify #963

jfy133:
Close #964 

Summary:

- Simplifies DAS_Tool preparation  to a single FASTATOCONTIG2BIN process rather than one per binner
- Adds (optional) parameter to save the files above
- Adds new default generation of a similar contig to bin mapping file, but more detailed/structured file that contains all contigs/bins but at binning rather than at refinement.

Also includes the obligatory typical Nextflow auto-formatting ;)

## PR checklist

### DIFF
--- a/conf/test_assembly_input.config
+++ b/conf/test_assembly_input.config
@@ -22,7 +22,7 @@ process {
 
 process {
     withName: METAEUK_EASYPREDICT {
-        cpus = 1  // to keep output order consistent
+        cpus     = 1
         ext.args = '-s 1 --max-seqs 10 --max-accept 1'
     }
 }
@@ -47,6 +47,7 @@ params {
     skip_concoct                    = false
     skip_comebin                    = true
     skip_metabinner                 = true
+    skip_semibin                    = true
 
     refine_bins_dastool             = true
     refine_bins_dastool_threshold   = 0.0
--- a/tests/test_assembly_input.nf.test.snap
+++ b/tests/test_assembly_input.nf.test.snap
@@ -65,7 +65,7 @@
     },
     "-profile assembly_input": {
         "content": [
-            294,
+            282,
             {
                 "ADJUST_MAXBIN2_EXT": {
                     "coreutils": 9.5
@@ -155,9 +155,6 @@
                 "RENAME_PREDASTOOL": {
                     "coreutils": 9.5
                 },
-                "SEMIBIN_SINGLEEASYBIN": {
-                    "SemiBin": "2.2.0"
-                },
                 "SEQKIT_STATS": {
                     "seqkit": "2.9.0"
                 },
@@ -175,7 +172,7 @@
             "nf-test": "0.9.2",
             "nextflow": "25.10.2"
         },
-        "timestamp": "2026-01-09T14:08:19.594040545"
+        "timestamp": "2026-01-13T17:10:10.368851859"
     },
     "binning": {
         "content": [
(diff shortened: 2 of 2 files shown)


prototaxites (approved):
LGTM, but one query: for the Pydamage PR, will you want to include the results for refined bins also? Because here the new contig to bin map with the custom code will exclude those as bin refinement happens after this map is already generated!

docs/output.md
```diff
@@ -555,6 +556,16 @@ For each bin or refined bin the median sequencing depth is computed based on the
 
 </details>
 
+### Contig to bin map
+
+The pipeline also generates a 'contig to bin map to help users explore track where each contig from a given assembly was assigned to which bin.
```
d4straub:
```suggestion
The pipeline also generates a 'contig to bin map' to help users explore track where each contig from a given assembly was assigned to which bin.
```
?

docs/output.md
```diff
@@ -555,6 +556,16 @@ For each bin or refined bin the median sequencing depth is computed based on the
...
+
+<details markdown="1">
+<summary>Output files</summary>
+- `GenomeBinning/depths/contig_to_bin/`
```
d4straub:
```suggestion
- `GenomeBinning/contig_to_bin/`
```
This seems to be not exported to the depths folder (and it shouldnt)

d4straub (commented):
Just two comments

docs/output.md
jfy133: oops, indeed! copy paste error!

jfy133:
> LGTM, but one query: for the Pydamage PR, will you want to include the results for refined bins also? Because here the new contig to bin map with the custom code will exclude those as bin refinement happens after this map is already generated!

Ah good point... I was thinking too aDNA centric again (bins are normally pretty poor so refinement doesn't help much.)

I'll try to see how easy it is to move it to post refinement with `ch_input_for_postbinning`

workflows/mag.nf
```diff
@@ -377,8 +375,9 @@ workflow MAG {
             ? ch_input_for_postbinning_bins
             : ch_input_for_postbinning_bins.mix(ch_input_for_postbinning_unbins)
 
-        // Combine short and long reads by meta.id and meta.group for DEPTHS, making sure that
-        // read channel are not empty
+        /* Combine short and long reads by meta.id and meta.group for DEPTHS, making sure that
+         read channel are not empty
+        */
```
prototaxites: <very minor> Why change the comment type here, but then in the contig2bin map section you still use `//`?

workflows/mag.nf
```diff
@@ -391,6 +390,28 @@ workflow MAG {
...
+            .splitFasta(record: [header: true], elem: 1)
+            .map { meta, contig_header ->
+                "assembly_id\tcontig_id\tbinner\tbin_id\n${meta['assembler']}-${meta['id']}\t${contig_header['header']}\t${meta['binner']}\t${meta['bin_id']}\n"
+            }
+            .collectFile(
+                name: 'contig_to_bin_map.tsv',
+                storeDir: params.outdir + '/GenomeBinning/contig_to_bin/',
+                keepHeader: true,
+                sort: true,
+            )
```
prototaxites, line 412: This looks good, only (very optional, maybe not nf-core-approved) suggestion is to move this to a native `exec:` process, and write the output to a file in a process task directory. 

Would be more appropriate approach future-wise with channel-based publishing (plus apparently `collectFile()` is slated to be removed in a revised operator list)...

workflows/mag.nf
jfy133: ah because it was more of a 'general section', but I forgot to to split it up. Will revert as it's not important.

workflows/mag.nf
jfy133: I'm spending too much time on optimising now rather than getting my paper finished ;) , so I'll leave this for a future but it's a fair point.

jfy133:
@nf-core-bot fix linting

jfy133:
Waiting for local `nf-test` snapshot to be updated, then I will merge!

Thanks @prototaxites !

jfy133:
Running on github actions also passes fine :( 
image

jfy133:
OK after investigation via @dialvarezs , there was variability with SEmiBin, so deactivated in the test where DAS_Tool is run
Now we only have the persisting checkm2 issue

Issue #985, opened 2026-03-04
Incorrect validation of `long_reads_platform`


vinisalazar:
### Description of the bug

Hi,

I'm trying to use `mag` with both short and long reads.

The [docs for `v5.4.0`](https://nf-co.re/mag/5.4.0/docs/usage/), the version I am using, state that:

> If long reads are provided (long_reads), the long_reads_platform column is required. Valid options include:
> 
> OXFORD_NANOPORE: Oxford Nanopore Technologies (ONT) reads, which may have higher error rates compared to newer technologies.
> OXFORD_NANOPORE_HQ: High-quality ONT reads, typically with an error rate of less than 5%, achievable with the latest ONT chemistries and sequencing platforms. This option should generally be used unless working with older data.
> PACBIO_SMRT: Pacific Biosciences Single Molecule Real-Time (SMRT) sequencing reads.

My samplesheet has "PACBIO_SMRT" as the `long_reads_platform`, but I get this error:

```
ERROR ~ Validation of pipeline parameters failed!

 -- Check '.nextflow.log' file for details
The following invalid input values have been detected:

* --input (samplesheet.csv): Validation of file failed:
        -> Entry 1: Error for field 'long_reads_platform' (PACBIO_SMRT): Expected any of [[OXFORD_NANOPORE, OXFORD_NANOPORE_HQ, PACBIO_CLR, PACBIO_HIFI]]
        -> Entry 2: Error for field 'long_reads_platform' (PACBIO_SMRT): Expected any of [[OXFORD_NANOPORE, OXFORD_NANOPORE_HQ, PACBIO_CLR, PACBIO_HIFI]]
        -> Entry 3: Error for field 'long_reads_platform' (PACBIO_SMRT): Expected any of [[OXFORD_NANOPORE, OXFORD_NANOPORE_HQ, PACBIO_CLR, PACBIO_HIFI]]
        -> Entry 4: Error for field 'long_reads_platform' (PACBIO_SMRT): Expected any of [[OXFORD_NANOPORE, OXFORD_NANOPORE_HQ, PACBIO_CLR, PACBIO_HIFI]]
        -> Entry 5: Error for field 'long_reads_platform' (PACBIO_SMRT): Expected any of [[OXFORD_NANOPORE, OXFORD_NANOPORE_HQ, PACBIO_CLR, PACBIO_HIFI]]
[...]

PR #988, opened 2026-03-05
Update docs re: long_reads_platform

vinisalazar:
See #985


dialvarezs (approved):
Thanks @vinisalazar!

Can you add an changelog entry for this?

PR #1039, opened 2026-05-21
feat: add test_longreadonly snapshot

dialvarezs:
## PR checklist


tests/test_longreadonly.nf.test
```diff
@@ -10,29 +10,165 @@ nextflow_pipeline {
...
+                    "MetaBAT2 produced classified or unclassified output: ${metabat2_bins.size() + metabat2_unbins.size() > 0}",
+                    // Summary row count checks
+                    "Bin summary non-empty: ${bin_summary.csv(sep: '\t').rowCount > 0}",
+                    "Bin summary row count matches bins: ${bin_summary.csv(sep: '\t').rowCount == maxbin2_bins.size() + metabat2_bins.size()}",
+                    "BUSCO summary non-empty: ${busco_summary.csv(sep: '\t').rowCount > 0}",
+                    "Bin depths summary non-empty: ${bin_depths_summary.csv(sep: '\t').rowCount > 0}",
+                    "QUAST bin summary non-empty: ${quast_bin_summary.csv(sep: '\t').rowCount > 0}",
+                ).match("content-checks") },
+
+                { assert snapshot(
```
jonasscheid, line 156: Maybe readlines is even faster here

tests/.nftignore
```diff
@@ -3,11 +3,10 @@ Ancient_DNA/pydamage/analyze/*/*.csv
 Ancient_DNA/variant_calling/*/*.vcf.gz
 Annotation/Prokka/**/*.{err,gbk,log,sqn}
 Annotation/Prokka/**/*.tmp.*
-Assembly/FLYE/*.log
-Assembly/MEGAHIT/*.log
+Assembly/*/*.log
+Assembly/*/QC/*/QUAST/*.{pdf,html,log}
+Assembly/*/QC/*/QUAST/**/*.{pdf,html}
```
jonasscheid, line 8: Double check if log should also be ignored here

tests/test_longreadonly.nf.test
```diff
@@ -10,29 +10,165 @@ nextflow_pipeline {
...
-                { assert snapshot(stable_name_multiqc, stable_path_multiqc).match('multiqc') },
+                { assert snapshot(stable_name_qc,         stable_path_qc        ).match('qc') },
+                { assert snapshot(stable_name_assembly,   stable_path_assembly  ).match('assembly') },
+                { assert snapshot(stable_name_binning,    stable_path_binning   ).match('binning') },
+                { assert snapshot(stable_name_annotation, stable_path_annotation).match('annotation') },
+                { assert snapshot(stable_name_multiqc,    stable_path_multiqc   ).match('multiqc') },
+
+                { assert snapshot(
+                    // FASTA content checks
+                    "Flye assemblies non-empty: ${flye_assemblies.collect { file -> path("${file}").fasta.size() > 0 }.every()}",
```
jonasscheid: I'm a bit overwhelmed by the verbosity. Wouldn't asserting per file be more concise?

```bash
          // ── correctness: hard asserts, never touch the .snap ──
          { assert flye.every { path(it).fasta.size() > 0 } },
          { assert nanostats.every { path(it).text.contains('General summary:') } },
          { assert bin_summary.csv(sep: '\t').rowCount == 4 },
          // …the rest of the content-/log-checks become asserts too
```

jonasscheid (changes requested):

tests/.nftignore
dialvarezs: Most logs include timestamps, that's why  we're excluding them from the snapshot.
I double-checked them.

tests/test_longreadonly.nf.test
dialvarezs: Yeah, but I guess using the CSV plugin also checks a valid CSV structure, and that's a plus.
These are not big CSVs, so the speed improvement should be pretty negligible.

tests/test_longreadonly.nf.test
dialvarezs: To be honest, we intentionally chose this pattern in MAG, and it’s already used in other profile snapshots as well. I find the messages on the snapshots useful, and in the end they work as assertions as well.
We can revisit this pattern across all profiles in the future, but for now I’d prefer to keep things consistent.
I did take your suggestion about using every instead of chaining collect and every, though, thanks for that!

dialvarezs:
Hi @jonasscheid, thanks for your review, and sorry for the late response.

It seems that all of your concerns were actually deliberate decisions on our side, so in the end I only picked the expression simplification suggestion.

Let me know if you have some other concerns!

jonasscheid:
Thanks for explaining, LGTM 👍🏼

jonasscheid (approved):

PR #1048, opened 2026-06-02
Add PyPOLCA polishing and use polished assemblies for long-read binning

Harshita-sriv:
## Summary

This PR adds a PyPOLCA polishing module and integrates polished assemblies into the long-read MAG workflow.

## Changes

* Added a new PyPOLCA module and conda environment.
* Added publication of PyPOLCA outputs.
* Added the `run_pypolca` parameter.
* Modified the workflow so that long-read binning uses `ch_polished_assemblies` instead of raw `ch_longread_assemblies`.


dialvarezs:
Hi @Harshita-sriv, thanks for this!
For the meantime, can you join the nf-core org? That way you will able to run the ci without explicit approval.
https://nf-co.re/join

modules/local/pypolca/main.nf
dialvarezs, line 1: Please use official nf-core pypolca module instead:
https://github.com/nf-core/modules/blob/master/modules/nf-core/pypolca/run/main.nf

workflows/mag.nf
```diff
@@ -206,7 +207,45 @@ workflow MAG {
         ch_shortread_assemblies = ch_assemblies.filter { meta, _contigs -> meta.assembler.toUpperCase() in ['SPADES', 'SPADESHYBRID', 'MEGAHIT'] }
         ch_longread_assemblies = ch_assemblies.filter { meta, _contigs -> meta.assembler.toUpperCase() in ['FLYE', 'METAMDBG'] }
     }
+    
+    if (params.run_pypolca) {
+    
+    ch_polished_assemblied = ch_longread_assemblies
+    
+    ASSEMBLY.out.longread_assemblies.view()
```
dialvarezs:
```suggestion
```
Please remove these debug lines.

workflows/mag.nf
dialvarezs, line 1: Can you format the file using the VSCode extension?
That will handle indentation

dialvarezs (changes requested):
In general I think it looks right, but needs to migrate to the nf-core module, and of course, pass the ci.

modules/local/pypolca/main.nf
Harshita-sriv: Okay!

workflows/mag.nf
Harshita-sriv: Okay, I will go through this and remove these debug lines

Harshita-sriv:
Thanks for the review. I investigated the CI failures and found that the current implementation does not define ch_polished_assemblies when run_pypolca is not enabled, which causes the test workflows to fail. I'll address this together with the requested migration to the official nf-core PyPOLCA module, removal of debug statements, and formatting fixes

dialvarezs:
Hi @Harshita-sriv! I think this looks fine, but it would be ideal to wire it up in some test. I'll think what would be the best option here and I'll come back to you.
For the meantime, can you update the changelog?
Thanks!

nextflow_schema.json
```diff
@@ -1141,5 +1142,10 @@
         {
             "$ref": "#/$defs/ancient_dna_assembly"
         }
-    ]
+    ],
+    "properties": {
+        "run_pypolca": {
```
d4straub: that seems not the optimal place for that parameter. I think it might fit into [Assembly options](https://nf-co.re/mag/dev/parameters/#assembly-options)?

workflows/mag.nf
```diff
@@ -207,6 +209,23 @@ workflow MAG {
         ch_longread_assemblies = ch_assemblies.filter { meta, _contigs -> meta.assembler.toUpperCase() in ['FLYE', 'METAMDBG'] }
     }
 
+    ch_polished_assemblies = ch_longread_assemblies
+
+    if (params.run_pypolca) {
+
+    PYPOLCA_RUN(
+       ch_short_reads,
+       ch_longread_assemblies
```
d4straub: doesnt that potentially mix up reads from different samples and assemblies? What about co-assemblies? shouldnt there be some sort of restriction on what short reads are used for polishing, maybe similar to https://nf-co.re/mag/dev/parameters/#coassemble_group & https://nf-co.re/mag/dev/parameters/#binning_map_mode?

conf/modules.config
```diff
@@ -436,6 +436,14 @@ process {
         ]
     }
 
+    withName: 'NFCORE_MAG:MAG:PYPOLCA' {
```
d4straub: I have seen the module specifies "--careful" when no args are given, but that might be removed at any time with a module update, so I would force that here. Because `--careful` seems important for metagenome polishing to not add errors during polishing.

```suggestion
    withName: 'NFCORE_MAG:MAG:PYPOLCA' {
        ext.args   = "--careful"
```

d4straub (commented):
Thanks for the addition, I think thats absolutely desired!
Looks good but I have some concerns with multiple samples.

workflows/mag.nf
Harshita-sriv: Good point. Looking into the workflow, the assembly subworkflow performs its own grouping when `--coassemble_group` is enabled, whereas the current PyPOLCA implementation is fed from `ch_short_reads` directly. Would you prefer that PyPOLCA mirrors the assembly grouping logic and uses pooled reads for coassemblies, or should polishing be restricted to non-coassembled assemblies?

d4straub: > Would you prefer that PyPOLCA mirrors the assembly grouping logic and uses pooled reads for coassemblies, or should polishing be restricted to non-coassembled assemblies?

I think I would prefer the assembly grouping logic. Because there its expected that similar taxa are in the samples, thats why co-assembly is applicable, therefore polishing should work as well within that sample group I think. Having polishing restricted to non-coassembles assemblies seems unnecessarily restrictive?

Harshita-sriv:
I noticed that the remaining lint failure is due to modifications in `modules/nf-core/pypolca/run`, which no longer matches the upstream module.

The changes were introduced because PyPOLCA was failing during testing, and with the modifications the pipeline completes successfully on the test dataset.

Would it be acceptable to keep these modifications to the PyPOLCA module, or would you prefer that the implementation uses the upstream module unchanged and handles the required logic elsewhere?

d4straub:
> Would it be acceptable to keep these modifications to the PyPOLCA module, or would you prefer that the implementation uses the upstream module unchanged and handles the required logic elsewhere?

Some of the changes are not acceptable: the output block and the versioning must be kept as it was originally. Additionally the line `ch_versions = ch_versions.mix(PYPOLCA_RUN.out.versions)` in `workflows/mag.nf` has to be removed.
The incoming data (contigs) should be compressed as in the original code. Even if it would be necessary to allow non-compressed input (which may be ok in same cases), copying the files when already uncompressed seems like an unnecessary burden on IO and file sizes to me (assemblies in nf-core/mag can be huge!).

Generally, small changes on nf-core modules can be done and accepted by the linting via https://nf-co.re/docs/nf-core-tools/cli/modules/patch

Harshita-sriv:
@d4straub I restored the PyPOLCA module to the upstream nf-core implementation and moved the coassembly-specific logic into the workflow layer (`workflows/mag.nf`). The workflow now groups short reads appropriately when `--coassemble_group` is enabled before passing them to PyPOLCA.

The remaining CI issue appears to be the `check_local_copy` lint check for `pypolca/run`. Locally, the module matches the installed nf-core module definition (`nf-core modules info pypolca/run`), and the workflow compiles and reaches `PYPOLCA_RUN` in stub runs. All other CI checks pass.

Could you advise whether there is an additional module sync/patch step expected for this module?

d4straub:
Looks good.
I think the easiest way of satisfying the linting is to re-install the module.

modules/nf-core/pypolca/run/main.nf
```diff
@@ -0,0 +1,52 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args          = task.ext.args   ?: '--careful'
+    prefix            = task.ext.prefix ?: "${meta.id}"
+    def read_files    = reads instanceof List ? reads : [reads]
+    def read_file_arg = read_files.size() > 1 ? "-1 ${read_files[0]} -2 ${read_files[1]}" : "-1 ${read_files[0]}"
+    """
+    gzip -cdf $contigs > contigs_uncompressed
```
dialvarezs: Can you update the module so it only decompresses when the input has a `.gz` extension? I was testing this locally, and pypolca received the uncompressed assembly, throwing an error.
And also I think it's a better design not assuming that the input is compressed.

dialvarezs (commented):
Hi again @Harshita-sriv! This is looking good!

A few last things:

Regarding the testing setup:
I think the best option is to use the `longreadonly_alternatives` but changing the samplesheet so it uses the hybrid one that has LR and SR data. Then disable SPAdes, SPAdes hybrid and MEGAHIT at least. That way we can get this tested in that profile.

Also I left a comment regarding the pypolca module requiring gzipped input (that will need updating the nf-core module).

modules/nf-core/pypolca/run/main.nf
Harshita-sriv: Yes, I also faced this problem, but I didn't change the module because I assumed it wouldn't be accepted. However, for the testing setup, could you confirm that for test_longreadonly_alternatives, I should replace samplesheet.long_read.v4.csv with samplesheet.hybrid.v4.csv, enable run_pypolca, and keep only Flye active by disabling SPAdes, SPAdes-hybrid, MEGAHIT, and MetaMDBG?

dialvarezs: It's a totally reasonable change if you execute gzip based on the extension, so it works with compressed and uncompressed input. So don't worry about opening a PR on the module. I can review it if you ping me.

And yes, that's the setup I'm proposing. MetaMDBG is already disabled in that profile, so you'll need to disable both SPAdes and MEGAHIT.

Harshita-sriv: I've implemented the requested changes:

* Updated `test_longreadonly_alternatives.config` to use the hybrid samplesheet and enable `run_pypolca` while disabling SPAdes/SPAdesHybrid/MEGAHIT.
* Updated the Pypolca module to support both gzipped and uncompressed assembly inputs.

The remaining issue is CI validation. The PR currently shows failures related to:

1. The Pypolca module modification.
2. The updated test profile / snapshot expectations.

Locally I'm unable to fully validate the nf-test because the test environment is failing before completion (missing tools such as `fastp`, `NanoPlot`, `porechop`), so any snapshot updates generated locally appear incomplete.

Could you please take a look and let me know how you'd like the test/snapshot updates handled, or whether the current implementation looks correct and can be reviewed from your side?

dialvarezs: Oh, sorry for the lack of clarification.
As Daniel said, updating the nf-core module directly in the mag repo is not the way.
To update a module you need open a PR on the modules repo: https://github.com/nf-core/modules. When your PR is accepted and merged, you can run `nf-core modules update pypolca/run` to update the module here.

Harshita-sriv: Okay, will do that. Can you also suggest about the errors due to test file modifications.

dialvarezs: You can ignore the errors for now. When you do the update via the official nf-core modules, the error will go away.

Harshita-sriv: Okay, thank you for help!

Harshita-sriv:
Hey @dialvarezs 

I'm still seeing the remaining CI failures, and I haven't been able to reproduce/update the snapshots locally. Since you made the latest changes, could you take a look and let me know whether there's anything else I should update on my side?

dialvarezs:
Hi @Harshita-sriv, it just needed to regenerate the snapshot and exclude the VCF, because it has the date.
I also realized that we missed one point: Nextflow channels are not guaranteed to be emited in the same order. CI pass because it's only one sample. I fixed that with a combine and multiMap mix.

dialvarezs:
For me this looks good to merge now. Can you give it a final look @d4straub?

Harshita-sriv:
Hey, @dialvarezs! 
Thanks for fixing those issues and for the explanation! I'll take a look through the changes to understand them better. Thanks again for the review.

tests/test_longreadonly_alternatives.nf.test.snap
```diff
@@ -48,80 +48,87 @@
...
-                "GenomeBinning/QC/BUSCO/FLYE-MaxBin2-unclassified-unrefined-minigut/minigut-bacteria_odb10-busco/FLYE-MaxBin2-minigut.002.fa/run_bacteria_odb10/short_summary.txt",
-                "GenomeBinning/QC/BUSCO/FLYE-MaxBin2-unclassified-unrefined-minigut/minigut-bacteria_odb10-busco/batch_summary.txt",
-                "GenomeBinning/QC/BUSCO/FLYE-MaxBin2-unclassified-unrefined-minigut/short_summary.specific.bacteria_odb10.FLYE-MaxBin2-minigut.001.fa.json",
-                "GenomeBinning/QC/BUSCO/FLYE-MaxBin2-unclassified-unrefined-minigut/short_summary.specific.bacteria_odb10.FLYE-MaxBin2-minigut.001.fa.txt",
-                "GenomeBinning/QC/BUSCO/FLYE-MaxBin2-unclassified-unrefined-minigut/short_summary.specific.bacteria_odb10.FLYE-MaxBin2-minigut.002.fa.json",
-                "GenomeBinning/QC/BUSCO/FLYE-MaxBin2-unclassified-unrefined-minigut/short_summary.specific.bacteria_odb10.FLYE-MaxBin2-minigut.002.fa.txt",
+                "GenomeBinning/MaxBin2/bins/null-MaxBin2-minigut.001.fa.gz",
```
d4straub: This looks bad, this shouldnt be `null` should it?

tests/test_longreadonly_alternatives.nf.test.snap
```diff
@@ -48,80 +48,87 @@
             [
                 "Assembly/FLYE/FLYE-minigut.assembly.fasta.gz",
                 "Assembly/FLYE/FLYE-minigut.assembly_graph.gfa.gz",
-                "Assembly/FLYE/FLYE-minigut.flye.log"
+                "Assembly/FLYE/FLYE-minigut.assembly_info.txt",
+                "Assembly/FLYE/FLYE-minigut.flye.log",
+                "Assembly/PYPOLCA/minigut/minigut/minigut.report",
+                "Assembly/PYPOLCA/minigut/minigut/minigut.vcf",
+                "Assembly/PYPOLCA/minigut/minigut/minigut_corrected.fasta"
```
d4straub: That file naming doesnt follow the usual pattern, does it? I would have expected the the prefix includes the assembler as well, such as FLYE-minigut.report and potentially even pypolca in the name. Also, the "corrected" is not ideal imho, because there might be additional polishing tools added and those also correct the assembly; I would rather use "pypolca" instead of "corrected" for "minigut_corrected.fasta".

workflows/mag.nf
```diff
@@ -207,6 +209,54 @@ workflow MAG {
...
+        if (params.coassemble_group) {
+
+            ch_short_reads_pypolca = ch_short_reads
+                .map { meta, reads -> [meta.group, meta, reads] }
+                .groupTuple(by: 0)
+                .map { group, metas, reads ->
+                    def assemble_as_single = params.single_end || (params.bbnorm && params.coassemble_group)
+
+                    def meta = [id: "group-${group}", group: group, single_end: assemble_as_single, sr_platform: metas.sr_platform[0]]
+
```
d4straub, line 222: to me those are too many empty lines ;D

d4straub (commented):
I think I still found some relevant inconsistencies.

tests/test_longreadonly_alternatives.nf.test.snap
Harshita-sriv: Looking at the PyPOLCA module, I noticed that it receives both read metadata (meta) and assembly metadata (meta2), but currently emits the polished assembly using meta. Since the assembler field is added to the assembly metadata upstream, I suspect this might be where the null-MaxBin2-* naming is coming from. Would switching the outputs to emit meta2 be the right fix here? If so, should this be addressed in the module itself (via nf-core/modules) or in the workflow integration?

tests/test_longreadonly_alternatives.nf.test.snap
Harshita-sriv: The current *_corrected.fasta naming also comes directly from the PyPOLCA module output. I agree that including the assembler and/or polishing step would make the output more explicit. Would you prefer handling this in the PyPOLCA module itself or in the workflow integration?

d4straub: In the workflow integration.

tests/test_longreadonly_alternatives.nf.test.snap
d4straub: Indeed, switching to the meta coming from the assembly should help.
Addressed in the workflow, not in the module.

conf/modules.config
```diff
@@ -435,16 +435,21 @@ process {
...
-    }
-
+   
+   withName: 'NFCORE_MAG:MAG:PYPOLCA_RUN' {
+      ext.args = '--careful'
+      publishDir = [
+          path: { "${params.outdir}/Assembly/PYPOLCA/${meta.id}" },
+          mode: params.publish_dir_mode,
+          saveAs: { filename ->
+              if (filename == 'versions.yml') {
```
dialvarezs: Pypolca module already uses channels, so you can remove this block (`versions.yml` will not be generated ever).

Harshita-sriv: Hi @dialvarezs 
I can see the CI failures related to snapshot after the PYPOLCA output rename. How should I regenerate the affected longreadonly snapshots locally and commit them? Is there another approach you'd recommend?
I tried regenerating the snapshots locally with nf-test --update-snapshot, but the tests fail before completion. What's the recommended way to update the affected snapshot files in this case?

dialvarezs: To update the snapshot, you need to run this:
```
nf-test test --profile=+docker --update-snapshot tests/test_longreadonly_alternatives.nf.test
```
(You can change docker by singularity or conda depending on your environment)
If that command fails, it mean that there was an execution error. What error are you seeing?

tests/test_longreadonly_alternatives.nf.test.snap
```diff
@@ -214,6 +236,9 @@
                 "PORECHOP_PORECHOP": {
                     "porechop": "0.2.4"
                 },
+                "PYPOLCA_RUN": {
+                    "pypolca": "pypolca, version 0.4.0"
```
dialvarezs: It seems that the command to report Pypolca version is wrong. Can you open a PR to update the module? It should return just the number (0.4.0 int this case).

Harshita-sriv: I looked into it and the module currently reports the version using:
eval('pypolca --version')
which returns:
pypolca, version 0.4.0

I checked a few other nf-core modules (e.g. FastQC, MultiQC, NanoPlot), and they normalize version strings before reporting them. I tested:

pypolca --version | sed 's/^pypolca, version //'
which correctly returns:
0.4.0
I'll open a small PR against nf-core/modules to update the PyPOLCA module accordingly, then update here.

Harshita-sriv:
Hey @d4straub and @dialvarezs,

I've addressed the review comments, updated MAG to the latest merged PyPOLCA module revision, and regenerated the affected snapshots. The longreadonly_alternatives test passes locally with Docker.

I had to rerun the CI jobs after canceling a workflow while investigating a long-running NF-test shard. The CI checks have now been completed successfully. 🎉

Do you think this is good to go? Please say yes! 😄

tests/test_longreadonly_alternatives.nf.test.snap
```diff
@@ -48,17 +48,24 @@
             [
                 "Assembly/FLYE/FLYE-minigut.assembly.fasta.gz",
                 "Assembly/FLYE/FLYE-minigut.assembly_graph.gfa.gz",
-                "Assembly/FLYE/FLYE-minigut.flye.log"
+                "Assembly/FLYE/FLYE-minigut.assembly_info.txt",
+                "Assembly/FLYE/FLYE-minigut.flye.log",
+                "Assembly/PYPOLCA/minigut/minigut/minigut.report",
+                "Assembly/PYPOLCA/minigut/minigut/minigut.vcf",
+                "Assembly/PYPOLCA/minigut/minigut/minigut_pypolca.fasta"
```
d4straub: I am still a bit concerned here: In this current test, only one assembler (FLYE) is run, produces `"Assembly/FLYE/FLYE-minigut.assembly.fasta.gz"` and corrected to file `"Assembly/PYPOLCA/minigut/minigut/minigut_pypolca.fasta"`. But if two assemblers are run, FLYE and metaMDBG, both corrected assemblies will be named to `"Assembly/PYPOLCA/minigut/minigut/minigut_pypolca.fasta"`, overwriting each other.
I think the file name must include the assembler. Or did I get something wrong?

conf/modules.config
```diff
@@ -436,6 +436,17 @@ process {
         ]
     }
 
+   withName: 'NFCORE_MAG:MAG:PYPOLCA_RUN' {
+      ext.args = '--careful'
```
d4straub, line 441: Adding here 
```suggestion
      ext.prefix = { "${meta.assembler}-${meta.id}" }
      ext.args = '--careful'
```
might solve the naming issue and make file names assembler-specific. I am not sure though whether `${meta.assembler}` exists at that point in the meta map?

tests/test_longreadonly_alternatives.nf.test.snap
Harshita-sriv: Done with the changes. meta2.assembler came in handy! I've updated the naming to preserve the assembler in the PyPOLCA outputs and regenerated the snapshots.

workflows/mag.nf
```diff
@@ -207,6 +209,68 @@ workflow MAG {
...
+                [meta.id, polished]
+            }
+            .join(
+                ch_longread_assemblies.map { meta, assembly ->
+                    [meta.id, meta]
+                }
+            )
+            .map { id, polished, meta ->
+                [meta, polished]
+            }
```
dialvarezs: If there are multiple assemblers enabled, this join could be problematic. We're not using the assembler for the join, so, if we have two assemblies for the same sample, the join will do a cross product, producing 4 outputs instead of 2 (each combination of original assembler + poslished one).

The problem is that pypolca emits the reads meta, not the assembly one, so we can't get that info easily.

In my opinion, we should update the module so it emits the meta from the assembly, not the one from the reads. The output is more closely related to the assembly than the reads, so it makes more sense. If we have that change, we can remove this whole block, because we will have the meta from the assembly here already, and will prevent the issue I'm describing.

dialvarezs: Well, I guess in this case I would just swap the inputs: make the assembly the first, because it's the primary one and leave the reads as the second input.
The other option is using meta2 in the outputs, but I think this feels like a poorer design.

Harshita-sriv: I've refactored the module to use the assembly metadata as the primary input, updated the MAG integration/tests, and verified the output structure locally. CI is now passing.

d4straub (approved):
No further comments, looks perfect now!
That was quite a journey, but a successful one ;)

dialvarezs (approved):
Hi @Harshita-sriv!

As a finishing touches I updated some variable names and moved some things around to improve semantics, and also updated output docs.
I think this is good to merge now.

You did a wonderful job! Thanks for contributing this!

Feel free to push the merge button :D

PR #1065, opened 2026-07-01
feat: remove GUNZIP_*BINS

dialvarezs:
This PR removes the use of GUNZIP on bins and unbinned contigs.

This was problematic for two reasons:
- Execution was per-bin, so this process commonly spawned around two orders of magnitude more jobs than any other process in the pipeline, flooding HPC clusters with many tiny jobs.
- The flatten -> gunzip -> merge pattern disrupted pipeline flow, because a `groupTuple()` with no size effectively "locks" the pipeline until all upstream channels are exhausted.

The requirement for this to work is that all downstream processes handle compressed files. This was covered either natively or with an explicit in-process gunzip, with two exceptions:
- CATPACK modules: solved by adding a conditional gunzip to the nf-core modules (updated in this PR).
- DAS Tool: `dastool_rename_pre` and `dastool_rename_post` handle decompression and compression, respectively.

Some minimal tweaks to Python scripts and configs were also needed to handle everything compressed.

Snapshots are mainly stable, most changes come from QUAST or processes that use the filename.

Decompressed files were not published in any way, to this change doesn't represent any change for users.

This PR also enables a path that was previously untested: CATPACK on unbinned contigs.

## PR checklist


subworkflows/local/binning_refinement/main.nf
```diff
@@ -66,29 +66,15 @@ workflow BINNING_REFINEMENT {
 
     // Prepare bins for downstream analysis (separate from unbins, add 'binner' info and group)
     // use DAS Tool as 'binner' info allowing according grouping of refined bin sets,
-    // while keeping information about original binning method in filenames and used binnames, e.g. "*-MaxBin2Refined-*.fa"
+    // while keeping information about original binning method in filenames and used binnames, e.g. "*-MaxBin2Refined-*.fa.gz"
     // (alternatively one could think of adding, for example, meta.orig_binner, if this would simplify code)
-    ch_dastool_bins_newmeta = DASTOOL_DASTOOL.out.bins
-        .transpose()
-        .map { meta, bin ->
-            if (bin.name != "unbinned.fa") {
```
jfy133, line 74: So this doesn't exist anymore?

tests/default.nf.test.snap
```diff
@@ -2041,83 +2124,83 @@
                 "SPAdes-SemiBin2-test_minigut-bins_bin2classification.names.txt:md5,dedb9199df302c607660a548e7e83ea1",
                 "SPAdes-SemiBin2-test_minigut-bins_summary.txt:md5,1666dbfcbf757029cc0e5b0bba712e00",
                 "bat_summary.tsv:md5,76a5177b83e5ccad2ca2d14f2dc7f5cc",
-                "MEGAHIT-MaxBin2-unclassified-unrefined-test_minigut.bac120.summary.tsv:md5,da268bb1219872565442f15232f9b6fd",
+                "MEGAHIT-MaxBin2-unclassified-unrefined-test_minigut.bac120.summary.tsv:md5,18b8156141c6b98440888d0df31747cf",
```
jfy133, line 2127: Should these be changing so much? The contents should be the same no? Or does it also include the file name?

workflows/mag.nf
```diff
@@ -578,7 +578,7 @@ workflow MAG {
             ch_bins_for_prokka = ch_input_for_postbinning
                 .transpose()
                 .map { meta, bin ->
-                    def meta_new = meta + [id: bin.getBaseName()]
+                    def meta_new = meta + [id: bin.getName() - ~/\.fa(sta)?(\.gz)?$/]
```
jfy133, line 581: WOuld it make sense to include all FASTA format extension variants for future proofing? (fa,fna,fas,fasta)?

Only an idea, not really relevant hrere

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1060](https://github.com/nf-core/mag/pull/1060) - Updated module tags to make them more specific (by @dialvarezs)
 - [#1061](https://github.com/nf-core/mag/pull/1061) - Speed up binning by not waiting for all mapping jobs to finish before starting (by @dialvarezs)
 - [#1063](https://github.com/nf-core/mag/pull/1063) - Run ALE with `--metagenome` and disable its per-base output by default (by @dialvarezs)
+- [#1065](https://github.com/nf-core/mag/pull/1065) - Remove usage of GUNZIP module for bins and unbinned contigs, feeding downstream processes with uncompressed files (by @dialvarezs)
```
jfy133:
```suggestion
- [#1065](https://github.com/nf-core/mag/pull/1065) - Improved efficiency by removing usage of per-bin GUNZIP module for bins and unbinned contigs and allowing gzip support for all modules (by @dialvarezs)
```

jfy133 (approved):
Makes sense to me, only a couple of queries! But I trust you've checked these :)

d4straub:
~~So now when all 6 binners+DAStool are used we will have increased the size of the results folder by 7 times the size of a unzipped metagenome? That can go into the 100s of GB. I find that inconvenient. Just wanted to point out that disadvantage. If the gain in speed/whatever is more valuable, so be it.~~
Edit: oh I am sorry, it seems that is kept zipped, just handled different. My bad!

workflows/mag.nf
dialvarezs: Tbh, `fasta.gz` is not needed either, because these come from the binners, a they always emit `.fa`/`.fa.gz`.

subworkflows/local/binning_refinement/main.nf
dialvarezs: Yeah, refined bins come directly from `RENAME_POSTDASTOOL.out.refined_bins` directly now, and that handles the splitting.

tests/default.nf.test.snap
dialvarezs: Yes, this is because of the filename change.

PR #1066, opened 2026-07-03
perf: remove remaining lockings from mag

dialvarezs:
This is (hopefully) the last of my optimization-related PRs. It removes the remaining channel "locks" in the binning -> QC -> downstream path, points where a size-less `groupTuple()` forces every upstream task to finish before anything downstream can start.

1. **seqkit / bin length filtering**
	seqkit was still locking the pipeline via the `groupTuple()` that regrouped bins to decide which pass the length filter. On top of that, I'd forgotten to drop the `.transpose()` from the previous PR, so seqkit was running per bin instead of per binner batch. This refactors that section so each binner batch runs seqkit on its whole output at once and is filtered independently.

2. **QUAST**
	The input to QUAST_BINS was regrouped with a size-less `groupTuple()`, blocking it until all binning finished. Since each entry there already has a unique key, the `groupTuple()` was just reshaping single items, so I replaced it with a plain `.map`.

3. **Bin QC -> GTDB-Tk**
	Bin QC + GTDB-Tk was another locking point: GTDB-Tk needs the QC metrics to decide which bins are evaluated for taxonomy, and it consumed the concatenated QC summaries (all bins). To operate per binner batch, I changed BIN_QC to emit the per-group, per-tool QC summaries directly, and gave GTDB-Tk's `groupTuple` a size (the number of the qc tools enabled), so each GTDB-Tk job fires as soon as its QC finishes.

Snapshot effect: fewer seqkit processes (now per binner batch, not per bin), nothing more.

Here's a timeline of a test run. It shows no more locking points (all processes are well mixing and overlapping).
[execution_timeline_2026-07-03_09-02-54.html](https://github.com/user-attachments/files/29635600/execution_timeline_2026-07-03_09-02-54.html)

We also got nice CI time cutoffs from this: on average 23% less time compared to yesterday's runs like https://github.com/nf-core/mag/actions/runs/28590406418/

## PR checklist

### DIFF
--- a/subworkflows/local/gtdbtk/main.nf
+++ b/subworkflows/local/gtdbtk/main.nf
@@ -39,7 +39,7 @@ workflow GTDBTK {
     ch_filtered_bins = ch_bins
         .join(ch_group_metrics)
         .map { meta, bins, tool_summaries ->
-            // build per-bin metrics: bin name -> {completeness: [...], contamination: [...]}
+            // build per-bin metrics: bin name -> [ [completeness, contamination], ... ] (one reading per QC tool)
             def metrics = [:]
             tool_summaries.each { tool, summary ->
                 def cols = qc_columns[tool]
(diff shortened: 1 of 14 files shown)


subworkflows/local/gtdbtk/main.nf
```diff
@@ -9,65 +9,74 @@ include { GTDBTK_SUMMARY        } from '../../../modules/local/gtdbtk_summary/ma
...
+                            entry.contamination << contamination
+                        }
+                        metrics[bin_name] = entry
+                    }
+            }
+
+            def passed = []
+            def discarded = []
+            [bins]
+                .flatten()
```
jfy133: This is an odd syntax, what's going on here?

jfy133 (approved):
Admittedly not super detailed check but I didn't see anything that jumped out at me

subworkflows/local/gtdbtk/main.nf
dialvarezs: It's a workaround if bins is not a list. So basically... if only one bin is emited by a binner.
I will dropping it, if it works in tests, then it should be good (it's basically impossible with real data).

PR #1083, opened 2026-07-25
Release 5.5.0 - Purple Penguin

dialvarezs:
### `Added`

- [#908](https://github.com/nf-core/mag/pull/908) - Add nf-test snapshot for `test_single_end` profile (by @dialvarezs)
- [#1028](https://github.com/nf-core/mag/pull/1028) - Add nf-test snapshot for `test_longread_alternatives` profile (by @dialvarezs)
- [#1029](https://github.com/nf-core/mag/pull/908) - Add nf-test snapshot for `test_hybrid` profile (by @dialvarezs)
- [#1037](https://github.com/nf-core/mag/pull/1037) - Complement usage documentation with guidance on pipeline defaults, choices and alternatives (by @d4straub and @jfy133 )
- [#1039](https://github.com/nf-core/mag/pull/1039) - Add nf-test snapshot for `test_longread` profile (started by @brovolia, finished by @dialvarezs)
- [#1041](https://github.com/nf-core/mag/pull/1041) - Refined and corrected unclear section of metromap (by @jfy133)
- [#1042](https://github.com/nf-core/mag/pull/1042) - Add nf-test snapshot for `test_default` profile (by @dialvarezs)
- [#1044](https://github.com/nf-core/mag/pull/1044) - Add new `--gtdbtk_place_species` parameter (by @dialvarezs)
- [#1047](https://github.com/nf-core/mag/issues/1007) - Add `--gtdbtk_single_job` to run GTDB-Tk classification for all bins in a single job (requested by @sarah-shah-bioinf, by @dialvarezs)
- [#1048](https://github.com/nf-core/mag/pull/1048) - Add optional PyPOLCA polishing for long-read assemblies via `--run_pypolca` (by @Harshita-sriv)
- [#1055](https://github.com/nf-core/mag/pull/1055) - Add dedicated page describing resource usage (by @jfy133)
- [#1057](https://github.com/nf-core/mag/pull/1057) - Add skip_fastqc switch (by @jorondo1)
- [#1059](https://github.com/nf-core/mag/pull/1059) - Add `--filtlong_filtering_by_shortreads` parameter to enable filtlong's short-read-based long read filtering (by @dialvarezs)
- [#1062](https://github.com/nf-core/mag/pull/1062) - Add `--bin_seqkit_stats_max_forks` parameter to cap concurrent bin-stats jobs (by @dialvarezs)
[...]

PR #1084, opened 2026-07-25
fix: conda-related fixes

dialvarezs:
Two changes to address the Conda failures / snapshot mismatches seen in #1083.
1. **Update the Prokka module.** GNU parallel `20260422+` fails every local job when `SHELL` isn't exported (`exec failed:No such file or directory`), which makes Prokka exit 2. The current module version pins `parallel=20241122` (nf-core/modules#11754). Containers ship an older parallel, hence Conda-only.
2. **Explicitly disable NanoPlot static images** (empty `-f`) in `tests/nextflow.config`. NanoPlot defaults to `--format png`, rendered via Kaleido, which needs Chrome. The runners have it, so Conda runs emit PNGs and container runs don't, causing snapshot drift.

## PR checklist

### DIFF
--- a/tests/nextflow.config
+++ b/tests/nextflow.config
@@ -10,16 +10,3 @@ params {
 }
 
 aws.client.anonymous = true // fixes S3 access issues on self-hosted runners
-
-// NanoPlot only exports static (PNG) plots when a Chrome installation is available to Kaleido,
-// which is the case for Conda (host Chrome) but not for the containers, so the published files
-// differ between profiles. Passing an empty `-f` skips them and keeps the output reproducible.
-process {
-    withName: NANOPLOT_RAW {
-        ext.args = { "-p raw_ --title ${meta.id}_raw -c darkblue -f" }
-    }
-
-    withName: NANOPLOT_FILTERED {
-        ext.args = { "-p filtered_ --title ${meta.id}_filtered -c darkblue -f" }
-    }
-}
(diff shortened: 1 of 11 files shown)


tests/nextflow.config
```diff
@@ -10,3 +10,16 @@ params {
 }
 
 aws.client.anonymous = true // fixes S3 access issues on self-hosted runners
+
+// NanoPlot only exports static (PNG) plots when a Chrome installation is available to Kaleido,
+// which is the case for Conda (host Chrome) but not for the containers, so the published files
+// differ between profiles. Passing an empty `-f` skips them and keeps the output reproducible.
```
jfy133: Wouldn't this apply anywhere there is no chrome though - could happen to users to? I sort of feel that they should maybe just be completely disabled. You get the HTML files anyway

jfy133 (approved):
LGTM, just missing CHANGELOG and OUTPUT.md updates regarding loss of nanoplot .png, if you agree with my comment

tests/nextflow.config
dialvarezs: Only applies to tests (`tests/nextflow.config`, not `nextflow.config`).
So, no outputs changes for end-user real runs, only tests.

dialvarezs: But yeah, it should happen to users without Chrome I guess....

dialvarezs: Ok, let's apply the change globally then.

PR #1086, opened 2026-07-26
Rename MaxBin2 bins inside the MAXBIN2 module

dialvarezs:
Fixes the AWS full test abort seen in run `1pFy1kkAk54Lr`:

```
DEBUG nextflow.splitter.AbstractSplitter - Creating gzip splitter for:
  .../75/a58504.../METAMDBG-MaxBin2-CAPES_S11.001.fa.gz
ERROR nextflow.extension.DataflowHelper - @unknown
java.util.zip.ZipException: Not in GZIP format
DEBUG nextflow.Session - Session aborted -- Cause: Not in GZIP format
```

## Cause

`ADJUST_MAXBIN2_EXT` renamed its staged inputs with `mv`. Nextflow stages inputs as symlinks, so `mv` renames the link, not the data. On Fusion/S3 the published object ends up as 123 bytes of plain text holding the link target, instead of the 64 KB gzip that MAXBIN2 wrote.

Tasks consuming those bins resolve the link through the Fusion mount, which is why `SPLIT_FASTA` and `SEQKIT_STATS` passed. The `splitFasta` building the contig-to-bin map runs on the head node and reads the raw bytes, so it fails.

The published MaxBin2 bins were already broken before this run; the `splitFasta` added in #1065 is just what made it fatal.

## Changes

- The rename now happens inside `MAXBIN2` itself, as a patch on the nf-core module, so it acts on files the task created rather than on staged symlinks. `ADJUST_MAXBIN2_EXT` is removed — one less local module.
  - Renaming *after* `gzip` keeps the archives byte-identical to what the pipeline produced before, since the original filename lives in the gzip header. Only the versions list and task count change in the snapshots.
- `TIARA_CLASSIFY` had the same pattern (`find -L ... -exec mv`) — `-L` only affects how `find` walks the tree, `mv` still moves the link. It now copies. `cp` dereferences by default for non-recursive copies, so no flag is needed.
- `TIARA_CLASSIFY`: now that the staged bins stay in place, the `find` is restricted to depth 1 and to an exact `${bin}.fa.gz` match. The previous `${bin}*` glob also matched bin `.10` while processing bin `.1`.

The other `mv` calls in `modules/local/` are fine — they either materialise the file first (`dastool_rename_pre`, `dastool_rename_post`) or act on files created inside the task (`metabinner_tooshort`, `bowtie2_assembly_align`).

## Testing

Snapshots regenerated for the six profiles that run MaxBin2. `test_minimal` and `test_alternatives` set `skip_maxbin2 = true` and are unaffected.

The AWS full test is the real check here, since local runs resolve the symlinks either way.


jfy133 (commented):
I'm not sure I like this, this will be very expensive HDD wise simply for a very highly specific filesystem...

Mulling over it I have different suggestions (not that I'm particularly happy with any of them 😓 ):

1. Symlink the staged symlink to the new name instead of mv/cp? I wonder if that would work?
2. Using Nextflow's `stageAs` somehow on downstream modules? (might be a lot of modules though)
2. Use `nf-core patch` the upstream MAXBIN2 module in the pipeline repo to use the rename the output within the MAXBIN2 itself (I think this might be the safest to be honest). This would then drop the need for the ADJUST_* module entirely.

Also if we are editing these local modules maybe makes sense to convert them to topics at the same time?

The TIARA module change I do not understand however - the files being moved SHOULD just be files generated by the tool itself - there should not be any issues with the weird-fusion filesystem link things? Maybe we need to change how the module works?

dialvarezs:
It's not that bad I guess. It happens only once per MaxBin2 execution (duplicating the files), and bin files are not so heavy anyway. Compared to other processes writing tens of GB this didn't seem so much additional storage stress.
But patching MaxBin module locally definitively is a more elegant solution.

Regarding Tiara, it's just for consistency, but considering that nothing depends on Tiara, we can leave the mv (or at least a `cp` without `-L`).

This was the easiest solution that came to my mind to fix the full test issue.

dialvarezs:
Update:
- Now the rename happens inside the MaxBin2 module (patched), removing the local module completely
- Changed the `cp -L` in the Tiara module to simply `cp`
- Will leave topics for later

Waiting for snapshots now...

CHANGELOG.md
```diff
@@ -61,6 +61,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1076](https://github.com/nf-core/mag/pull/1076) - Add missing PyPOLCA citations to README, `CITATIONS.md` and the pipeline citation/bibliography text (by @dialvarezs)
 - [#1078](https://github.com/nf-core/mag/pull/1078) - Sort bins before GTDB-Tk classification so order-sensitive outputs are reproducible across environments (by @dialvarezs)
 - [#1084](https://github.com/nf-core/mag/pull/1084) - Update Prokka nf-core module, which pins GNU parallel to fix Prokka failures with Conda (by @dialvarezs)
+- [#1086](https://github.com/nf-core/mag/pull/1086) - Rename MaxBin2 bins inside the `MAXBIN2` module via a patch instead of the `ADJUST_MAXBIN2_EXT` local module, which emitted dangling symlinks rather than bin contents on remote filesystems (by @dialvarezs)
+- [#1086](https://github.com/nf-core/mag/pull/1086) - Copy instead of move staged bins in `ADJUST_MAXBIN2_EXT` and `TIARA_CLASSIFY`, which emitted dangling symlinks rather than bin contents on remote filesystems (by @dialvarezs)
```
jfy133:
```suggestion
- [#1086](https://github.com/nf-core/mag/pull/1086) - Copy instead of move staged bins in `TIARA_CLASSIFY`, which emitted dangling symlinks rather than bin contents on remote filesystems (by @dialvarezs)
```

jfy133 (approved):

PR #1088, opened 2026-07-27
Update Prokka, Prodigal and PyDamage modules (fixes conda snapshot drift)

dialvarezs:
## Why

The `conda` shards on #1083 fail on four profiles (`default`, `test_hybrid`, `test_longreadonly`, `test_single_end`) while `docker`/`singularity` pass. Not resolution drift: the conda md5s are stable across CI runs and reproduce exactly on a local machine.

Root cause, isolated outside the pipeline: `prodigal -p meta` (the args mag uses) gives different output under conda than under the old mulled biocontainer, even though both report prodigal 2.6.3 + pigz 2.6 — the image was built with different dependencies than a fresh conda solve.

```
conda                                        → 9eb376c2… / a0ccda43…
old container (mulled biocontainer)          → db3defda… / 53c5cc98…   ✗
new container (wave, nf-core/modules#12432)  → 9eb376c2… / a0ccda43…   ✓
```

Same story for Prokka (old `prokka_openjdk` image didn't match its `environment.yml`) and PyDamage (frozen numerics deps → 6 of 27 `*_pydamage_bin_results.tsv` differed).

## What

- Update `prokka` (nf-core/modules#12431), `prodigal` (#12432) and `pydamage` (#12433). All now carry conda lock files and wave containers built from their `environment.yml`.
- Prokka 1.14.6 → 1.15.6, so its outputs change. Pipeline snapshots regenerated with `docker`.
- Prodigal and PyDamage now report versions through the `versions` topic channel, so the explicit `mix()` calls are dropped.
- Run GTDB-Tk with `cpus = 1` on `test_single_end`: its marker identification and pplacer placement vary with thread count, which made `pplacer.out` and `bac120.filtered.tsv` depend on the machine running the test.

## Verification

Locally, 4 profiles, snapshots regenerated with `docker` then re-run with `conda`:

| profile | conda before | conda after |
|---|---|---|
| default | ❌ drift (same md5s as CI) | ✅ |
| test_hybrid | ❌ | ✅ |
| test_longreadonly | ❌ | ✅ |
| test_single_end | ❌ | ✅ |

The regenerated docker md5s are exactly the values conda was already producing, i.e. the two engines are now aligned.

Two caveats worth knowing:

- The GTDB-Tk mismatch never reproduced on my 8-core box (only on the 4-core runner), so `cpus = 1` is a reasoned fix, not one I could verify locally. The conda shards will tell.
- Conda is not run on this PR (the workflow only enables it when the base is `main`), so this is exercised when #1083 runs again.

Unrelated: the `singularity` shard 6 failure on #1083 is an image pull timeout (`singularity pull … status: 143`), not drift.


mashehu (approved):

dialvarezs:
Thanks @mashehu!
BTW, I'm adding the pydamage update here too, so I won't be creating a new PR basically for the same purpose.

PR #1089, opened 2026-07-27
Increase CONVERT_DEPTHS memory in the full test config

dialvarezs:
## Problem

The AWS full-size test run `1VGAEYl9hrOScI` aborted on a `CONVERT_DEPTHS` task that exited with status 174 after sitting in `RUNNING` for 2h04:

```
Caused by:
  Essential container in task exited
Command exit status:
  174
Command output:
  (empty)
```

`.command.out` was never written and `.command.err` contained only the Fusion startup banner, so the script produced nothing. The five sibling `CONVERT_DEPTHS` tasks were also still running when the session aborted.

## Changes

Raise `CONVERT_DEPTHS` from 1 GB to 2 GB in `conf/test_full.config`, so the starting allocation has headroom rather than relying on retry escalation.

The script only streams (`gunzip`, a one-line `awk`, then `bioawk` passes), so peak RSS is small, but the container also hosts the Fusion client and 1 GB left nothing spare. 2 GB matches the tier already used by the other small processes in this config.

## On exit code 174

This PR originally also added 174 to the retriable exit-code lists. Dropped per review — and the docs confirm that was the right call. Seqera documents 174 as `ErrorExitCode`, *"Fusion I/O error, application-level input/output error"*, explicitly not an out-of-memory condition:

> The `sysexits.h` standard uses exit code 74 for "input/output error" and reserves 150-199 for application use. In Fusion's context, 174 means "application input/output error".

The listed causes are all FUSE/mount-level — failing to start or stop the FUSE process, errors during filesystem shutdown or unmount. The OOM codes inherited from the nf-core template are a different category, so 174 does not belong there, in this pipeline or upstream.

That does mean this PR does not address the two-hour stall itself, which still looks like a Fusion or Spot issue worth tracking separately.

Source: [Fusion error codes and exit messages](https://docs.seqera.io/fusion/troubleshooting/error-codes-exit-messages#exit-codes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,7 +65,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1086](https://github.com/nf-core/mag/pull/1086) - Copy instead of move staged bins in `TIARA_CLASSIFY`, which emitted dangling symlinks rather than bin contents on remote filesystems (by @dialvarezs)
 - [#1086](https://github.com/nf-core/mag/pull/1086) - Match bin filenames exactly in `TIARA_CLASSIFY` so that e.g. bin `.1` no longer also picks up bin `.10` (by @dialvarezs)
 - [#1087](https://github.com/nf-core/mag/pull/1087) - Use non-mutating `toSorted` instead of in-place `sort` on shared channel items, which could cause `ConcurrentModificationException` failures (by @dialvarezs)
-- [#1088](https://github.com/nf-core/mag/pull/1088) - Retry on exit status 174, which was missing from the retriable Fusion exit codes and aborted full-size runs on Spot reclamation (by @dialvarezs)
+- [#1089](https://github.com/nf-core/mag/pull/1089) - Retry on exit status 174, which was missing from the retriable Fusion exit codes and aborted full-size runs on Spot reclamation (by @dialvarezs)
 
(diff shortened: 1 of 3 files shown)


mashehu (commented):
wouldn't it be better to increate the memory for these processes in the first run instead of doing it in a sequential one?

dialvarezs:
Fair, but I'm not sure sure how much memory it will use. It's just a bioawk script. I can try doubling (2 GB instead of 1).

What do you think of doing both things? Memory increase + retry on 174? Or is it a bad idea to retry on 174.

mashehu:
I just would avoid changing too many things from the template (but am biased there 🙂). 
Also that it errors means the config is incorrect, so better to fix that

jfy133:
I agree, in this case just bump starting value to the CONVERT module, and if we can confirm 174 is truely OOM error code(!), should go into the template - means less template merge conflicts.

Maybe we can ask a seqeran to verify that `174` is a valid exit code for OOM errors with Fusion (@FriederikeHanssen @edmundmiller maybe?)

mashehu:
see https://docs.seqera.io/fusion/troubleshooting/fusion-snapshots#exit-code-175-checkpoint-dump-failed

dialvarezs:
Reverted the exit code changes, just increasing memory now.

jfy133 (approved):

PR #1102, opened 2026-08-13
Add new nf-core/mag ancient DNA and v5 citations

jfy133:
## PR checklist

- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/mag/tree/main/docs/CONTRIBUTING.md)


jfy133:
@nf-core-bot fix linting

d4straub (approved):
Looks fine to me.
Congratulations!

dialvarezs (approved):
Congratulations!!!

dialvarezs:
Hey @jfy133 @d4straub @prototaxites
I added the v5 citation too, but I guess we should discuss how exactly are we are going to recommend citing?
aDNA has a clear path, both v5 is a general update over the original one. I would recommend citing both, but James has a point on that people won't follow that always (specially when constrained by the number of references).
Do you have any suggestions about this?

README.md
```diff
@@ -10,7 +10,7 @@
 [![Nextflow](https://img.shields.io/badge/version-%E2%89%A526.04.0-green?style=flat&logo=nextflow&logoColor=white&color=%230DC09D&link=https%3A%2F%2Fnextflow.io)](https://www.nextflow.io/)
 [![nf-core template version](https://img.shields.io/badge/nf--core_template-4.1.0-green?style=flat&logo=nfcore&logoColor=white&color=%2324B064&link=https%3A%2F%2Fnf-co.re)](https://github.com/nf-core/tools/releases/tag/4.1.0)
 [![Cite Zenodo](http://img.shields.io/badge/DOI-10.5281/zenodo.3589527-1073c8?labelColor=000000)](https://doi.org/10.5281/zenodo.3589527)
-[![Cite Publication](https://img.shields.io/badge/cite%20-Publication-orange)](https://doi.org/10.1093/nargab/lqac007)
+[![Cite Publication](https://img.shields.io/badge/cite%20-Publication-orange)](https://doi.org/10.1093/bioinformatics/btag628)
```
d4straub, line 13: So thats now only pointing towards the most recent paper. (Not criticizing, but that sets a choice sort of...). Not sure what source people take for their papers though. The bottom of the README I hope :D

d4straub:
For me, mentioning both papers (original + v5) is fine, the most recent publication first is also fine (this will be likely the one that will be cited most in case only one citation is chosen). However, I am on both papers, so I dont care that much :D
@skrakau, @ggabernet  and @HadrienG  might think otherwise (only on the original paper)?

HadrienG:
> For me, mentioning both papers (original + v5) is fine, the most recent publication first is also fine (this will be likely the one that will be cited most in case only one citation is chosen). However, I am on both papers, so I dont care that much :D @skrakau, @ggabernet and @HadrienG might think otherwise (only on the original paper)?

Personally I think it is acceptable, I have done similarly in the past for other software I have developped. (won't complain either if you keep the paper I'm on too 😛 )

skrakau:
Using the most recent one in the badges is fine I think. But I would still clearly recommend citing both papers, something like "If you use nf-core/mag in your work, please cite the original paper and, where applicable, the follow-up paper on v5:". The order could stay. If people have to choose and used the newer version, they would still omit the original one I would guess. What do you think?

jfy133:
Note the cite publication button is a custom one for mag, there is no reason not to add multiple!

But I think generally we are all generally in-line: recommend both ideally. But maybe if only one - prioritise v5 if using long-read? (or something along the lines: of pick the one that better matches the functionality you used - e.g. if you mostly used long-reads, or the more recent binners use v5 else v1-4)? Or is that too complicated?

dialvarezs:
Hey @HadrienG, @skrakau, @d4straub, @jfy133, thanks for your comments!

Sounds like we all agree the main recommendation is to cite both.

Regarding of what we should do when the user only wants to cite one:

> But maybe if only one - prioritise v5 if using long-read?

I'd rather not tie it to the features used. Long-read is the obvious case, but most of what the v5 paper covers isn't in the original: the newer binners, the non-bacterial taxonomy options, bin refinement, etc. So "pick the one matching the functionality you used" ends up a judgement call for most users, and I suspect v5 would be the answer for nearly any run with a current version of the pipeline anyway.

Using the version as a proxy is simpler to understand: v5 onwards -> new publication, up to v4 -> original. And it's what I think most people do when citing some software: just pick the newest one, unless they know which paper matches what they actually did (in our case, aDNA would be an example of that).

On the badge: two badges with the same "Cite Publication" label could be confusing (even if we put a slightly different label), so if we don't want to prioritize I think the best is to simply remove it. I personally would keep it with the new one, because that's the one that resembles the current state of mag better, but with the general recommendation of citing both in the README.

For the wording, I suggest leaving it like this:

> If you use nf-core/mag, please cite both nf-core/mag publications:
>
> [v5 publication]
>
> [original publication]
>
> If you are limited in the number of references, cite according to the version you used: v5 onwards -> Alvarez Saravia et al. 2026, up to v4 -> Krakau et al. 2022.

jfy133:
> Hey @HadrienG, @skrakau, @d4straub, @jfy133, thanks for your comments!
> 
> Sounds like we all agree the main recommendation is to cite both.
> 
> Regarding of what we should do when the user only wants to cite one:
> 
> > But maybe if only one - prioritise v5 if using long-read?
> 
> I'd rather not tie it to the features used. Long-read is the obvious case, but most of what the v5 paper covers isn't in the original: the newer binners, the non-bacterial taxonomy options, bin refinement, etc. So "pick the one matching the functionality you used" ends up a judgement call for most users, and I suspect v5 would be the answer for nearly any run with a current version of the pipeline anyway.
> 
> Using the version as a proxy is simpler to understand: v5 onwards -> new publication, up to v4 -> original. And it's what I think most people do when citing some software: just pick the newest one, unless they know which paper matches what they actually did (in our case, aDNA would be an example of that).
> 
> On the badge: two badges with the same "Cite Publication" label could be confusing (even if we put a slightly different label), so if we don't want to prioritize I think the best is to simply remove it. I personally would keep it with the new one, because that's the one that resembles the current state of mag better, but with the general recommendation of citing both in the README.
> 
> For the wording, I suggest leaving it like this:
> 
> > If you use nf-core/mag, please cite both nf-core/mag publications:
> >
> > [v5 publication]
> >
> > [original publication]
> >
> > If you are limited in the number of references, cite according to the version you used: v5 onwards -> Alvarez-Saravia et al. 2026, up to v4 -> Krakau et al. 2022.

Works for me!

CITATIONS.md
```diff
@@ -8,6 +8,22 @@
 
 > Di Tommaso P, Chatzou M, Floden EW, Barja PP, Palumbo E, Notredame C. Nextflow enables reproducible computational workflows. Nat Biotechnol. 2017 Apr 11;35(4):316-319. doi: 10.1038/nbt.3820. PubMed PMID: 28398311.
 
+## Pipeline
+
+If you use nf-core/mag please cite:
```
jfy133: I would add the guidance of both here too

PR #1105, opened 2026-08-18
Fix bin path corruption when a process emits a single file

dialvarezs:
## Description

When a `path("dir/*.fa.gz")` output has no `arity` and the glob matches exactly one file, Nextflow emits a bare `Path` instead of a one-element list. `Path` is `Iterable` over its **path segments**, so `findAll`, `toSorted`, `[0]` and friends iterate the directory names instead of the bins. What survives is a relative one-segment path, which resolves against `launchDir` and gives a broken symlink.

This made GTDB-Tk fail with `There are N paths in the batchfile which do not exist` for any group left with a single prokaryotic bin after Tiara. Groups with two or more bins were fine, which is why it went unnoticed.

## Changes

- Added `arity` to the glob bin outputs in `tiara_classify`, `dastool_rename_pre`, `dastool_rename_post` and `metabinner_bins`, so the channel is a list from the start. Optional outputs use `'0..*'`, because `'1..*'` with `optional: true` fails outright when the glob matches nothing.
- Guarded the consumers with the `[bins].flatten()` idiom already used elsewhere in the pipeline, since bins can also come from `modules/nf-core/*`: GTDB-Tk bin filtering (the reported failure) and CheckM v1 bin sorting (same bug, latent because `run_checkm` is off by default).
- Fixed `reads.toSorted { files -> files[0].getName() }` in the co-assembly grouping. With single-end data `files[0]` returned the first path segment, the same for every sample, so the sort key collapsed. Ordering only, no broken paths.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### `Fixed`
 
 - [#1098](https://github.com/nf-core/mag/issues/1098) - Fix invalid `--gunc_database_type` options to match the values accepted by `gunc download_db` (reported by @cdiener, fix by @dialvarezs)
+- [#1105](https://github.com/nf-core/mag/pull/1105) - Fix corrupted bin paths when a process emits a single file: `java.nio.file.Path` is `Iterable` over its path segments, so list operations on an unwrapped output silently iterated the path components (by @dialvarezs)
 
 ### `Dependencies`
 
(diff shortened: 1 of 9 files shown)


prototaxites (approved):
Nice catch! Do any of the nf-core modules need fixing?

jfy133 (approved):

dialvarezs:
> Nice catch! Do any of the nf-core modules need fixing?
All binners probably. But I guess it's very uncommon to get a single bin from them.
On the other side, with Tiara it can happens more frequently.

PR #1109, opened 2026-09-08
Fix checkm_download_url blocking parameter validation unconditionally

erikrikarddaniel:
## PR checklist

- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/mag/tree/main/docs/CONTRIBUTING.md)

## Description

`checkm_download_url`'s schema entry declares `"exists": true` on a non-null default (a Zenodo URL). nf-schema validates that unconditionally at pipeline startup, regardless of whether `--run_checkm` is actually enabled — unlike every other optional database param in this section (`checkm2_db`, `gtdb_db`, `cat_db`, ...), which default to `null` and are only validated when supplied.

That means **every single nf-core/mag run has a silent, unconditional dependency on Zenodo being reachable**, whether or not CheckM1 is used at all. Hit this independently on two different networks while trying to run `--run_checkm2` (not `--run_checkm`) during unrelated work — Zenodo was briefly unreachable from both, and the pipeline failed at parameter validation before a single process ran.

Reproduces deterministically, independent of Zenodo's actual availability:

```
nextflow run . -profile test_minimal --checkm_download_url /does/not/exist
```

### Fix

Drop `exists: true`. The real existence check already happens correctly at the point of actual use — `file(params.checkm_download_url, checkIfExists: true)` in `subworkflows/local/bin_qc/main.nf`, inside the `run_checkm` branch — so nothing is lost when `--run_checkm` is genuinely enabled with a bad URL; it just stops being checked when the feature isn't even on. `format: file-path` is kept (harmless, preserves documentation value; doesn't affect the check, confirmed by testing with and without it — the `exists` key is what triggers validation).

### Test

`tests/test_checkm_download_url_validation.nf.test` uses `-preview` (skips executing any process, but still runs `PIPELINE_INITIALISATION`'s parameter validation) so the regression test exercises only what this bug affects, as cheaply as possible — no real task execution needed.

https://claude.ai/code/session_01NJ75Ui2JZjJwJ7T7o5acvg


tests/test_checkm_download_url_validation.nf.test
dialvarezs, line 1: I don't think the regression test is actually useful here. This won't fail unless someone adds the nf-schema validation again.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [#1098](https://github.com/nf-core/mag/issues/1098) - Fix invalid `--gunc_database_type` options to match the values accepted by `gunc download_db` (reported by @cdiener, fix by @dialvarezs)
 - [#1105](https://github.com/nf-core/mag/pull/1105) - Fix corrupted bin paths when a process emits a single file: `java.nio.file.Path` is `Iterable` over its path segments, so list operations on an unwrapped output silently iterated the path components (by @dialvarezs)
+- [#NN](https://github.com/nf-core/mag/pull/NN) - Fix `--checkm_download_url` blocking parameter validation for every run (even with `--run_checkm` off) whenever Zenodo isn't reachable, by dropping its unconditional `exists` schema check (by @erikrikarddaniel)
```
dialvarezs:
```suggestion
- [#1109](https://github.com/nf-core/mag/pull/1109) - Fix `--checkm_download_url` blocking parameter validation for every run (even with `--run_checkm` off) whenever Zenodo isn't reachable, by dropping its unconditional `exists` schema check (by @erikrikarddaniel)
```

dialvarezs (changes requested):

dialvarezs (approved):
LGTM!

jfy133 (approved):

Issue #413, opened 2023-03-27
Add dereplication with dRep


erikrikarddaniel:
### Description of feature

dRep takes a set of genomes, with CheckM data, and dereplicates them to produce a set of non-overlapping genomes at a specified ANI. As this is basically just pointing to certain MAGs as the representatives of clusters, the output could possibly be summarised by a column in the bin summary table: `dRep`. If the column says e.g. `95` and hence indicates the ANI, one could potentially run dRep multiple times. There is no existing nf-core module.

prototaxites:
I have a module for Galah that I wrote for a personal pipeline processing the output of mag, as well as a process that takes the busco_summary.tsv file and converts it to the format required by Galah to use the completeness/contamination information.

I had planned to add them to mag at some point, but haven't found time yet.

jfy133:
:tada: awesome!

prototaxites:
https://github.com/nf-core/modules/pull/3666

jfy133:
Requires: 

- https://github.com/nf-core/modules/issues/5591
- https://github.com/nf-core/modules/issues/5590

erikrikarddaniel:
I asked Claude to give me a plan for how this could be implemented:

Sketching out what implementing this via [Galah](https://github.com/wwood/galah) would actually look like, building on what @jfy133 and @prototaxites discussed above. Both the Galah module and the CheckM/CheckM2-format QC input it needs already exist in nf-core/modules, so nothing is blocked upstream anymore.

## The core design question

DAS Tool and Galah solve two *different* problems that get conflated if you're not careful:

- **DAS Tool** (existing) operates *within one sample/assembly*: multiple binners propose different bin sets for the same underlying genome; DAS Tool picks the best single bin definition per genome, per sample.
- **Galah** (new) should operate *across the whole study*: the same genome/species can be independently recovered from multiple samples. Dereplication clusters those near-duplicates and picks one representative, so downstream steps (GTDB-Tk, Prokka, MetaEuk, CAT/BAT) don't redundantly process near-identical genomes multiple times.

[...]

PR #1110, opened 2026-09-09
Add optional study-wide bin dereplication with Galah

erikrikarddaniel:
## PR checklist

- [x] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/mag/tree/main/docs/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/mag _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository. (not needed -- no new fixtures, reused existing modules-testdata genomes)

## Description

Implements the dereplication feature discussed in #413.

[Galah](https://github.com/wwood/galah) clusters bins by average nucleotide identity (ANI) across the _whole study_ and picks one representative genome per cluster, so that near-identical genomes recovered independently from multiple samples aren't redundantly taxonomically classified or annotated multiple times. This is different from DAS Tool's refinement: DAS Tool picks the best bin definition among multiple binners for the _same_ sample/assembly; dereplication clusters bins _across_ samples that likely represent the same organism.

Off by default (`--dereplicate`). When on:

- GTDB-Tk, the CAT/BAT bin-classification leg, and (unless `--dereplicate_annotate_all`) Prokka run on cluster representatives only.
- Representative taxonomy is propagated to the rest of each cluster in `bin_summary.tsv`, with new `dereplication_representative`/`dereplication_is_representative` columns and `..._propagated` flags so a propagated assignment is never mistaken for one independently determined for that genome.
- Only bins passing `--dereplicate_min_completeness`/`--dereplicate_max_contamination` (default 50%/10%) are dereplicated; ANI estimates on poor-quality bins aren't trustworthy, so failing bins are excluded from clustering rather than force-assigned. `--dereplicate_ani` (default 95%, species-level) controls the clustering threshold itself.
- Leftover unbinned-contig pseudo-bins are excluded from dereplication entirely -- they aren't real genomes, and CheckM2 doesn't reliably produce QC for them.

### A Galah bug found and fixed along the way

Galah 0.4.2 panics (`index out of bounds`) instead of erroring cleanly when zero genomes pass its quality thresholds. Filed as [wwood/galah#75](https://github.com/wwood/galah/issues/75) and fixed upstream within the hour -- not released yet, though, so `DEREPLICATION` counts qualifying genomes itself before invoking Galah and skips it gracefully (with a warning) rather than crash the pipeline when nothing qualifies. Covered by a dedicated test case.

### Testing

- Two `nf-test` cases for the `DEREPLICATION` subworkflow: normal clustering with metadata recovery, domain/unbinned-pseudo-bin exclusion, and CheckM2/Galah filename-format reconciliation; and the zero-qualifying-genomes guard.
- Full pipeline runs (`-profile test,docker`, locally) exercising `GALAH` → `GTDBTK`/`CATPACK`/`PROKKA` representative-only routing → `BIN_SUMMARY` taxonomy propagation end-to-end.
- Validated against a real production run (~5,300 DAS-Tool-refined bins across a real multi-sample study, dereplicating to ~500 representatives) -- caught several real integration bugs along the way that none of the above would have (see individual commit messages).


subworkflows/local/dereplication/main.nf
```diff
@@ -0,0 +1,106 @@
...
+            }
+
+        ch_galah_routed.skip.subscribe {
+            log.warn("[nf-core/mag] Dereplication: no bins passed --dereplicate_min_completeness ${params.dereplicate_min_completeness} / --dereplicate_max_contamination ${params.dereplicate_max_contamination}; skipping Galah for this run (works around a Galah crash on zero qualifying genomes, see https://github.com/wwood/galah/issues/75).")
+        }
+
+        GALAH(ch_galah_routed.cluster)
+    }
+
+    // Recover each representative's original per-sample metadata by
```
dialvarezs: Can we make the comments less verbose?
Opus 5 loves to overexplain the same logic that is actually clear in the code 😅.

subworkflows/local/dereplication/tests/main.nf.test
dialvarezs, line 1: Can we actually add the tool to one of the full pipeline profiles? This test is not triggered by ci, only manually.

docs/usage.md
```diff
@@ -483,6 +483,29 @@ In this case, DAS Tool has not necessarily failed but was unable to complete the
 
 If you are regularly getting such errors, you can try reducing the `--refine_bins_dastool_threshold` value, which will modify the scoring threshold defined in the [DAS Tool publication](https://www.nature.com/articles/s41564-018-0171-1).
 
+### Reducing downstream runtime with `--postbinning_input`
```
dialvarezs, line 486: This new section was intended?
If it was, I guess it deserves an entry on the changelog.

main.nf
```diff
@@ -223,6 +223,12 @@ params {
     max_multiqc_email_size: String = '25.MB'
     multiqc_methods_description: Path?
     generate_bigmag_file: Boolean
+    dereplicate: Boolean
+    dereplication_tool: String = 'galah'
+    dereplicate_annotate_all: Boolean
+    dereplicate_min_completeness: Float = 50
+    dereplicate_max_contamination: Float = 10
+    dereplicate_ani: Float = 95
```
dialvarezs, line 230: IIRC we discussed about selecting the dereplication target: "all" / "group" / "sample".
We should have that option, because in some cases would make more sense dereplicate in a more granular way.

CITATIONS.md
```diff
@@ -90,6 +90,10 @@
 
   > Garrison E, Marth G. Haplotype-based variant detection from short-read sequencing. arXiv preprint arXiv:1207.3907 [q-bio.GN] 2012
 
+- [Galah](https://doi.org/10.5281/zenodo.13637856)
```
dialvarezs, line 93: Add the cite in the subworkflow too so it appears in the MultiQC report.

dialvarezs (commented):
Quick first review for now.

erikrikarddaniel:
Ran an independent `/code-review` pass over this PR after opening it (3 reviewer agents + a verification pass) and pushed fixes for what it found (018896e...25d2c97e):

- **Critical**: `DEREPLICATION`'s zero-qualifying-genomes fallback left `ch_dereplicated_bins` empty instead of falling back to the full bin set, so GTDB-Tk/CAT-BAT/Prokka silently produced *nothing* for the whole run whenever every bin failed the quality threshold -- which happens for real against this pipeline's own `test_minigut` CI dataset. Fixed to pass every bin through unclustered in that case, matching what the docs already claimed happens. Verified end-to-end with `-profile test,docker --run_checkm2 --dereplicate`: all three downstream tools now produce real, non-empty output.
- CAT/BAT's bins leg was silently losing eukaryotic bins whenever `--dereplicate` was on (unlike GTDB-Tk, CAT/BAT normally classifies every domain; `DEREPLICATION` excludes eukaryotic bins from clustering, but that exclusion was leaking into CAT/BAT's input too). Fixed by mixing them back in.
- No validation on `--dereplication_tool`: an invalid value silently no-op'd clustering and crashed later with an opaque error. Added a schema enum + fail-fast check, matching `postbinning_input`'s existing pattern.
- Vectorized a row-at-a-time taxonomy-propagation loop in `combine_tables.py` (verified byte-identical output against the old version first).

Updated the nf-test case that was asserting the old (buggy) empty-output behavior for the zero-qualifying-genomes path.

subworkflows/local/dereplication/tests/main.nf.test
erikrikarddaniel: I agree. Being done right now.

subworkflows/local/dereplication/main.nf
erikrikarddaniel: Being done.

docs/usage.md
erikrikarddaniel: Being done.

CITATIONS.md
erikrikarddaniel: Yes, being done.

main.nf
erikrikarddaniel: Yes we did, and I basically agree but wanted to keep things simple with this first version. Happy to backtrack there if you feel it's important to have from the start.

CITATIONS.md
erikrikarddaniel: Traced it: this module's own `meta.yml` has `doi: "10.1111/NODOI"` -- nf-core/modules' placeholder for a missing DOI -- even though Galah does have a real one (its Zenodo release). That's what's blocking it from surfacing automatically. Fixed at the source: nf-core/modules#12911.

subworkflows/local/dereplication/main.nf
erikrikarddaniel: Trimmed -- kept the genuinely non-obvious bits (Nextflow scoping quirks, the Galah upstream-crash workaround, the CheckM2/Galah filename mismatch) and cut everything that just restated what the code already says.

subworkflows/local/dereplication/tests/main.nf.test
erikrikarddaniel: Added `conf/test_dereplicate.config` + `tests/test_dereplicate.nf.test`, wired into the pipeline's CI matrix like the other test_* profiles. It reuses the default test dataset with `--run_checkm2 --dereplicate` layered on -- every bin there fails the quality threshold, so it regression-tests exactly the zero-qualifying-genomes fallback path (see the earlier code-review-fixes commit); real Galah clustering is covered separately by the subworkflow-level nf-test with real genomes chosen to pass the threshold.

docs/output.md
```diff
@@ -777,6 +777,24 @@ GUNC will be run if specified with `--run_gunc` as a standalone, unless CheckM i
...
+
+Dereplication is off by default. Enable it with `--dereplicate` (requires `--run_checkm2` and cannot be combined with `--skip_binqc`, since Galah picks representatives using CheckM2 quality estimates). When enabled, GTDB-Tk and the CAT/BAT bin-classification step run on cluster representatives only, and their taxonomic assignment is copied to the rest of each cluster in `bin_summary.tsv` (see [below](#additional-summary-for-binned-genomes)) rather than left blank. Prokka annotation also runs on representatives only by default; set `--dereplicate_annotate_all` to annotate every bin regardless of cluster membership, e.g. for pan-genome analyses that need per-strain gene content. MetaEuk (eukaryotic bins) and the CAT/BAT unbinned-contig classification leg are unaffected by dereplication, since neither of those operates on the set of bins being dereplicated.
```
prototaxites: Why do you require CheckM2? Galah should be able to read any arbitrary TSV of completeness/contam scores?

docs/usage.md
```diff
@@ -483,6 +483,29 @@ In this case, DAS Tool has not necessarily failed but was unable to complete the
...
+
+This is most worth enabling for multi-sample studies where the same organism is likely to be recovered independently in several samples; for single-sample runs, or studies where samples are unlikely to share organisms, dereplication has little to cluster and can safely be left off.
+
+Only bins passing `--dereplicate_min_completeness` (default 50%) and `--dereplicate_max_contamination` (default 10%) are dereplicated; bins below/above those are excluded from clustering entirely rather than being force-assigned to a cluster, since ANI estimates on poor-quality bins aren't trustworthy. If every bin in a study happens to fail that threshold, dereplication is skipped for that run with a warning rather than failing the pipeline.
+
+The ANI clustering threshold itself is `--dereplicate_ani` (default 95%, species-level). Prokka annotation runs on cluster representatives only by default; set `--dereplicate_annotate_all` to annotate every bin regardless of cluster membership, e.g. for pan-genome analyses that need per-strain gene content.
```
prototaxites, line 505: Is 95% a sensible default? Not 97%?

subworkflows/local/dereplication/main.nf
```diff
@@ -0,0 +1,105 @@
...
+    // scoping reason as GALAH.out.* below.
+    ch_fallback_bins = channel.empty()
+
+    if (params.dereplication_tool == "galah") {
+        ch_bins_for_galah = ch_bins_flat
+            .map { _meta, bin -> bin }
+            .collect()
+            .map { bins -> [[id: 'study'], bins] }
+
+        // CheckM2's Name column has no extension; Galah looks entries up by
```
prototaxites: I did report this bug or something like it to the Galah folk and I believe it got fixed, but may not be in a release? Could be worth asking them to tag a new one.

prototaxites:
I’ll take a proper look at the implementation tomorrow but the above are my initial thoughts reading the documentation

docs/output.md
erikrikarddaniel: I agree, an oversight.

docs/usage.md
erikrikarddaniel: 95% is GTDB's species criterion and Galah's default. For prokaryotes I think this makes a lot of sense. It is based on the observation that there's a trough of sequence identity in many prokaryotic groups around 95%.

subworkflows/local/dereplication/main.nf
erikrikarddaniel: We also reported a bug yesterday that got fixed right away, but not in any release yet. Claude found yours, also newer than any release. We have built around it so it's not urgent. The author seems very quick though.

docs/output.md
erikrikarddaniel: Fair point -- not just an example, a real unnecessary restriction. Traced it: Galah's own module already supports CheckM natively (--checkm-tab-table), and mag already builds a study-wide checkm_summary the same way it does checkm2_summary (CHECKM_QA already runs with --tab_table, exactly the format Galah's checkm-rs dependency expects -- verified against its source). Now accepts --run_checkm2 and/or --run_checkm (CheckM2 preferred when both are enabled), with a dedicated nf-test case using a real-format CheckM fixture to prove the path actually works, not just reachable in theory. True arbitrary-TSV support would need a bigger adapter (the drep-style genome_info format Galah also supports isn't something mag produces anywhere) -- happy to scope that as a fast-follow if there's a concrete use case, but this covers every QC tool mag itself already runs.

erikrikarddaniel: Has been corrected now, at least so we support CheckM. Not BUSCO since that apparently does not directly translate to the numbers needed by Galah. We opened an issue (#1111) where we can discuss this. (I've never used the tool so I have no opinion.)

erikrikarddaniel:
> I’ll take a proper look at the implementation tomorrow but the above are my initial thoughts reading the documentation

Thanks! I appreciate all efforts to get this as good as we can.

docs/output.md
prototaxites: I strongly think we should support all binQC tools, particularly BUSCO. Of the three, BUSCO is the only tool being actively maintained - particularly with active updates to the actual marker gene databases. It's also the default BinQC tool that runs in mag, so to run dereplicate on a default mag run is not so straightforward as just tagging `--dereplicate`.

I would instead suggest moving the code from the GTDB-Tk workflow that collates the completeness/contamination scores into the BinQC workflow. This code currently passes any bin that would be passed by any of the enabled BinQC tools. I would dump that to a file with `.collectFile()` and use that as input to Galah using "dRep" format.

subworkflows/local/dereplication/main.nf
```diff
@@ -0,0 +1,110 @@
...
+    ch_qc_summary  // channel: path(tsv), single study-wide CheckM or CheckM2 summary from BIN_QC.out.checkm_summary/checkm2_summary (bare path, no meta)
+    val_qc_format  // val(string): 'checkm' or 'checkm2', which format ch_qc_summary is in
+
+    main:
+    // Unbinned-contig pseudo-bins aren't real genomes -- CheckM/CheckM2
+    // don't reliably QC them, and Galah panics on any bin missing from its
+    // QC report -- so exclude them here rather than crashing downstream.
+    ch_bins_flat = ch_bins
+        .filter { meta, _bins -> meta.domain != "eukarya" && !meta.refinement.endsWith("unbinned") }
+        .transpose()
```
prototaxites, line 24: What happens these "unbins" in terms of annotation/GTDB-Tk etc.?

subworkflows/local/dereplication/main.nf
```diff
@@ -0,0 +1,110 @@
...
+        // Both CheckM's (--tab_table) and CheckM2's report have the bin ID
+        // (extension-less) in column 1, and Galah looks entries up by the
+        // bin's original extension (.gz stripped) -- same fix nf-core/
+        // modules' own galah module test uses for the same mismatch.
+        GAWK(
+            ch_qc_summary.map { tsv -> [[id: val_qc_format], tsv] },
+            [],
+            false,
+        )
+        ch_qc_summary_for_galah = GAWK.out.output.map { _meta, tsv -> tsv }
```
prototaxites: Can we declare the GAWK program inline? See https://github.com/sanger-tol/nf-core-modules/blob/937bf2d22e930957256aacc56cfe5d23d7ac0cf9/subworkflows/sanger-tol/feature_density/main.nf#L56 for an example.

Are we guaranteed to see `.fa` extensions for every bin?

subworkflows/local/dereplication/main.nf
```diff
@@ -0,0 +1,110 @@
...
+                    return count
+            }
+
+        ch_galah_routed.skip.subscribe {
+            log.warn("[nf-core/mag] Dereplication: no bins passed --dereplicate_min_completeness ${params.dereplicate_min_completeness} / --dereplicate_max_contamination ${params.dereplicate_max_contamination}; skipping Galah for this run (works around a Galah crash on zero qualifying genomes, see https://github.com/wwood/galah/issues/75). Every bin is passed through downstream unclustered, as if --dereplicate had not been set, rather than dropped.")
+        }
+
+        GALAH(ch_galah_routed.cluster)
+
+        // If Galah was skipped, GALAH.out stays empty -- pass every bin
```
prototaxites: Agree with @dialvarezs about simplifying/removing verbose comments!

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
```diff
@@ -424,6 +424,14 @@ def validateInputParameters(hybrid) {
         error('[nf-core/mag] ERROR: To generate the BIgMAG file you need to include the parameters `--run_checkm2` and `--run_gunc`, and you cannot skip BINQC, GTDB-TK, QUAST nor BUSCO.')
     }
 
+    // Check dereplication parameters
+    if (params.dereplicate && (params.skip_binqc || (!params.run_checkm2 && !params.run_checkm))) {
+        error('[nf-core/mag] ERROR: To dereplicate bins with Galah you need to include the parameter `--run_checkm2` and/or `--run_checkm`, and you cannot skip BINQC.')
+    }
+    if (params.dereplicate && params.dereplication_tool != 'galah') {
+        error("[nf-core/mag] ERROR: '${params.dereplication_tool}' is not a supported value for --dereplication_tool; only 'galah' is currently implemented.")
+    }
```
prototaxites: I don't think we need a `params.dereplication_tool` if there's only 1 supported option.

main.nf
prototaxites: I'm happy enough for now that study-wide dereplication only is OK. I think group-wise could also be useful later, but I don't see any use for a sample-level dereplication personally.

prototaxites (commented):
Some more thoughts. Can pick through the documentation properly once we're settled on an implementation.

subworkflows/local/dereplication/main.nf
erikrikarddaniel: Traced it: excluded from Galah dereplication unconditionally, regardless of what any QC tool says (they aren't real curated genomes, and CheckM/CheckM2/BUSCO don't reliably assess them anyway -- Galah panics if it's handed a bin with no matching QC row).

For GTDB-Tk it's actually the same outcome whether or not `--dereplicate` is set, just via different mechanisms: with dereplication on, they're excluded before Galah ever sees them, so they never reach GTDB-Tk's input at all; without dereplication, they do reach GTDB-Tk's input, but GTDB-Tk's own filtering drops any bin missing a QC row -- which, per the above, they normally are. So GTDB-Tk essentially never classifies them either way.

The CAT/BAT unbinned-contig classification leg is unaffected either way (already documented) -- that's a separate, dedicated code path that always runs on them regardless of `--dereplicate`.

Added a clarifying sentence to docs/output.md so this doesn't have to be re-derived from the code next time.

docs/output.md
erikrikarddaniel: Went with your suggested shape rather than a smaller patch: moved the completeness/contamination collation out of GTDB-Tk's subworkflow and into `BIN_QC`, which now builds a study-wide `genome,completeness,contamination` CSV (Galah's native `--genome-info`/dRep format) via `.collectFile()` from whichever of BUSCO/CheckM/CheckM2 ran, preferring CheckM2 > CheckM > BUSCO per bin when more than one assessed it. `DEREPLICATION` always consumes that format now.

That also let me delete the GAWK massaging step entirely, since `genome-info` doesn't need the CheckM/CheckM2-specific filename-suffix fix -- I generate the CSV with the right key convention directly (verified against Galah's actual Rust source: `--genome-info` reuses the exact same fasta-path lookup as `--checkm-tab-table`/`--checkm2-quality-report`, so the same `.gz`-stripped-basename convention applies).

Confirmed end-to-end with a real run of `test_dereplicate` (BUSCO + CheckM2 both on, as in that profile) through GTDB-Tk/CAT-BAT/Prokka/MultiQC.

One thing I didn't take from your suggestion: I kept CheckM2/CheckM preferred over BUSCO rather than treating all three as equally weighted, since CheckM2/CheckM's marker-based completeness/contamination estimates are more established for this purpose than BUSCO's Complete/Duplicated as a proxy -- happy to revisit if you feel differently.

subworkflows/local/dereplication/main.nf
erikrikarddaniel: Resolved this by removing GAWK entirely rather than inlining it -- see the reply on the BUSCO thread. The step existed only to reconcile CheckM/CheckM2's raw bin-ID format with Galah's lookup convention; now that BIN_QC produces the genome_info CSV directly with the correct key already, DEREPLICATION doesn't need to massage anything before handing it to Galah.

And no, `.fa` isn't guaranteed in general, but it is within this pipeline: every binner/refinement module that can produce a bin (`conf/modules.config`) writes `.fa.gz`, consistently, so hardcoding it was safe -- confirmed by grepping the actual publishDir patterns rather than assuming.

subworkflows/local/utils_nfcore_mag_pipeline/main.nf
erikrikarddaniel: Agreed, done -- removed `--dereplication_tool` entirely (schema, main.nf, validation, and the subworkflow's own `if` gate).

subworkflows/local/dereplication/main.nf
erikrikarddaniel: Did another pass -- cut the duplicate Nextflow-scoping explanation and the redundant fallback-routing prose. The subworkflow's also net shorter now that the GAWK step and the `dereplication_tool` `if` gate are both gone (see the other threads).

docs/usage.md
```diff
@@ -483,6 +483,29 @@ In this case, DAS Tool has not necessarily failed but was unable to complete the
...
+
+## A note on bin dereplication
+
+With `--dereplicate`, [Galah](https://github.com/wwood/galah) clusters bins by average nucleotide identity (ANI) across the _whole study_ and picks one representative genome per cluster, using each bin's best-available completeness/contamination estimate to choose the representative (`--run_checkm2`, `--run_checkm` and/or `--run_busco` is required, and `--skip_binqc` must not be set; CheckM2 is preferred over CheckM over BUSCO when more than one ran for the same bin). This is different from DAS Tool's bin refinement above: DAS Tool picks the best bin definition among multiple binners for the _same_ sample/assembly, while dereplication clusters bins _across_ samples that likely represent the same organism, so GTDB-Tk, the CAT/BAT bin-classification leg, and Prokka annotation aren't run redundantly on near-identical genomes recovered independently from multiple samples.
```
prototaxites, line 499: > CheckM2 is preferred over CheckM over BUSCO when more than one ran for the same bin).

Why? For GTDB-Tk, we agreed an approach where any bin that passes any BinQC tool would pass for classification. I'm not sure I like the idea of privileging one tool over another in a hierarchy (all three tools have their weaknesses).

I would say this either recapitulates the GTDB-Tk logic, and/or we need to have a "params.primary_binqc" tool (with a value set to one of the enabled binQC tools) that is used in both situations if set.

subworkflows/local/dereplication/tests/main.nf.test
dialvarezs: I think adding a new testing profile just to test dereplication is overkill.
Let's try to integrate it in a existing one instead, `test_single_end` seems a good match for that (not tested).

subworkflows/local/bin_qc/main.nf
```diff
@@ -227,11 +227,56 @@ workflow BIN_QC {
...
+    // QC summary columns per tool: [bin ID column, completeness column, contamination column]
+    // Same mapping GTDBTK uses to filter bins -- see that subworkflow for why
+    // BUSCO's Complete/Duplicated stand in for completeness/contamination.
+    qc_columns = [
+        checkm2: ['Name', 'Completeness', 'Contamination'],
+        checkm: ['Bin Id', 'Completeness', 'Contamination'],
+        busco: ['Input_file', 'Complete', 'Duplicated'],
+    ]
+
+    // Study-wide "genome,completeness,contamination" table (dRep's genome_info
```
dialvarezs, line 245: Still some verbose comments here.

tests/test_dereplicate.nf.test
dialvarezs, line 1: I think adding a new testing profile just to test dereplication is overkill, specially considering that Galah doesn't even run here.
Let's try to integrate it in a existing one instead, test_single_end seems a good match for that (not tested, but has multiple bins and we can reduce the ANI percentage and completeness to force the colapse).

subworkflows/local/dereplication/main.nf
```diff
@@ -0,0 +1,83 @@
...
+    ch_galah_input = ch_bins_for_galah
+        .combine(ch_genome_info)
+        .map { meta, bins, qc -> [meta, bins, qc, 'genome-info'] }
+
+    // Galah panics on zero qualifying genomes instead of erroring cleanly
+    // (https://github.com/wwood/galah/issues/75), so count them ourselves
+    // first and skip Galah gracefully instead.
+    ch_qualifying_count = ch_genome_info.map { csv ->
+        csv.splitCsv(header: true).count { row ->
+            (row.completeness as Double) >= params.dereplicate_min_completeness && (row.contamination as Double) <= params.dereplicate_max_contamination
```
dialvarezs, line 40: Let's define the used params as workflows inputs instead of using them directly.
I'll try pushing that for mag, it's a better design.

subworkflows/local/dereplication/main.nf
```diff
@@ -0,0 +1,83 @@
...
+    // Unbinned-contig pseudo-bins aren't real genomes -- QC tools don't
+    // reliably assess them, and Galah panics on any bin missing from its
+    // QC report -- so exclude them here rather than crashing downstream.
+    ch_bins_flat = ch_bins
+        .filter { meta, _bins -> meta.domain != "eukarya" && !meta.refinement.endsWith("unbinned") }
+        .transpose()
+
+    ch_bins_for_galah = ch_bins_flat
+        .map { _meta, bin -> bin }
+        .collect()
```
dialvarezs, line 28: This collect could case reproducibility problems due to bin ordering, let's add a sorting step or use `toSortedList { bin -> bin.name }`.

docs/usage.md
```diff
@@ -483,6 +483,29 @@ In this case, DAS Tool has not necessarily failed but was unable to complete the
...
+
+Taxonomic classification (GTDB-Tk, CAT/BAT) and Prokka annotation are affected differently, and it depends on whether `--dereplicate` is also set.
+Without dereplication, they too run directly on whatever `--postbinning_input` selects, so `refined_bins_only` reduces their cost the same way it does for CheckM2/BUSCO.
+With dereplication, they instead run only on Galah's cluster representatives, regardless of `--postbinning_input` -- so that parameter no longer directly bounds their cost.
+What it changes in that case is the pool Galah dereplicates from: `raw_bins_only` hands Galah every binner's bins together, so it ends up deduplicating across binners as well as across samples, while `refined_bins_only` gives Galah DAS Tool's already per-sample-deduplicated bins, a smaller and cleaner pool that's cheaper for Galah itself and less likely to cluster near-duplicate same-sample bins for uninteresting reasons.
```
dialvarezs, line 495: There is another group that is excluded from Prokka and GTDB-Tk when Galah is enabled: the bins that don't pass the quality requirements required by Galah.
We should make that explicit.

subworkflows/local/bin_qc/main.nf
```diff
@@ -227,11 +227,56 @@ workflow BIN_QC {
...
+    // format, also read natively by Galah) for dereplication, built from
+    // whichever QC tool(s) are enabled. Preference order (later entries win on a
+    // shared bin) is CheckM2 > CheckM > BUSCO, reflecting each tool's estimate
+    // quality; a bin only ends up on BUSCO here if neither CheckM tool assessed
+    // it, e.g. --run_busco alone.
+    ch_genome_info = ch_qc_metrics
+        .toList()
```
dialvarezs, line 252: Sorted (for the same reason I mentioned before).

subworkflows/local/dereplication/main.nf
dialvarezs: Also: if `genome_info` doesn't have a row for a bin (because BUSCO or CheckM2 failed), the file will be passed to Galah anyway. Does Galah handle that correctly?

docs/output.md
```diff
@@ -777,6 +777,24 @@ GUNC will be run if specified with `--run_gunc` as a standalone, unless CheckM i
...
+
+<details markdown="1">
+<summary>Output files</summary>
+
+- `GenomeBinning/Dereplication/`
+  - `dereplicated_bins/`
+    - `[bin].fa`: One representative genome fasta per cluster, named exactly as its original bin file.
```
dialvarezs, line 793:
```suggestion
    - `[bin].fa.gz`: One representative genome fasta per cluster, named exactly as its original bin file.
```

dialvarezs (commented):
Another round of comments / questions from me.

docs/usage.md
erikrikarddaniel: I wasn't part of the GTDB-Tk discussion. If we do not prioritize one tool over another, do you suggest that we run Galah once for every combination of CheckM2/M/BUSCO or? I think I would prefer the param solution then, with a default which I personally would set to CheckM2. The problem with that is that it would need to be a list, I believe.

tests/test_dereplicate.nf.test
erikrikarddaniel: Fine with me as long as everything is covered, so, following your suggestion.

subworkflows/local/dereplication/main.nf
erikrikarddaniel: It is. I have it in "my" pipelines. Will fix.

subworkflows/local/bin_qc/main.nf
erikrikarddaniel: Done.

docs/output.md
erikrikarddaniel: Done.

subworkflows/local/dereplication/main.nf
erikrikarddaniel: Ordering implemented.
Galah crashed on bins without quality info; implementing a check for that. This was exposed by `test_single`.

subworkflows/local/bin_qc/main.nf
erikrikarddaniel: Fixed.

docs/usage.md
erikrikarddaniel: Done.

docs/usage.md
prototaxites: Maybe others have some thoughts here? @dialvarezs @jfy133 ?

dialvarezs: Personally, the "any tool that passes" strategy makes more sense to me.
If we don't make Galah and GTDB-Tk strategies consistent, it's possible that some representative bins won't be annotated by GTDB-Tk, leaving the whole cluster without taxonomy info.

prototaxites: I agree. I would say we should mimic the existing approach, and if needed in another PR we can nominate a "primary binqc" param that would apply in both cases.

erikrikarddaniel: I also think "any tool that passes" is good, but prioritization is still required, right? For bins that have quality scores from more than one tool, one needs to be selected. I'm implementing CheckM2 > CheckM > BUSCO as the priority order. Objections to that?

prototaxites: Oh, I see the problem - the scores still need to be fed into Galah 🤦‍♂️ Yes, we need to find a way to prioritise in that case. But I don't like CheckM2 > CheckM > BUSCO at all - I have... questions about the quality of CheckM2, and CheckM2 is unmaintained for years except bugfixes.

I would go BUSCO > CheckM2 > CheckM, personally, but actually would prefer to nominate a param.

erikrikarddaniel: That param would have to be a list. Fine for me. Since this opens all possibilities, I'll go ahead.

prototaxites: I don’t understand why it would have to be a list. If you nominate a tool that isn’t enabled, the pipeline should just fail.

erikrikarddaniel: To accommodate users that run more than one. I never do, but this is the prioritization problem we are facing. If we could rely on users running exactly their favorite tool and nothing else, there's no need to prioritize. But then we don't have the "any tool that passes" situation either.

erikrikarddaniel:
@dialvarezs and @prototaxites : I've made updates, partly through a separate PR, since your last comments. Do you have time to look again?

PR #1114, opened 2026-09-16
Fix test_single_end snapshot flakiness from CPU-dependent floating-point outputs

erikrikarddaniel:
## PR checklist

- [x] Ensure the test suite passes (`nextflow run . -profile test_single_end,docker --outdir <OUTDIR>`).

## Description

Split out of #1110, where I ran into this as an unrelated, pre-existing flakiness issue while working on study-wide bin dereplication.

`test_single_end` CI runs were intermittently failing on `nf-test`'s exact-hash snapshot comparison for several files: `MEGAHIT-{CONCOCT,MaxBin2}-*_pydamage_bin_results.tsv`, `pydamage_bins_summary.tsv`, `bin_summary.tsv`, and (unexpectedly) geNomad's `*_virus_summary.tsv`.

**This went through two iterations while I chased it down (visible in the commit history):**

1. First I assumed it was thread-count-dependent floating-point non-associativity in `CONCOCT_CONCOCT`/`MAXBIN2` (both pass `$task.cpus` as their thread count, and `test_single_end`'s `resourceLimits` gives them more than one CPU) -- the same class of issue `GTDBTK_CLASSIFYWF` already works around in this profile. I pinned both to `cpus = 1` and verified the fix worked identically across two different physical machines.
2. But CI kept failing anyway, on the exact same files, even with the pin in place. Investigating further: CI runs on a `runs-on`-provisioned AWS instance (`m8i-flex.xlarge`), a different CPU generation than either machine I'd verified against. And critically, geNomad's `*_virus_summary.tsv` -- which has nothing to do with binning or thread count at all -- showed the identical pattern: stable across my two machines, consistently different on CI.

That ruled out thread count as the cause. What all the affected files actually have in common is that they're floating-point-derived numeric outputs (ancient-damage statistics, viral classification scores) sensitive to the CI runner's specific CPU (likely something like BLAS/numpy's runtime SIMD dispatch choosing a different code path per CPU generation, though I haven't pinned that down further). The actual bin *composition* -- `.fa.gz` bin files, CONCOCT's `clustering_gt1000.csv` -- never differed across any environment in any of my captures, pinned or not.

So this PR now takes a different approach: **exclude those specific floating-point-derived files from the exact-hash snapshot comparison via `.nftignore`, and replace them with content assertions** (row counts, parseability, and for `pydamage_bins_summary.tsv`, an exact row-count match against the number of per-bin result files it should summarize -- see `bin/summarise_pydamagebins.py`), matching the pattern the test file already used for `bin_summary.tsv`. The `cpus = 1` pin turned out to be unnecessary once these files stopped being hash-compared, and re-verifying across the same two machines with it reverted confirms bin composition itself is unaffected either way -- so it's not included here.

### Worth discussing

I don't have a firm read on whether "exclude + content-check" is the right general pattern for outputs like this, or whether it's worth chasing the actual CPU-dispatch root cause further (e.g. pinning a specific BLAS backend). Opening this as a small PR specifically to get your take, since the same underlying issue likely affects other floating-point-heavy tools in this pipeline's test suite (CheckM2's neural net scoring is one candidate) and may come up again.

https://claude.ai/code/session_01NJ75Ui2JZjJwJ7T7o5acvg


tests/.nftignore
```diff
@@ -12,10 +12,13 @@ Assembly/MEGAHIT/QC/*/*.bowtie2.log
 GenomeBinning/CONCOCT/stats/*_{original,PCA_components,PCA_transformed}_data_gt1000.csv
 GenomeBinning/CONCOCT/stats/*_log.txt
 GenomeBinning/CONCOCT/stats/*.tsv
+GenomeBinning/bin_summary.tsv
```
dialvarezs: Personally I wouldn't exclude `bin_summary.tsv` globally.
It's the main file for checking result stability since it contains almost everything.
A better altenative for me would be excluding it only in the profile affected by the instability: `single_end`. That way we handle the instability from pydamage and genomad without sacrifying reproducibility on other profiles that don't run those tools.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Fixed`
 
+- [#NN](https://github.com/nf-core/mag/pull/NN) - Fix `test_single_end` snapshot flakiness: exclude floating-point-derived per-bin/QC files (pydamage stats, `bin_summary.tsv`, geNomad's virus summary) from exact-hash comparison and replace with content assertions, since these vary by CI runner CPU regardless of thread count (by @erikrikarddaniel)
```
dialvarezs:
```suggestion
- [#1114](https://github.com/nf-core/mag/pull/1114) - Fix `test_single_end` snapshot flakiness: exclude floating-point-derived per-bin/QC files (pydamage stats, `bin_summary.tsv`, geNomad's virus summary) from exact-hash comparison and replace with content assertions, since these vary by CI runner CPU regardless of thread count (by @erikrikarddaniel)
```

dialvarezs (changes requested):
Hey @erikrikarddaniel, thanks for handling this.

I was aware of this issue. Fortunately for me, the main machine where I generally generate snapshots has a Zen4 CPU, that seems to match the one on the CI runners.
But clearly becomes a problem when someone doesn't have access to a CPU with the specific march.

I have only one comment: don't exclude the bin summary globally, because this issue only affects PyDamage and geNomad.

erikrikarddaniel:
> Hey @erikrikarddaniel, thanks for handling this.
> 
> I was aware of this issue. Fortunately for me, the main machine where I generally generate snapshots has a Zen4 CPU, that seems to match the one on the CI runners. But clearly becomes a problem when someone doesn't have access to a CPU with the specific march.
> 
We discovered this when working on the dereplication. After fiddling a bit there, we decided it was better to address on its own. Quite thorny for a small thing actually.

> I have only one comment: don't exclude the bin summary globally, because this issue only affects PyDamage and geNomad.

Done.

dialvarezs (approved):
Looks good!

